Issue #44, opened 2019-07-17
Move to OpenSSL 1.1.x


SalusaSecondus:
While v3 is coming, we do need to at least move to 1.1.0 for better flexibility and more supported algorithms.

PR #60, opened 2019-10-02
Moves to OpenSSL 1.1.1

SalusaSecondus:
*Issue #, if available:* #44

*Description of changes:*

* This moves ACCP from OpenSSL 1.1.1d
* Adds a few more things to .gitignore
* A manual benchmark suggests no significant performance changes


csrc/keyutils.cpp
```diff
@@ -38,25 +38,48 @@ EVP_PKEY* der2EvpPrivateKey(const unsigned char* der, const int derLen, const bo
...
+          } else if (dmq1 && BN_is_zero(dmq1)) {
+            need_rebuild = true;
+          } else if (iqmp && BN_is_zero(iqmp)) {
+            need_rebuild = true;
           }
 
-          // get1_RSA incremented the key's reference count, so we need to free to decrement it again
-          RSA_free(rsa);
+          if (need_rebuild) {
+            // This key likely only has (n, d) set. Very weird, but it happens in java sometimes.
```
bdonlan: If this is the case, where does `e` come from?

csrc/rsa_cipher.cpp
```diff
@@ -76,18 +76,43 @@ JNIEXPORT jint JNICALL Java_com_amazon_corretto_crypto_provider_RsaCipher_cipher
...
+                bn_e = BN_new();
+            }
+            if (!bn_d) {
+                bn_d = BN_new();
+            }
+            if (!RSA_set0_key(r, bn_n, bn_e, bn_d)) {
+                throw_openssl(EX_RUNTIME_CRYPTO, "Unable to set key parameters");
+            }
 
+            if (p && q && !RSA_set0_factors(r, bn_p, bn_q)) {
```
bdonlan: Won't we leak these bignums if we are unable to set them for any reason?

csrc/keyutils.cpp
SalusaSecondus: `e` is set to `0` (as that is how this particular non-standard RSA key is encoded by Java). More importantly, it is ignored provided we never use the key as a public key (which we don't). I'm not 100% certain what would happen if we did try to use the key as a public key.  The important thing about this logic is that we need to unset the CRT parameters. OpenSSL doesn't give a way to clear them, just set them. If they are set (even with the nonsensical value of `0`), OpenSSL will attempt to use them when performing private key operations. This fails. So, by creating a new (clean) RSA key with no CRT parameters set and then explicitly setting `d` and `n` (and the ignored value of `e`), we get a key which can be safely used for private key operations.

csrc/rsa_cipher.cpp
SalusaSecondus: You are correct. I'll need to think through the proper clean-up strategy. `_set0_` methods take ownership of the parameters IIRC, so the exact method of handling this needs to be figured out. There are a bunch of hard to test cases here.

bdonlan: Probably best to use a `take()` pattern for these `set0`s.

SalusaSecondus: I don't think that a `take()` pattern works. I thought about it but the problem is that we don't know if there is a problem until after the method returns. However, we need to `take()` when the parameters are passed in. This means that if the method fails, we've already given up ownership of the objects and thus will leak the memory.

bdonlan: The manpage for RSA_set0_factors states:

> Calling this function transfers the memory management of the values to the RSA object, and therefore the values that have been passed in should not be freed by the caller after this function has been called.

This, to me, implies that it owns the objects even on failure. Is this not the case?

bdonlan: https://github.com/openssl/openssl/blob/e7fb44e7c3f7a37ff83a6b69ba51a738e549bf5c/crypto/rsa/rsa_lib.c#L334

This particular function can only fail if one of the inputs is NULL, but this results in failure to take ownership. I'll file a bug regarding this unclear documentation.

bdonlan: https://github.com/openssl/openssl/issues/10178

SalusaSecondus:
Flakey entropy test failed. Restarted it.

csrc/rsa_cipher.cpp
SalusaSecondus: Thanks for the bug to OpenSSL. Does this mean that the code (as written in ACCP) is correct for memory management?

SalusaSecondus:
Closing and re-opening to restart tests.

csrc/rsa_cipher.cpp
bdonlan: No. If we throw at line 106 we will leak `p`, `q`, `dmp1`, dmq1`, and `iqmp`

SalusaSecondus: Please make sure you are looking at the newest version of the code. I've moved entirely over to `BigNumObj` and think that I'm properly managing the memory.

bdonlan: Ah, I might be looking at an old diff.

csrc/rsa_cipher.cpp
```diff
@@ -70,49 +50,52 @@ JNIEXPORT jint JNICALL Java_com_amazon_corretto_crypto_provider_RsaCipher_cipher
...
+            BigNumObj bn_p = BigNumObj::fromJavaArray(env, p);
+            BigNumObj bn_q = BigNumObj::fromJavaArray(env, q);
+            BigNumObj bn_dmp1 = BigNumObj::fromJavaArray(env, dmp1);
+            BigNumObj bn_dmq1 = BigNumObj::fromJavaArray(env, dmq1);
+            BigNumObj bn_iqmp = BigNumObj::fromJavaArray(env, iqmp);
+
             if (!RSA_set0_key(r, bn_n, bn_e, bn_d)) {
                 throw_openssl(EX_RUNTIME_CRYPTO, "Unable to set key parameters");
+            } else {
+                bn_n.releaseOwnership();
```
bdonlan: Per response on https://github.com/openssl/openssl/issues/10178 this looks correct.

csrc/rsa_cipher.cpp
```diff
@@ -79,14 +59,42 @@ JNIEXPORT jint JNICALL Java_com_amazon_corretto_crypto_provider_RsaCipher_cipher
...
-            r->d = opt_jarr2bn(env, d);
-            r->p = opt_jarr2bn(env, p);
-            r->q = opt_jarr2bn(env, q);
-            r->dmp1 = opt_jarr2bn(env, dmp1);
-            r->dmq1 = opt_jarr2bn(env, dmq1);
-            r->iqmp = opt_jarr2bn(env, iqmp);
+        {
+            // When used with a set0 method, memory ownership transfers to the receiving object.
+            // Thus, after successful ownership transfer, we release ownership of the BIGNUMs.
+            // Once the RSA key owns them, since it is is an RSA_auto class, it cleans itself
```
seebees:
```suggestion
            // Once the RSA key owns them, since it is an RSA_auto class, it cleans itself
```

csrc/aes_gcm.cpp
```diff
@@ -28,37 +29,9 @@
...
-
-        ~native_context() {
-            if (unlikely(leak_tracking)) __sync_add_and_fetch(&active_instances, -1);
-        }
-};
-
 static void initContext(
   raii_env &env,
-  EVP_CIPHER_CTX *ctx,
+  raii_cipher_ctx &ctx,
```
ttjsu-aws: Q: What is the `raii_cipher_ctx` used for and how does it relate to the  `EVP_CIPHER_CTX`? I understand that this replaces the native_ctx used earlier.

ttjsu-aws: Found the answer to this question in this [class raii_cipher_ctx definition](https://github.com/corretto/amazon-corretto-crypto-provider/pull/60/files#diff-cc4e15e2836daff9a013fc62f827643dR93). 

Resolving this conversation.

PR #62, opened 2019-10-03
Yet more fixes to unit test heap errors.

SalusaSecondus:
*Description of changes:*

This cleans up more instance variables in our unit tests to work around JUnit holding onto all test instances and exhausting our heap when running unit tests.

### DIFF
--- a/tst/com/amazon/corretto/crypto/provider/test/EvpKeyAgreementSpecificTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/EvpKeyAgreementSpecificTest.java
@@ -32,27 +32,25 @@ public class EvpKeyAgreementSpecificTest {
     private static final Class<?> SPI_CLASS;
     private static final int EC_TYPE = 408;
     private static final int DH_TYPE = 28;
-    private final KeyPair EC_KEYPAIR;
-    private final KeyPair DH_KEYPAIR;
+    private static final KeyPair EC_KEYPAIR;
+    private static final KeyPair DH_KEYPAIR;
 
     static {
       try {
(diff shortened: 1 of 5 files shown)


tst/com/amazon/corretto/crypto/provider/test/EvpKeyAgreementSpecificTest.java
```diff
@@ -32,27 +32,25 @@
...
-        gen = KeyPairGenerator.getInstance("DH");
-        gen.initialize(1024);
-        DH_KEYPAIR = gen.generateKeyPair();
+          KeyPairGenerator gen = KeyPairGenerator.getInstance("EC");
+          gen.initialize(new ECGenParameterSpec("NIST P-224"));
+          EC_KEYPAIR = gen.generateKeyPair();
+          gen = KeyPairGenerator.getInstance("DH");
+          gen.initialize(1024);
+          DH_KEYPAIR = gen.generateKeyPair();
+      } catch (final Exception ex) {
```
lavaleri: Why are we putting all of this into a generic try/catch?

SalusaSecondus: It's the static initializer for a test case. If this fails in any way, we should just abort. There is no value to a more specific try-catch block since all exceptions result in an "AssertionError" in the end anyway.

lavaleri: ah, got it. I see now that it's all part of the static initializer now. :+1:

Issue #39, opened 2019-07-11
Add "Will Fail" icon to test output


SalusaSecondus:
We should add some small character to all of the individual test outputs letting us know if there has already been a failed test and thus the total process will fail.

SalusaSecondus:
Completed in #63

PR #63, opened 2019-10-08
Add indicator if test suite will fail.

SalusaSecondus:
*Issue #, if available:* #39

*Description of changes:*
Adds a one character prefix to each line of test output which indicates if a test has already failed and thus if the entire suite will fail. 

This has been manually tested by forcing a test failure and observing the output.

### DIFF
--- a/tst/com/amazon/corretto/crypto/provider/test/TestRunner.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/TestRunner.java
@@ -147,7 +147,7 @@ public class TestRunner {
             }
 
             final boolean failed = failureCount > 0;
-            System.out.format("%s Suite %s ran %d tests in %.2f seconds with %d failures. (Ignored %d)\n",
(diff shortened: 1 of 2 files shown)


SalusaSecondus:
Like all of our current test output, it looks best when printed to a console which supports colors. We should make another modification to support no colors (likely controlled by an environment variable).

tst/com/amazon/corretto/crypto/provider/test/TestRunner.java
```diff
@@ -147,20 +147,20 @@ public static void main(String[] args) throws InitializationError {
             }
 
             final boolean failed = failureCount > 0;
-            System.out.println(format("%s Suite %s ran %d tests in %.2f seconds with %d failures. (Ignored %d)",
+            System.out.format("%s Suite %s ran %d tests in %.2f seconds with %d failures. (Ignored %d)\n",
```
WesleyRosenblum: You could use %n instead of \n to be platform independent

PR #67, opened 2019-10-30
Add check for broken Ryzen RDRAND

SalusaSecondus:
**It is important that the `dieharder` acceptance test passes before this PR is merged.**

*Description of changes:*
Some AMD Ryzen CPUs get into a bad state where the output of RDRAND becomes "stuck" on a value of all 1s (e.g., every bit in the output is a one). Unfortunately, it doesn't correctly set the carry flag to indicate an RNG failure.

This change detects this case and handles it as a standard RNG error. It does this by watching for the suspicious value of `0xffffffffffffffff` and if it is seen, checks to see if the RNG is stuck on that value. We take this strategy rather than failing upon simply seeing a single instance of `0xffffffffffffffff` for two reasons:
1. We don't like the odds of a 1/2^64 false positive. It's low, but this library is used in masssively large scale systems.
2. We don't like the (negligible) bias introduced by rejecting any output with the value `0xffffffffffffffff`. This avoids all biased output.

We also check for the suspicious value of `0` using the same technique. While we believe all modern systems to correctly set the carry flag on failure, some old/non-standard systems used a `0` to indicate failure.

For more context please see this [Ars Technical article](https://arstechnica.com/gadgets/2019/10/how-a-months-old-amd-microcode-bug-destroyed-my-weekend/) and this [systemd patch](https://github.com/systemd/systemd/commit/b62bc66018fa1ada09554e7ee46abbbfc8e6b3ad).

To support some libraries which do not define `UINT64_MAX`, I also added a definition for this. As evidence that it is being correctly defined, here is the disassembled version of `rng_rdseed` which is correctly checking against the suspicious value of 64 bits of 1s.
```
00000000000419d0 <_ZN28AmazonCorrettoCryptoProvider10rng_rdrandEPm>:
   419d0:       31 d2                   xor    %edx,%edx
   419d2:       89 d1                   mov    %edx,%ecx
   419d4:       89 d0                   mov    %edx,%eax
   419d6:       48 0f c7 f1             rdrand %rcx
   419da:       0f 92 c0                setb   %al
   419dd:       84 c0                   test   %al,%al
   419df:       41 89 c0                mov    %eax,%r8d
   419e2:       48 89 ce                mov    %rcx,%rsi
   419e5:       48 89 0f                mov    %rcx,(%rdi)
   419e8:       74 0a                   je     419f4 <_ZN28AmazonCorrettoCryptoProvider10rng_rdrandEPm+0x24>
   419ea:       48 8d 49 ff             lea    -0x1(%rcx),%rcx
   419ee:       48 83 f9 fd             cmp    $0xfffffffffffffffd,%rcx
   419f2:       77 04                   ja     419f8 <_ZN28AmazonCorrettoCryptoProvider10rng_rdrandEPm+0x28>
   419f4:       44 89 c0                mov    %r8d,%eax
   419f7:       c3                      retq
   419f8:       89 d1                   mov    %edx,%ecx
   419fa:       89 d0                   mov    %edx,%eax
   419fc:       48 0f c7 f1             rdrand %rcx
   41a00:       0f 92 c0                setb   %al
   41a03:       48 39 ce                cmp    %rcx,%rsi
   41a06:       41 89 c0                mov    %eax,%r8d
   41a09:       75 e9                   jne    419f4 <_ZN28AmazonCorrettoCryptoProvider10rng_rdrandEPm+0x24>
   41a0b:       48 c7 07 00 00 00 00    movq   $0x0,(%rdi)
   41a12:       45 31 c0                xor    %r8d,%r8d
   41a15:       eb dd                   jmp    419f4 <_ZN28AmazonCorrettoCryptoProvider10rng_rdrandEPm+0x24>
   41a17:       66 0f 1f 84 00 00 00    nopw   0x0(%rax,%rax,1)
   41a1e:       00 00
```

### DIFF
--- a/csrc/rdrand.cpp
+++ b/csrc/rdrand.cpp
@@ -292,7 +292,7 @@ bool rng_retry_rdrand(uint64_t *dest) {
             // negligible (1/2^63). Finally, adding processor specific logic would greatly
             // increase the complexity and would cause us to "miss" any unknown processors with
             // similar bugs.
-            if (unlikely(*dest != UINT64_MAX || *dest != 0)) {
+            if (likely(*dest != UINT64_MAX && *dest != 0)) {
                 return true;
             }
         }
--- a/csrc/rdrand.cpp
+++ b/csrc/rdrand.cpp
@@ -144,7 +144,7 @@ bool rng_rdseed(uint64_t *out) {
 
     bool success;
     __asm__ __volatile__(
-	ASM_RDSEED_RCX
+        ASM_RDSEED_RCX
         "setc %%al\n" // rax = 1 if success, 0 if fail
         : "=c" (*out), "=a" (success)
         : "c" (0), "a" (0)
--- a/csrc/rdrand.cpp
+++ b/csrc/rdrand.cpp
@@ -27,11 +27,24 @@ using namespace AmazonCorrettoCryptoProvider;
 
 namespace AmazonCorrettoCryptoProvider {
 
+// Some inline machine-code which cannot always be represented
+// by standard inline assembly mnemonics.
+
+// rdrand %rcx
+#define ASM_RDRAND_RCX ".byte 0x48, 0x0f, 0xc7, 0xf1\n"
+// rdseed %rcx
+#define ASM_RDSEED_RCX ".byte 0x48, 0x0f, 0xc7, 0xf9\n"
+// PAUSE instruction. REP NOP
+#define ASM_REP_NOP ".byte 0xf3, 0x90\n"
+
+// We reuse these same macros for the test results
+// of our RNGs stored in rng_tested
 #define CPUID_PROBE_DONE (1 << 0)
 #define CPUID_HAS_RDRAND (1 << 1)
 #define CPUID_HAS_RDSEED (1 << 2)
 
 volatile static uint32_t cpuid_info = 0;
+volatile static uint32_t rng_tested = 0;
 
 static inline int getCpuid(
         unsigned int level,
@@ -111,13 +124,37 @@ bool rng_rdrand(uint64_t *out) {
 
     bool success = 0;
     __asm__ __volatile__(
-        ".byte 0x48, 0x0f, 0xc7, 0xf1\n" // RDRAND %rcx
+        ASM_RDRAND_RCX
         "setc %%al\n" // rax = 1 if success, 0 if fail
         : "=c" (*out), "=a" (success)
         : "c" (0), "a" (0)
         : "cc" // clobbers condition codes
     );
 
+    // Some AMD CPUs will find that RDRAND "sticks" on all 1s but still reports success.
+    // If we encounter this suspicious value (a 1/2^64 chance) we'll generate a second
+    // value and compare it to the first. If they are equal (indicating it is stuck) then
+    // we'll return an error. Else, we'll return the first value.
+    // This reduces the risk of a false positive to 1/2^128 (negligible) and avoids biasing
+    // the results at all as all 1s can still be returned by a valid RNG.
+    // We also check for 0 as some old/non-standard systems may use that as an error value.
+
+    if (likely(success)) {
+      if (unlikely(*out == UINT64_MAX || *out == 0)) {
+	uint64_t tmp;
+	__asm__ __volatile__(
+            ASM_RDRAND_RCX
+	    "setc %%al\n" // rax = 1 if success, 0 if fail
+	    : "=c" (tmp), "=a" (success)
+	    : "c" (0), "a" (0)
+	    : "cc" // clobbers condition codes
+        );
+	if (tmp == *out) {
+	  *out = 0;
+	  success = 0;
+	}
+      }
+    }
     return success;
 }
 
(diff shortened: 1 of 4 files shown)


csrc/rdrand.cpp
```diff
@@ -111,13 +124,37 @@ bool rng_rdrand(uint64_t *out) {
...
+    // Some AMD CPUs will find that RDRAND "sticks" on all 1s but still reports success.
+    // If we encounter this suspicious value (a 1/2^64 chance) we'll generate a second
+    // value and compare it to the first. If they are equal (indicating it is stuck) then
+    // we'll return an error. Else, we'll return the first value.
+    // This reduces the risk of a false positive to 1/2^128 (negligible) and avoids biasing
+    // the results at all as all 1s can still be returned by a valid RNG.
+    // We also check for 0 as some old/non-standard systems may use that as an error value.
+
+    if (likely(success)) {
+      if (unlikely(*out == UINT64_MAX || *out == 0)) {
```
bdonlan: This introduces a (very low probability) timing side channel if we make use of this value. As such, if we're going to test for these special values, we should instead simply treat them as a failure and obtain a new random value instead.

Further, note that this particular function is replacable in unit tests as a hook point, so any workarounds should be in `rng_retry_rdrand` or `rng_retry_rdseed` instead; this would allow us to actually unit test these workarounds.

SalusaSecondus: 1. I agree with the timing side-channel concern. There will still be a slight bias away from these values, but it certainly looks to be negligible.

2. I'll take a second look at how `rng_retry*` are written and see about moving the logic in there.

bdonlan: The DRBG we feed these values into should debias for us; so we'd just be losing a very very tiny amount of entropy by limiting the space of possible values (less than an attobit per sample, if I'm doing my logarithms properly)

csrc/rdrand.cpp
```diff
@@ -126,15 +163,17 @@ bool rng_rdseed(uint64_t *out) {
         return (*hook_rdseed)(out);
     }
 
-    if (unlikely(!supportsRdSeed())) {
+    // We don't call supportsRdSeed() here to avoid circular dependencies
+    // during RNG testing.
+    if (unlikely(!(get_cpuinfo() & CPUID_HAS_RDSEED))) {
```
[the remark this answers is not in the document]
SalusaSecondus: No longer applicable.

csrc/rdrand.cpp
```diff
@@ -268,13 +278,30 @@ bool rdseed_fallback(uint64_t *dest) {
...
+        uint64_t oldValue = 13; // Must not be initialized to 0 or UINT64_MAX. Anything else is fine.
     do {
-        if (likely(rng_rdrand(dest))) {
-            return true;
-        }
-
-        pause_and_decrement(tries);
+      if (likely(rng_rdrand(dest))) {
+        // Some AMD CPUs will find that RDRAND "sticks" on all 1s but still reports success.
+        // If we encounter this suspicious value (a 1/2^64 chance) we'll generate a second
```
bdonlan: Technically, since we're also checking for 0, this is a 1/2^63 chance.

csrc/rdrand.cpp
```diff
@@ -268,13 +278,30 @@ bool rdseed_fallback(uint64_t *dest) {
...
+        // value and compare it to the first. If they are equal (indicating it is stuck) then
+        // we'll return an error. Else, we'll return the first value.
+        // This reduces the risk of a false positive to 1/2^128 (negligible) and avoids biasing
+        // the results at all as all 1s can still be returned by a valid RNG.
+        // We also check for 0 as some old/non-standard systems may use that as an error value.
+
+        if (*dest == UINT64_MAX || *dest == 0) {
+          if (*dest == oldValue) {
+            // We've gotten the same suspicious value twice in a row. We're likely stuck
```
bdonlan: We already have a perfectly good retry counter, just discard the suspicious values and treat them as a RNG error.

csrc/rdrand.cpp
SalusaSecondus: I'll fix the comment.

csrc/rdrand.cpp
SalusaSecondus: I think that this minor additional complexity is worth it. Just discarding them removes two values entirely from the output. Doing it this way still allows them but detects a stuck `rdrand` in fewer calls.

bdonlan: We cannot allow these values in the output, because if we take a branch on the value, but then make use of the value, we have introduced a timing side-channel. Therefore, we _must_ discard these values and reroll the HWRNG to obtain a different value (one which we will not leak as a timing signal).

SalusaSecondus: Upon further thought, this loop doesn't work exactly how I intended anyway and already (essentially) behaves exactly as you are describing. I will simplify it.

SalusaSecondus:
As discussed with @bdonlan, I've simplified the logic to just discard the suspicious values of `0` and `UINT64_MAX`.

I've also restarted the `dieharder` run. The specific run looked to fail with a false-positive (as happens in all entropy tests). The failure is pasted below so we can see if it recurs.
```
        diehard_sums|   0|       100|     100|0.00072921|   WEAK    
        diehard_sums|   0|       100|     200|0.00288711|   WEAK    
        diehard_sums|   0|       100|     300|0.00005085|   WEAK    
        diehard_sums|   0|       100|     400|0.00005684|   WEAK    
        diehard_sums|   0|       100|     500|0.00005652|   WEAK    
        diehard_sums|   0|       100|     600|0.00005056|   WEAK    
        diehard_sums|   0|       100|     700|0.00000350|   WEAK    
        diehard_sums|   0|       100|     800|0.00000494|   WEAK    
        diehard_sums|   0|       100|     900|0.00000016|  FAILED   
```

PR #72, opened 2019-11-19
Allow resources to be located while in jmod form

alvdavi:
*Description of changes:*
Some of the getResourceAsStream calls will fail if running as a jmod
integrated in a custom jdk. This is due to resource file encapsulation
blocking access using jar absolute paths for non open modules.

By always using paths relative to the Loader.class we can ensure the
code works as a jar or as a jmod.

Resource files within the jar have been moved to package: `com.amazon.corretto.crypto.provider.testdata`


src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -52,10 +51,10 @@ static InputStream getTestData(String fileName) {
             throw new IllegalArgumentException("Invalid filename: " + fileName);
         }
         final InputStream result = AccessController.doPrivileged(
-                (PrivilegedAction<InputStream>) () -> Loader.class.getResourceAsStream("/test-data/" + fileName)
+                (PrivilegedAction<InputStream>) () -> Loader.class.getResourceAsStream("testdata/" + fileName)
```
SalusaSecondus: Please fix JavaDoc on this method.

alvdavi: I saw the CMakeList.txt and I got a bit confused, as that specific variable  seems to point to this directory:
https://github.com/corretto/amazon-corretto-crypto-provider/tree/develop/test-data

Thile the one I've moved is:
https://github.com/corretto/amazon-corretto-crypto-provider/tree/develop/extra-jar-files/test-data

Although both folders have the same names, they do not have the same contents.

Regarding the javadoc, I'll fix it

SalusaSecondus: Hmm, don't move it then. The CMAKE variable is for test data we don't currently ship in the JAR. I got confused.

tst/com/amazon/corretto/crypto/provider/test/AesCtrDrbgTest.java
```diff
@@ -198,7 +198,7 @@ public void nistTestVectors() throws ReflectiveOperationException, SecurityExcep
         int tests = 0;
 
         try (final Scanner in = new Scanner(
-                AmazonCorrettoCryptoProvider.class.getResourceAsStream("/test-data/ctr-drbg.txt"), StandardCharsets.US_ASCII.name())) {
+                AmazonCorrettoCryptoProvider.class.getResourceAsStream("testdata/ctr-drbg.txt"), StandardCharsets.US_ASCII.name())) {
```
SalusaSecondus: As discussed, this may have problems when testing older versions of this library (which we must support).  Try the following code instead

```
Class<?> loaderClass =Class.forName("com.amazon.corretto.crypto.provider.Loader"); // Maybe do this once and make it a public final constant on `TestUtil`?
// Misc code
new Scanner((InputStream) TestUtil.sneakyInvoke(loaderClass, "getTestData", "ctr-drbg.txt"), StandardCharsets.US_ASCII.name());
```

PR #75, opened 2019-12-17
Adds support for IEEE P1363 formatted ECDSA signatures

SalusaSecondus:
*Description of changes:* Adds support for IEEE P1363 formatted ECDSA signatures. These signatures consist of a direct concatenation of the `r` and `s` values as opposed to the default ASN.1 encoding.  Correctness is checked by comparing them to the equivalent "withPLAIN-ECDSA" algorithms from BouncyCastle and Wycheproof.

The specific language from the IEEE specification follows:
> For DL/ECSSA, the output of the signature generation function (see Section 10.2.2) is a pair of integers (c, d). Let r denote the order of the generator (g or G) in the DL or EC settings (see Sections 6.1 and 7.1), and let l = ceil(log256 r) (i.e., l is the length of r in octets). The output (c, d) may be formatted as an octet string as follows: convert the integers c and d to octet strings C and D, respectively, of length l octets each, using the primitive I2OSP, and output the concatenation C || D. To parse the signature, split the octet string into two components C and D, of length l each, and convert them to integers c and d, respectively, using OS2IP. Note that it is essential that both C and D be of length l, even if it means that they have leading zero octets.
>
> NOTE— The output of DL/ECSSA may also be formatted according to the following method, described in more detail in X9.57 [ANS97c] and X9.62 [ANS98e]. Combine c and d into an ASN.1 structure [ISO98a] and encode the structure using some encoding rules, such as Basic Encoding Rules (BER) or Distinguished Encoding Rules (DER) [ISO98e].

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/EvpSignature.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpSignature.java
@@ -7,7 +7,6 @@ import java.nio.ByteBuffer;
 import java.security.SignatureException;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
-import java.util.Arrays;
 
 class EvpSignature extends EvpSignatureBase {
     /** The number a times a key must be reused prior to keeping it in native memory rather than freeing it each time. **/
--- a/src/com/amazon/corretto/crypto/provider/EvpSignature.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpSignature.java
@@ -7,6 +7,7 @@ import java.nio.ByteBuffer;
 import java.security.SignatureException;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
+import java.util.Arrays;
 
 class EvpSignature extends EvpSignatureBase {
     /** The number a times a key must be reused prior to keeping it in native memory rather than freeing it each time. **/
@@ -482,7 +483,7 @@ class EvpSignature extends EvpSignatureBase {
     protected synchronized byte[] engineSign() throws SignatureException {
         ensureInitialized(true);
         try {
-            return signingBuffer.doFinal();
+            return maybeConvertSignatureToReturn(signingBuffer.doFinal());
         } finally {
             engineReset();
         }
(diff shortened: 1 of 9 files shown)


SalusaSecondus:
Overkill failed due to Java Heap Space

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureTest.java
```diff
@@ -76,7 +67,7 @@
         }
     }
 
-    @Parameters(name = "{1}with{0} message_ length {2}. Read-only: {3}, Sliced: {4}")
+    @Parameters(name = "{1} message_ length {2}. Read-only: {3}, Sliced: {4}")
```
scottarc: Do the parameter numbers need to be decremented here since `{0}` is removed?

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureTest.java
```diff
@@ -63,7 +54,7 @@
             tmpMap.put("RSA", kg.generateKeyPair());
 
             kg = KeyPairGenerator.getInstance("EC");
-            kg.initialize(new ECGenParameterSpec("NIST P-384"));
+            kg.initialize(new ECGenParameterSpec("NIST P-521"));
```
scottarc: Is the change from P-384 to P-521 intentional as part of this PR, or an artifact left over from local testing? (It looks harmless either way.)

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureTest.java
SalusaSecondus: They do not. The `{0}` parameter is no longer needed in this format string. Previously `{1}` corresponded to the hash algorithm and `{0}` corresponded with the base asymmetric algorithm (so "{1}with{0}" might become "SHA256withECDSA"). Now the meaning of `{1}` is changed to contain the full algorithm name.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureTest.java
SalusaSecondus: This is intentional. P-521 is large enough that the DER encoded signature needs a "definite, long" length encoding rather than a single-byte "definite, short" length encoding. By making it larger we cover more edge-cases.

scottarc: That makes perfect sense to me.

SalusaSecondus:
Overkill failed due to heap space error on testing. This is a known issue with our test harness. I've restarted the test, but that is a non-blocking issue.

src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
```diff
@@ -154,4 +165,139 @@ protected EvpContext(final long ptr) {
...
+     * If {@code null} is returned then the passed in parameters should be used for later verification.
+     * Otherwise, the entire returned array should be used.
+     * This method has a somewhat odd API since we want to avoid unneccessary array copies/allocations and it is an
+     * internal API anyway.
+     *
+     * @return the converted signature or {@code null} if no conversion is necessary
+     * @throws SignatureException if the signature is badly malformed
+     */
+    protected byte[] maybeConvertSignatureToVerify(byte[] signature, int offset, int length) throws SignatureException {
+        if (algorithmName_ != null && algorithmName_.endsWith("withECDSAinP1363Format")) {
```
WesleyRosenblum: You could make "withECDSAinP1363Format" a constant

src/com/amazon/corretto/crypto/provider/EvpSignature.java
```diff
@@ -7,6 +7,7 @@
 import java.security.SignatureException;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
+import java.util.Arrays;
```
WesleyRosenblum: This isn't being used

src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
```diff
@@ -12,10 +13,14 @@
 import java.security.PublicKey;
 import java.security.SignatureException;
 import java.security.SignatureSpi;
+import java.security.interfaces.ECKey;
+import java.security.interfaces.ECPublicKey;
```
WesleyRosenblum: This isn't used

src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
```diff
@@ -154,4 +165,139 @@ protected EvpContext(final long ptr) {
         }
     }
 
+    /**
+     * Converts and returns the modified signature to verify <em>only if necessary</em> and returns {@code null} otherwise.
+     * If {@code null} is returned then the passed in parameters should be used for later verification.
+     * Otherwise, the entire returned array should be used.
+     * This method has a somewhat odd API since we want to avoid unneccessary array copies/allocations and it is an
```
WesleyRosenblum: typo: "unnecessary"

PR #81, opened 2020-01-09
More test memory fixes

SalusaSecondus:
*Description of changes:*

Yet another attempt to reduce the memory footprint of our tests. It looks like JUnit holds onto all the parameters as well, so this reduces the size of some of them.


SalusaSecondus:
Clearly this doesn't completely fix the problem as overkill still failed (with memory issues). I still suspect this to be an improvement.

SalusaSecondus:
Closing as unneeded for now.

PR #82, opened 2020-01-10
Allow tests to not be parallel

SalusaSecondus:
*Description of changes:*

You can now disable parallel execution of tests by setting the `ACCP_TEST_PARALLEL` environment variable to `false`.  This is necessary to work around some test framework bug in on the `overkill` tests which causes some `OutOfMemoryError`s. We're still trying to figure out the root-cause of these errors (which only occur in testing) but this should unblock us.

I have already modified the `overkill` CodeBuild project to have this environment variable.

### DIFF
--- a/tst/com/amazon/corretto/crypto/provider/test/TestRunner.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/TestRunner.java
@@ -4,6 +4,7 @@
 package com.amazon.corretto.crypto.provider.test;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


SalusaSecondus:
Closing and re-opening to retrigger tests.

tst/com/amazon/corretto/crypto/provider/test/TestRunner.java
```diff
@@ -116,8 +117,8 @@ public static void main(String[] args) throws InitializationError {
         } else {
             printSystemInfo();
             final String suiteName = args[1];
-            final Class[] parallel_classes = SUITES_PARALLEL.get(suiteName);
-            final Class[] serial_classes = SUITES_SERIAL.get(suiteName);
+            Class[] parallel_classes = SUITES_PARALLEL.get(suiteName);
+            Class[] serial_classes = SUITES_SERIAL.get(suiteName);
```
lrstewart: These might be easier to manipulate if you replace them with ```List<Class> parallel_classes = new ArrayList<Class>(Optional.ofNullable(SUITS_PARALLEL).orElse(new Class[0]))``` and then do normal List manipulation below? But entirely a style thing.

PR #114, opened 2020-04-28
Only support explicitly configured keysizes for EC Key generation

SalusaSecondus:
When initialized with an `int`,`KeyPairGenerator` for "EC" keys now only accepts the following explicit values.
  For these values it will always select the corresponding NIST Prime Curve.
  This change removes possible disagreeing behavior between ACCP and SunEC.
  Previously, ACCP would _always_ select the corresponding "secp*r1" curve while SunEC would use undefined behavior to select the curves for any key sizes _not_ on the following list.
  **This changes behavior for keysize values  *other* than the following to throwing an `InvalidParameterException`.**

  * 192
  * 224
  * 256
  * 384
  * 521

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
   This change removes possible disagreeing behavior between ACCP and SunEC.
   Previously, ACCP would _always_ select the corresponding "secp*r1" curve while SunEC would use undefined behavior to select the curves for any key sizes _not_ on the following list.
   **This changes behavior for keysize values  *other* than the following to throwing an `InvalidParameterException`.**
+  [PR #114](https://github.com/corretto/amazon-corretto-crypto-provider/pull/114)
   * 192
   * 224
   * 256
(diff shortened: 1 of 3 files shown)


SalusaSecondus:
Closing to re-think how we roll this out.

CHANGELOG.md
```diff
@@ -4,6 +4,19 @@
...
+  For these values it will always select the corresponding NIST Prime Curve.
+  This change removes possible disagreeing behavior between ACCP and SunEC.
+  Previously, ACCP would _always_ select the corresponding "secp*r1" curve while SunEC would use undefined behavior to select the curves for any key sizes _not_ on the following list.
+  **This changes behavior for keysize values  *other* than the following to throwing an `InvalidParameterException`.**
+  [PR #114](https://github.com/corretto/amazon-corretto-crypto-provider/pull/114)
+  * 192
+  * 224
+  * 256
+  * 384
+  * 521
```
juneb, line 18: When the `KeyPairGenerator` for "EC" keys is initialized with an `int`, that `int` must be one of the following values. `KeyPairGenerator` uses these values to select the corresponding NIST prime curve [PR #114](https://github.com/corretto/amazon-corretto-crypto-provider/pull/114).

  * 192
  * 224
  * 256
  * 384
  * 521

For other values, `KeyPairGenerator` throws an `InvalidParameterException`.

**This is a breaking change to the `KeyPairGenerator` behavior.** Previously, ACCP would _always_ select the corresponding "secp*r1" curve for the specified value. We eliminated support for other values to avoid potential conflicting behavior between ACCP and SunEC. SunEC behavior for other values is undefined.

PR #120, opened 2020-07-26
Implement Java agent as an alternative installation method

jkataja:
*Issue #, if available:*

*Description of changes:*
Adds support for installing ACCP using the Java agent interface, with verification and optional assertion check.
This allows the safe installation of ACCP using the `java -javaagent:AmazonCorrettoCryptoProvider.jar`  argument.
Introducing ACCP thus can involve only ops and can use existing jars, without the need for code changes or rebuilds.

 - Introduces the `Agent` class with the `premain` method invoked by the JVM. 
 - Adds the `Premain-Class` attribute to JAR manifest to configure `Agent` as the agent class.
 - Document runtime option in README.md

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,6 +107,7 @@ set(ACCP_SRC
         src/com/amazon/corretto/crypto/provider/AccessibleByteArrayOutputStream.java
         src/com/amazon/corretto/crypto/provider/AesCtrDrbg.java
         src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
+        src/com/amazon/corretto/crypto/provider/Agent.java
         src/com/amazon/corretto/crypto/provider/EcGen.java
         src/com/amazon/corretto/crypto/provider/EvpKeyAgreement.java
         src/com/amazon/corretto/crypto/provider/EvpKeyType.java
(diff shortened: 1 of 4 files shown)


SalusaSecondus:
Thank you for this PR and we'll look at it.

I would like to better understand the problem which you are solving here and understand if any other JCE providers implement this behavior. Thank you.

jkataja:
As ops, I wish to add ACCP with existing JAR artifacts with just runtime configuration, without changing builds or maintaining multiple branches/configurations, while keeping the runtime safety checks.
As ops, I wish to update ACCP it independently from the application. For instance, when OpenSSL library has to be changed.

I am not knowledgeable in other JCE providers and don't know if other providers implement this installation.

Thanks.

SalusaSecondus:
Thank you. This feedback is very helpful. We currently already support two ways of enabling ACCP at a system level, both of which are standard for JCE providers.

1. Add the ACCP jar to your standard classpath and use the `java.security.properties` system property to install it [as documented in the README](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/README.md#via-security-properties).  So if your old command-line was `java -cp $CLASS_PATH com.main.class.name` your new command-line is `java -cp /path/to/AmazonCorrettoCryptoProvider.jar:$CLASS_PATH -Djava.security.properties=/path/to/amazon-corretto-crypto-provider.security com.main.class.name`
2. Add ACCP directly to your JDK and modify the `java.security` file. This is different depending on which version of the JDK you are using. 
   - JDK8: Add the ACCP Jar to `$JAVA_HOME/jre/lib/ext` and edit `$JAVA_HOME/jre/lib/security/java.security` so that ACCP is listed as provider 1 (with all other providers in the list having their number incremented by one) and modifying the `securerandom.strongAlgorithms` line such that it now starts with `NIST800-90A/AES-CTR-256:AmazonCorrettoCryptoProvider,` (Please see [amazon-corretto-crypto-provider.security](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/etc/amazon-corretto-crypto-provider.security) for examples.)
   - JDK11: I'll need to get back to you on the proper way to handle this. It is possible but we may want to push a code-change to make it easier for you.

Please let us know if these work for you or if you need assistance. As using Java agents to install JCE providers appears non-standard I do not believe that we'll make this change to ACCP.

jkataja:
Thank you for the feedback. I understand the PR might not be merged. I was considering other runtime checks like running on non-libc based distros like Alpine. Perhaps I should instead start a more general purpose optimization agent.

SalusaSecondus:
Right now we don't support non-libc distros. For deeper dependency changes I'll encourage you to look at building the provider yourself and seeing what is needed to make these changes.

SalusaSecondus:
Marking this as "will close" unless the conversation restarts.

PR #122, opened 2020-08-27
Throw IllegalBlockSizeException for overlarge RSA input

SalusaSecondus:
*Description of changes:*
Throw `IllegalBlockSizeException` for overlarge RSA input. This is for better compatibility with the default Java implementation of RSA.


praus:
We generally consider exceptions to be part of the API and thus avoid changing them. However, we are making an exception because this change brings ACCP in line with the documented behavior of Java API which is the behavior our customers are more likely to expect than the current one.

PR #123, opened 2020-08-27
Do not modify output when Cipher.doFinal fails

SalusaSecondus:
*Description of changes:*
This change modifies `Cipher.doFinal` such that when an exception is thrown the out-parameter remain unchanged.

The diff appears to be choking on some indentation changes and thinking this is a bigger change than it appears. I strongly recommend that viewers select the "ignore white-space" option when reviewing this PR as it makes the change much easier to see and read.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@
 /.project
 /.settings/
 /.vscode/
+/.idea/
 *~
+\#*#
 *.ipr
 *.iws
+*.iml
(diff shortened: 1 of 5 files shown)


praus:
I would like to have two people review this change. I can be one of them though if you find two other people before I manage to look at this, feel free to merge. My reasoning for 2-person review is that we've made a mistake in this critical logic once already (presumably we didn't intend to have the behavior we are fixing) and I'd like to reduce the likelihood of another mistake :-).

PR #131, opened 2020-10-15
Reuse state in MessageDigest

SalusaSecondus:
*Description of changes:*
By reusing the underlying state in our implementations of `MessageDigest` we can reduce our allocation rate from ~304 bytes per hash (minimum) to ~48 bytes per hash.

This also removes some related unused code and simplifies some related logic to avoid initializing state multiple times.


AzeemJiva (approved):
LGTM

PR #138, opened 2020-12-18
Better output size estimates. Fixes #135

### LINKED ISSUE #135: AES/GCM getOutputSize can overestimate space needed
Though [Cipher.getOutputSize(int)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/javax/crypto/Cipher.html#getOutputSize(int)) is permitted to overestimate the amount of space needed, ACCP should avoid doing so. Currently it is assuming that OpenSSL may be caching up to one block internally (due to the internal management of the counter). We should either:
1. Determine that this isn't the case and we can just return `inputLen + tagLength`, or
2. Track bytes handled so far so we can give a completely correct answer.

This will make it easier to use our APIs correctly and also more efficient as callers will not need to allocate extra memory (which then must be truncated.

See [AesGcmSpi.engineGetOutputSize(int)](https://code.amazon.com/packages/NativeJCEBindings/blobs/479c9b9bc77d236ec009a3c90b563ea4d7db9e22/--/src/com/amazon/nativejcebindings/AesGcmSpi.java#L237).

SalusaSecondus:
Fixes #135

This improves the estimates provided by `Cipher.getOutputSize()` such that for AES-GCM they are completely accurate (when used with `doFinal` and not overestimates. (For use with `update` they must remain overestimates because they must take into account the size of the tag output from `doFinal`).

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
@@ -237,9 +237,7 @@ final class AesGcmSpi extends CipherSpi {
     protected int engineGetOutputSize(int inputLen) {
         switch (opMode) {
             case NATIVE_MODE_ENCRYPT:
-                // Ensure we have room (on doFinal) for either an extra ciphertext block that was buffered in OpenSSL,
-                // or the final tag (whichever is bigger)
-                return inputLen + Math.max(tagLength, engineGetBlockSize() - 1);
+                return getUpdateOutputSize(inputLen) + tagLength;
             case NATIVE_MODE_DECRYPT:
                 return Math.max(0, decryptInputBuf.size() + inputLen - tagLength);
             default:
@@ -253,7 +251,7 @@ final class AesGcmSpi extends CipherSpi {
     private synchronized int getUpdateOutputSize(int inputLen) {
         switch (opMode) {
             case NATIVE_MODE_ENCRYPT:
-                return inputLen + engineGetBlockSize() - 1;
+                return inputLen;
             case NATIVE_MODE_DECRYPT:
                 // We do not return data from engineUpdate when decrypting - all data is returned from engineDoFinal()
                 return 0;
@@ -854,6 +852,9 @@ final class AesGcmSpi extends CipherSpi {
     }
 
     private void checkOutputBuffer(int inputLength, byte[] output, int outputOffset) throws ShortBufferException {
+        if (inputLength < 0 || outputOffset < 0 || outputOffset > output.length) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
         if (output.length - outputOffset < getUpdateOutputSize(inputLength)) {
             throw new ShortBufferException("Expected a buffer of at least " + engineGetOutputSize(inputLength) + " bytes; got " + (output.length - outputOffset));
         }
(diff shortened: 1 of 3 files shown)


farleyb-amazon (approved):

tst/com/amazon/corretto/crypto/provider/test/AesTest.java
```diff
@@ -400,11 +504,6 @@ public void test_getOutputSize_encrypt() throws Throwable {
 
         // Allows room for the final tag
         assertEquals(12345 + 16, sneakyInvoke_int(spi, "engineGetOutputSize", 12345));
-
-        // Allows room for a partial buffered block
-        sneakyInvoke(spi, "engineInit", Cipher.ENCRYPT_MODE, key,
-                new GCMParameterSpec(12 * 8, randomIV()), rnd);
-        assertEquals(12345 + 15, sneakyInvoke_int(spi, "engineGetOutputSize", 12345));
```
seebees, line 407: Just wondering why we can delete this? Since it is not immediately obvious to me I figured I'd as while I'm reading :)

tst/com/amazon/corretto/crypto/provider/test/AesTest.java
```diff
@@ -538,7 +637,7 @@ public void test_bufferOverflows() throws Throwable {
         sneakyInvoke(spi, "engineInit", Cipher.ENCRYPT_MODE, key,
                 new GCMParameterSpec(128, randomIV()), rnd);
         assertThrows(ShortBufferException.class,
-                () -> sneakyInvoke(spi, "engineUpdate", new byte[16], 0, 16, new byte[32], 16));
```
seebees, line 541: Same comment, why 16 -> 17?

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -854,6 +852,9 @@ protected int engineDoFinal(ByteBuffer input, ByteBuffer output) throws ShortBuf
     }
 
     private void checkOutputBuffer(int inputLength, byte[] output, int outputOffset) throws ShortBufferException {
+        if (inputLength < 0 || outputOffset < 0 || outputOffset > output.length) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
```
seebees, line 857: Given that we added this condition, it would be good to have a test for this condition.

tst/com/amazon/corretto/crypto/provider/test/AesTest.java
SalusaSecondus: This is actually testing for the buggy behavior. Since the buggy behavior was never actually required, we are removing the test.

tst/com/amazon/corretto/crypto/provider/test/AesTest.java
SalusaSecondus: Same answer: This is actually testing for the buggy behavior. Since the buggy behavior was never actually required, we are removing the test.

seebees (approved):

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
SalusaSecondus: Tested by existing tests. That's how I discovered this new line was needed.

PR #145, opened 2021-03-08
Move to OpenSSL 1.1.1j

alex-chew:
Upgrades ACCP to OpenSSL version 1.1.1j.

Note that ACCP is not impacted by the CVEs listed in the [corresponding advisory](https://www.openssl.org/news/secadv/20210216.txt).

* ACCP does not use `X509_issuer_and_serial_hash` and so is not affected by CVE-2021-23841.
* ACCP does not use `RSA_padding_check_SSLv23` and so is not affected by CVE-2021-23839.
* Where ACCP uses `EVP_CipherUpdate` the input length never exceeds 256KB (far from supported platforms' maximum integer size); where it uses `EVP_EncryptUpdate` the input length is always `16` (far from maximum); and it does not use `EVP_DecryptUpdate` - so ACCP is not affected by CVE-2021-23840.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ method.
 * Stricter guarantees about which curves are used for EC key generation. [PR #127](https://github.com/corretto/amazon-corretto-crypto-provider/pull/127)
 * Reduce timing signal from trimming zeros of TLSPremasterSecrets from DH KeyAgreement. [PR #129](https://github.com/corretto/amazon-corretto-crypto-provider/pull/129)
 * Reuse state in `MessageDigest` to decrease object allocation rate. [PR #131](https://github.com/corretto/amazon-corretto-crypto-provider/pull/131)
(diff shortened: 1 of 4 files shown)


AzeemJiva (approved):

SalusaSecondus:
The code looks good. I do have a comment on the PR description however.

When ACCP uses `EVP_CipherUpdate` it does correctly look at the returned output length (called `outl` in the [relevant code](https://github.com/corretto/amazon-corretto-crypto-provider/blob/bdfb7cf955007532efd7aa43b0452d1493bdf4a7/csrc/aes_gcm.cpp#L68-L101)). However, you can also see on line 78 that ACCP doesn't handle more than `CHUNK_SIZE` bytes at a time where `CHUNK_SIZE` is 256KB (much lower than the affected limits of 2^31-1).

SalusaSecondus (approved):
LGTM

alex-chew:
> When ACCP uses EVP_CipherUpdate it does correctly look at the returned output length (called outl in the relevant code). However, you can also see on line 78 that ACCP doesn't handle more than CHUNK_SIZE bytes at a time where CHUNK_SIZE is 256KB (much lower than the affected limits of 2^31-1).

Ah, I missed that callsite. I'll add this to the PR description. (I originally only saw the callsite at aes_gcm.cpp:280)

Issue #144, opened 2021-03-04
Cipher init does not validate key argument


bauer-matthews:
A call to Cipher init should validate that the `key` argument is not null. This is done for RSA Cipher instances but not for AES/GCM cipher instances.  Below is a simple test case to reproduce the issue. 

```
    @Test(expectedExceptions = {InvalidKeyException.class, IllegalArgumentException.class})
    public void cipherInitNullKeyCorretto() throws NoSuchAlgorithmException, NoSuchPaddingException,
            NoSuchProviderException, InvalidKeyException {

        com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider.install();

        Random rand = new Random();
        byte[] input = new byte[128];
        rand.nextBytes(input);

        Key key = null;

        Cipher cipher = Cipher.getInstance("AES/GCM/NoPadding", "AmazonCorrettoCryptoProvider");
        cipher.init(Cipher.ENCRYPT_MODE, key); // This should throw InvalidKeyException or IllegalArgumentException
        cipher.update(input);
    }
```

SalusaSecondus:
I see that `NullPointerException` and this is definitely something we should fix. Thank you for the report.

```
Exception in thread "main" java.lang.NullPointerException: Null array passed
        at com.amazon.corretto.crypto.provider.AesGcmSpi.encryptInit(Native Method)
        at com.amazon.corretto.crypto.provider.AesGcmSpi.lazyInit(AesGcmSpi.java:888)
        at com.amazon.corretto.crypto.provider.AesGcmSpi.engineUpdate(AesGcmSpi.java:427)
        at com.amazon.corretto.crypto.provider.AesGcmSpi.engineUpdate(AesGcmSpi.java:399)
        at java.base/javax.crypto.Cipher.update(Cipher.java:1829)
        at CipherInitNullKeyCorretto.main(CipherInitNullKeyCorretto.java:14)
```

alex-chew:
Closing as this will be fixed in the upcoming release.

PR #146, opened 2021-03-11
Validate that AesGcmSpi#engineInit gets non-null key

alex-chew:
*Issue #, if available:* #144

*Description of changes:* Ensure that all initialization paths in AesGcmSpi validate that the passed Key is non-null, like RsaCipher does.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
@@ -294,6 +294,10 @@ final class AesGcmSpi extends CipherSpi {
     protected synchronized void engineInit(
         int jceOpMode, Key key, AlgorithmParameterSpec algorithmParameterSpec, SecureRandom secureRandom
     ) throws InvalidKeyException, InvalidAlgorithmParameterException {
+	if (key == null) {
+	    throw new InvalidKeyException("Key can't be null");
+	}
+
         final GCMParameterSpec spec;
         if (algorithmParameterSpec instanceof GCMParameterSpec) {
             spec = (GCMParameterSpec) algorithmParameterSpec;
--- a/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
@@ -294,9 +294,9 @@ final class AesGcmSpi extends CipherSpi {
     protected synchronized void engineInit(
         int jceOpMode, Key key, AlgorithmParameterSpec algorithmParameterSpec, SecureRandom secureRandom
     ) throws InvalidKeyException, InvalidAlgorithmParameterException {
-	if (key == null) {
-	    throw new InvalidKeyException("Key can't be null");
-	}
+        if (key == null) {
+            throw new InvalidKeyException("Key can't be null");
+        }
 
         final GCMParameterSpec spec;
         if (algorithmParameterSpec instanceof GCMParameterSpec) {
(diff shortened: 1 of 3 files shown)


tst/com/amazon/corretto/crypto/provider/test/AesTest.java
```diff
@@ -579,6 +581,24 @@ public void test_initParameters() throws Throwable {
         assertArrayEquals(PLAINTEXT, decrypted);
     }
 
+    @SuppressWarnings("ConstantConditions")
```
SalusaSecondus, line 584: I'm not familiar with this one. Not an objection, but what is it?

tst/com/amazon/corretto/crypto/provider/test/AesTest.java
```diff
@@ -579,6 +581,24 @@ public void test_initParameters() throws Throwable {
...
+    @Test
+    public void test_initNullKey() throws Throwable {
+        jceC.init(Cipher.ENCRYPT_MODE, key);
+
+        Key key = null;
```
SalusaSecondus, line 589: NitPick: Variables should be `final` if possible.

tst/com/amazon/corretto/crypto/provider/test/AesTest.java
```diff
@@ -579,6 +581,24 @@ public void test_initParameters() throws Throwable {
         assertArrayEquals(PLAINTEXT, decrypted);
     }
 
+    @SuppressWarnings("ConstantConditions")
+    @Test
+    public void test_initNullKey() throws Throwable {
+        jceC.init(Cipher.ENCRYPT_MODE, key);
```
SalusaSecondus, line 587: Since this test will fail on earlier versions of ACCP, we need to add a check at the beginning of the test which bails out early (without failing) if an older version is under test.

```suggestion
        assumeMinimumVersion("1.6.0", AmazonCorrettoCryptoProvider.INSTANCE);
        jceC.init(Cipher.ENCRYPT_MODE, key);
```

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -294,6 +294,10 @@ protected synchronized void engineInit(int opMode, Key key, SecureRandom secureR
     protected synchronized void engineInit(
         int jceOpMode, Key key, AlgorithmParameterSpec algorithmParameterSpec, SecureRandom secureRandom
     ) throws InvalidKeyException, InvalidAlgorithmParameterException {
+	if (key == null) {
+	    throw new InvalidKeyException("Key can't be null");
+	}
```
SalusaSecondus, line 299: Check indent? This looks like it might not be the same level as other code in this method.

SalusaSecondus (changes requested):

tst/com/amazon/corretto/crypto/provider/test/AesTest.java
alex-chew: Ah, this just tells IntelliJ IDEA not to complain that `key` is `null` everywhere in the method body.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
alex-chew: Yep, my bad. It's fixed now.

tst/com/amazon/corretto/crypto/provider/test/AesTest.java
alex-chew: Fixed.

alex-chew:
A test failed in the `accp_github_pr` build, but it was unrelated to these changes and I couldn't reproduce the failure on my machine. Re-running the build.

SalusaSecondus (approved):
LGTM

Issue #143, opened 2021-03-04
Cipher doFinal gives null dereference when input bytes are missing


bauer-matthews:
Cipher doFinal gives null dereference when input bytes are missing. Below is a simple test case to demonstrate the issue. With the SunJCE provider (openJDK 11) no exception is raised and the returned byte array is all zeros.  

```
    @Test
    public void cipherNoInputCorretto() throws NoSuchAlgorithmException, NoSuchPaddingException,
            InvalidKeyException, IllegalBlockSizeException, BadPaddingException, NoSuchProviderException {

        com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider.install();

        KeyPairGenerator keygen = KeyPairGenerator.getInstance("RSA");
        keygen.initialize(2048);
        Key key = keygen.generateKeyPair().getPublic();

        Cipher cipher = Cipher.getInstance("RSA/ECB/NoPadding", "AmazonCorrettoCryptoProvider");
        cipher.init(Cipher.ENCRYPT_MODE, key);
        cipher.doFinal();
    }
```

SalusaSecondus:
Thank you for this report. I do see that we have a `NullPointerException` in this case and that should be fixed.

----
My repro.

```
Exception in thread "main" java.lang.NullPointerException
        at java.base/java.lang.System.arraycopy(Native Method)
        at com.amazon.corretto.crypto.provider.RsaCipher.engineDoFinal(RsaCipher.java:195)
        at com.amazon.corretto.crypto.provider.RsaCipher.engineDoFinal(RsaCipher.java:148)
        at java.base/javax.crypto.Cipher.doFinal(Cipher.java:2085)
        at CipherNoInputCorretto.main(CipherNoInputCorretto.java:14)
```

alex-chew:
Closing as this will be fixed in the upcoming release.

PR #147, opened 2021-03-13
Handle RsaCipher#engineDoFinal with no input bytes

alex-chew:
*Issue #, if available:* #143

*Description of changes:* Ensure that RsaCipher correctly returns all 0 bytes when calling `Cipher#doFinal` without any input bytes provided (via `update` or `doFinal`), like SunJCE does.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/RsaCipher.java
+++ b/src/com/amazon/corretto/crypto/provider/RsaCipher.java
@@ -181,7 +181,7 @@ class RsaCipher extends CipherSpi {
             // that don't take an input buffer, and in those cases, inputOffset and inputLen are 0.
             // We set them here anyways to be safe, because the API makes no such guarantee.
             else if (input == null) {
-                input = new byte[0];
+                input = Utils.EMPTY_ARRAY;
                 inputOffset = 0;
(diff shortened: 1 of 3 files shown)


src/com/amazon/corretto/crypto/provider/RsaCipher.java
```diff
@@ -177,6 +177,14 @@ protected int engineDoFinal(byte[] input, int inputOffset, int inputLen, final b
                 inputOffset = 0;
                 inputLen = buffer_.size();
             }
+            // One-shot, no input. Cipher only calls engineDoFinal with null input in doFinal overloads
+            // that don't take an input buffer, and in those cases, inputOffset and inputLen are 0.
+            // We set them here anyways to be safe, because the API makes no such guarantee.
+            else if (input == null) {
+                input = new byte[0];
```
SalusaSecondus, line 184: Please use the zero-byte array constant instead.
```suggestion
                input = Utils.EMPTY_ARRAY;
```

SalusaSecondus (changes requested):
One minor change needed and this is good.

SalusaSecondus (approved):
LGTM

Issue #141, opened 2021-02-19
Ensure all tests pass under JDK15


SalusaSecondus:
We must ensure that all of our current tests pass when run under JDK15. Right now some are failing because JDK15 has removed support for certain curves. In the short term we should not test those curves on JDK15. In the long term we should probably completely remove support for those curves and restrict ourselves to a much smaller set of known good curves.

PR #165, opened 2021-08-31
Jdk17 tests

lavaleri:
*Issue #, if available:* #141

*Description of changes:* 

Updates to build on JDK17:
- Upgrade gradle wrapper
- Fix some java doc issues
- separate ECCurve tests by JRE

TODO:
- Update Codebuild to test on JDK17


CMakeLists.txt
```diff
@@ -590,8 +590,19 @@ add_custom_command(
 add_custom_target(tests-jar DEPENDS ${TESTS_JAR})
 set_property(TARGET tests-jar PROPERTY JAR_FILE ${TESTS_JAR})
 
+# These flags are necessary in Java17+ in order for certain unit tests to
+# perform deep reflection on nonpublic members
+if (Java_VERSION_MAJOR VERSION_GREATER_EQUAL 17)
+  set(ADD_OPENS
```
SalusaSecondus, line 596: Let's rename this to `TEST_ADD_OPENS` so that it is even more clear that this is only to be used for tests.

build.gradle
```diff
@@ -437,7 +437,7 @@ if (project.hasProperty('mavenUser') && project.hasProperty('jcecertAlias')) {
         }
     }
 
-    task sign(overwrite: true) {
+    task sign {
```
SalusaSecondus, line 440: Any more context on this change?

tst/com/amazon/corretto/crypto/provider/test/EcGenTest.java
```diff
@@ -52,6 +54,14 @@
 @ResourceLock(value = TestUtil.RESOURCE_GLOBAL, mode = ResourceAccessMode.READ)
 public class EcGenTest {
     public static final String[][] KNOWN_CURVES = new String[][] {
+            // Not supported by Openssl
```
SalusaSecondus, line 57: Let's remove this comment. It is old and I'm not sure what it originally intended. (Same with other copy of this comment elsewhere in the same file.)

tst/com/amazon/corretto/crypto/provider/test/EcGenTest.java
```diff
@@ -146,40 +166,60 @@ public void teardown() {
...
-            // Check encoding/decoding
-            Key bouncedKey = KEY_FACTORY.generatePublic(new X509EncodedKeySpec(nativePair.getPublic().getEncoded()));
-            assertEquals(nativePair.getPublic(), bouncedKey, "Public key survives encoding");
-            bouncedKey = KEY_FACTORY.generatePrivate(new PKCS8EncodedKeySpec(nativePair.getPrivate().getEncoded()));
-            assertEquals(nativePair.getPrivate(), bouncedKey, "Private key survives encoding");
+            testCurveByName(name);
         }
     }
 
+    private void testCurveByName(Object name) throws GeneralSecurityException {
```
SalusaSecondus, line 173: Please have this take a `String` as an argument and force callers to cast.

tst/com/amazon/corretto/crypto/provider/test/EcGenTest.java
```diff
@@ -205,6 +245,8 @@ public void unknownSize() throws GeneralSecurityException {
                 () -> nativeGen.initialize(13));
     }
 
+    // TODO Update EXPLICIT_CURVE to non-legacy curve
```
SalusaSecondus, line 248: Let's fix this now. It isn't that hard to do. Just pull the values from here: https://github.com/corretto/corretto-8/blob/e6eb5c6a8c985ff8f1aa299d87fe5b798b5c2a3e/src/jdk/src/share/classes/sun/security/util/CurveDB.java#L298-L305

build.gradle
lavaleri: With gradle upgrade from 5 to 7, using 'overwrite' is invalid if the task doesn't already exist. If we don't make this change, gradle fails with the error:
```
> Could not create task ':sign'.
   > Unnecessarily replacing a task that does not exist is not supported.  Use create() or register() directly instead.  You attempted to replace a task named 'sign', but there is no existing task with that name.
```
My guess is it was not necessary for `sign` to overwrite in the first place.

SalusaSecondus:
Changes so far look good.

CMakeLists.txt
```diff
@@ -590,8 +590,20 @@ add_custom_command(
 add_custom_target(tests-jar DEPENDS ${TESTS_JAR})
 set_property(TARGET tests-jar PROPERTY JAR_FILE ${TESTS_JAR})
 
+# These flags are necessary in Java17+ in order for certain unit tests to
+# perform deep reflection on nonpublic members
+if (Java_VERSION_MAJOR VERSION_GREATER_EQUAL 17)
```
lavaleri, line 595: Using `Java_VERSION_MAJOR` is not quite right because `JAVA_HOME` can be different from `TEST_JAVA_HOME`.

SalusaSecondus (approved):
LGTM

SalusaSecondus (approved):
Please delete the now unneeded patch file.

SalusaSecondus (dismissed):
LGTM

SalusaSecondus (approved):
LGTM

PR #169, opened 2021-11-29
Add TLS1.3 to local integ tests.

### LINKED ISSUE #166: Update LocalHTTPSIntegrationTest to test TLS 1.3
`LocalHTTPSIntegrationTest` does not currently test TLS 1.3 and we should add support for this.

SalusaSecondus:
Add TLS1.3 to local integ tests. (Also fixes "AACP" typo throughout.)

Resolves #166. 

Manual verification of test output shows that TLS 1.3 suites are being properly tested. Manual testing against JDK16 shows that the ChaCha20 suites are also tested when the rest of the system supports it. (Additional manual testing against JDK8 shows that these tests work there too. Default JDK for testing is 11.)


alvdavi (approved):
LGTM

PR #176, opened 2022-02-25
Allow SignatureException on ECDSA verify failure

WillChilds-Klein:
*Issue #, if available:* CryptoAlg-1021

*Description of changes:*

This commit modifies the EvpSignatureTest to allow for
`SignatureException`s to be thrown upon ECDSA verification failure (as
opposed to simply returning `false`). Before this change, a number of
tests failed due to `SignatureException`s being thrown on ECDSA
verification failure (but not RSA verification†).

A relevant difference in behavior between OpenSSL and AWS-LC is that
OpenSSL [**does not** set][1] an error code when it fails to verify an
otherwise valid ECDSA signature while [AWS-LC **does**][2] (i.e. the
signature is well-formed and uncorrupted, but it is inauthentic per the
specified public key). ~~Interestingly, AWS-LC's ancestor BoringSSL
[doesn't set an error][4] on ECDSA verification failure either.~~
AWS-LC's setting of a `BAD_SIGNATURE` error for _DSA_ verification
feailure is consistent with [AWS-LC's treatment of _RSA_ signature
verification failures][3].

When we (ACCP) see an error after attempting (non-RSA) verification, we
[throw an exception][8]. Unfortunately, because AWS-LC currently
conflates _corrupted_ and _inauthentic_ signatures under the same
`BAD_SIGNATURE` error code, we can not differentiate the cases, and thus
will never return `false` on ECDSA verification -- we'll either return
`true` if the signature is valid and authentic, else throw. The
[javadocs][9] for `Signature.verify` seem to indicate that we should
indeed return boolean `false` on verification failure. In order to do
this, we'll need to make changes to AWS-LC to either set no error code
upon verification failure or to set a distinct code which we can key off
of.

---

† A [prior commit][5] indicates that the JCA/JCE specification
stipulates that we suppress exceptions for RSA, but I was unable to find
any such stipulation in the [latest JCA docs][6] or [the javadoc for
`SignatureException`][7].

[1]: https://github.com/openssl/openssl/blob/5cbd2ea3f94aa8adec9b4486ac757d4d688e3f8c/crypto/ec/ecdsa_ossl.c#L458
[2]: https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/ecdsa/ecdsa.c#L194-L197
[3]: https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/rsa/rsa.c#L620-L621
[4]: https://boringssl.googlesource.com/boringssl/+/2214/crypto/ecdsa/ecdsa.c#215
[5]: https://github.com/corretto/amazon-corretto-crypto-provider/commit/dcbf8a7dc6fb5064b979739e2ffc590de61c39a2#diff-7fe3b81789f62d5bbf2ff0d8a6a8db06393630449ff914f9127f788084b06e49R481-R484
[6]: https://docs.oracle.com/en/java/javase/17/security/java-cryptography-architecture-jca-reference-guide.html
[7]: https://docs.oracle.com/javase/7/docs/api/java/security/SignatureException.html
[8]: https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/csrc/sign.cpp#L483
[9]: https://docs.oracle.com/javase/7/docs/api/java/security/Signature.html#verify(byte[])

### DIFF
--- a/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureTest.java
@@ -311,14 +311,7 @@ public class EvpSignatureTest {
         params.verifier.update(params.message);
         byte[] badSignature = params.goodSignature.clone();
         badSignature[badSignature.length - 1]++;
-        try {
-            assertFalse(params.verifier.verify(badSignature));
-        } catch (final SignatureException ex) {
-            if (params.algorithm.contains("RSA")) {
-                // RSA is not allowed to fail with an exception
-                throw ex;
-            }
-        }
+        expectInvalidSignature(params, badSignature);
     }
 
     @ParameterizedTest
@@ -327,14 +320,7 @@ public class EvpSignatureTest {
         params.verifier.update(params.message);
         byte[] badSignature = Arrays.copyOf(params.goodSignature, params.goodSignature.length - 1);
         // Truncated signatures will sometime return false and sometimes throw an exception. Both are acceptable
-        try {
-            assertFalse(params.verifier.verify(badSignature));
-        } catch (final SignatureException ex) {
-            if (params.algorithm.contains("RSA")) {
-                // RSA is not allowed to fail with an exception
-                throw ex;
-            }
-        }
+        expectInvalidSignature(params, badSignature);
     }
 
     // Modification of body of the message only works
@@ -349,7 +335,7 @@ public class EvpSignatureTest {
         byte[] msgCopy = params.message.clone();
         msgCopy[0]++;
         params.verifier.update(msgCopy);
-        assertFalse(params.verifier.verify(params.goodSignature));
+        expectInvalidSignature(params, params.goodSignature);
     }
 
     // If we're already beyond the message size limit, we expect truncation to be ignored
@@ -361,7 +347,7 @@ public class EvpSignatureTest {
     @MethodSource("verifyTruncatedMessageParams")
     public void verifyTruncatedMessage(TestParams params) throws Exception {
         params.verifier.update(Arrays.copyOf(params.message, params.message.length - 1));
-        assertFalse(params.verifier.verify(params.goodSignature));
+        expectInvalidSignature(params, params.goodSignature);
     }
 
     // If we're just at the message size limit, any additional bytes will be ignored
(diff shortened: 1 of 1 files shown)


WillChilds-Klein:
bad news is that CodeBuild checks are failing because some tests are still failing. good news is that the number of failing tests is down to 2. here's a test run of this commit:

```
root@b57f6f874903:/accp# ./gradlew test
...
Test run finished after 309739 ms
[        88 containers found      ]
[         0 containers skipped    ]
[        88 containers started    ]
[         0 containers aborted    ]
[        88 containers successful ]
[         0 containers failed     ]
[     12396 tests found           ]
[         0 tests skipped         ]
[     12396 tests started         ]
[         2 tests aborted         ]
[     12392 tests successful      ]
[         2 tests failed          ]
...
BUILD FAILED in 6m 25s
```

SalusaSecondus:
The expected signature exception  prior commit is due to failures in the [JCK](https://en.wikipedia.org/wiki/Technology_Compatibility_Kit#TCK_for_the_Java_platform) without these exceptions properly managed. I'll dig up some links for you. Actual behavior does not appear to be documented.

In general:
1. We should avoid throwing an exception rather than just returning `false`. This is confusing behavior for callers and its better to have just one way a signature can be rejected whenever possible.
2. We need to be very careful changing behavior of our public APIs and changing when a method throws exceptions is a behavior change.

WillChilds-Klein:
ok, makes sense. i'll cut a SIM to track the changes to AWS-LC. if they agree that it's sensible to elide error codes entirely on verification failure (as OpenSSL + BoringSSL do), we can discard this PR. if they want to break out and return a distinct error code, i'll amend this PR to look for that error code. i expect we'll probably go the former route.

WillChilds-Klein:
**CORRECTION**: in the commit message above, i was looking at an outdated mirror of BoringSSL. the current [googlesource repo](https://github.com/google/boringssl/blob/master/crypto/fipsmodule/ecdsa/ecdsa.c#L194) and its [github mirror](https://github.com/google/boringssl/blob/master/crypto/fipsmodule/ecdsa/ecdsa.c#L194) show that BoringSSL also diverges from OpenSSL in that it sets `ECDSA_R_BAD_SIGNATURE` on verification failure of an otherwise well-formed signature.

WillChilds-Klein:
follow-ups PRs

ACCP: https://github.com/corretto/amazon-corretto-crypto-provider/pull/180
AWS-LC: https://github.com/awslabs/aws-lc/pull/413
Create issue with selection Actions

PR #178, opened 2022-03-02
Track AWS-LC as a git submodule

WillChilds-Klein:
*Issue #, if available:* n/a

*Description of changes:*

This change should allow for easier co-development of AWS-LC when
embedded in the ACCP source tree, as well as allowing us to track more
granular commits/rollbacks than if we were downloading releases.

# Testing

## Recursive Clone Builds

```
$ cd $(mktemp -d)

$ git clone --recurse-submodules ~/workplace/github/corretto/amazon-corretto-crypto-provider/
Cloning into 'amazon-corretto-crypto-provider'...
done.
Submodule 'aws-lc' (https://github.com/awslabs/aws-lc) registered for path 'aws-lc'
Cloning into '/private/var/folders/sn/gqzwgx256pb_7qzy7118wxr80000gs/T/tmp.CMVTWXKX/amazon-corretto-crypto-provider/aws-lc'...
remote: Enumerating objects: 79325, done.
remote: Counting objects: 100% (1915/1915), done.
remote: Compressing objects: 100% (954/954), done.
remote: Total 79325 (delta 1075), reused 1506 (delta 887), pack-reused 77410
Receiving objects: 100% (79325/79325), 115.67 MiB | 1.78 MiB/s, done.
Resolving deltas: 100% (56370/56370), done.
Submodule path 'aws-lc': checked out 'cb3f6d0c75f7708806f5e435c0c08578163e5103'

$ cd amazon-corretto-crypto-provider/

$ docker run -v $(pwd):/accp --rm -it accp:latest
root@06f679adcad6:/accp# ./gradlew release
...
(still building)
```

## Normal Clone Fails Build With Error Message

```
$ cd $(mktemp -d)

$ git clone ~/workplace/github/corretto/amazon-corretto-crypto-provider/
Cloning into 'amazon-corretto-crypto-provider'...
done.

$ cd amazon-corretto-crypto-provider/

$ docker run -v $(pwd):/accp --rm -it accp:latest
root@5e2946987412:/accp# ./gradlew release
Starting a Gradle Daemon (subsequent builds will be faster)
> Task :buildAwsLc FAILED

FAILURE: Build failed with an exception.

* Where:
Build file '/accp/build.gradle' line: 55

* What went wrong:
Execution failed for task ':buildAwsLc'.
> aws-lc dir empty! run 'git submodule update --init --recursive' to populate.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 33s
1 actionable task: 1 executed
```


build.gradle
```diff
@@ -53,35 +41,27 @@ dependencies {
...
-
-    doFirst {
-        exec {
-            workingDir tarFile.parent
-            commandLine 'sha256sum', '--check', "${projectDir}/awslc.sha256"
-        }
-
-        mkdir "${buildDir}/awslc/bin"
+// NOTE: implemented as a function because gradle didn't appear to block on git
+//       commands until completion when run as a task
```
SalusaSecondus, line 49: Shouldn't ensuring the sub-module is updated be the responsibility of the developer running git rather than gradle?

SalusaSecondus (commented):
All the tests are failing.

build.gradle
WillChilds-Klein: sort of. this command actually pulls the submodule code into the ACCP source tree. without running this command, `./aws-lc` is just an empty directory and ACCP won't build. so, without this logic, ACCP users would need to either clone the ACCP repo with `git clone --recursive` (which is non-standard), or they'd need to run the `git submodule` command(s) themselves. my goal here was to keep the same workflow as the curl/tar approach -- just run `./gradlew release` after cloning the repo and you're all set.

i suppose one disadvantage of the implementation as-written is that it introduces a dependency on `git` (which is why we're seeing the codebuild failures -- that build environment doesn't have git installed), but we already have external dependencies not modeled in `gradle.build` et al. (cmake, gradle, etc.) so this doesn't seem like a significant imposition.

of course, this submodule fetching logic is trivial to remove. let me know what you think.

SalusaSecondus: I think that we can just document that clones need to be recursive. It might make sense to check for an empty `aws-lc` directory in `build.gradle` and return a useful error message though.

WillChilds-Klein: alright. that should be easy enough to do. i'll post another revision.

WillChilds-Klein:
closing/re-opening to re-do code checks after [submodule fetching](https://docs.aws.amazon.com/codebuild/latest/APIReference/API_GitSubmodulesConfig.html) was enabled in CodeBuild.

Dockerfile.dev
```diff
@@ -12,6 +12,7 @@ RUN apt-get install -y lcov
 RUN apt-get install -y cmake
 RUN apt-get install -y dieharder
 RUN apt-get install -y build-essential
+RUN apt-get install -y git
```
SalusaSecondus, line 15: Is this still needed?

WillChilds-Klein: no, no it is not. thanks for catching.

SalusaSecondus (approved):
Failing tests are still expected on this branch.

Issue #139, opened 2021-01-25
Remove custom SecureRandom implementation


SalusaSecondus:
Now that OpenSSL supports a sufficiently high quality RNG, we can remove our internal implementation of AES-CTR-DRBG and just use theirs. This reduces the amount of code we need to maintain and also opens up a path to supporting `SecureRandom` on `aarch64`  (see #40) as we will depend on OpenSSL for proper (re)seeding as well. (Our manual seeding solution is X86_64 only.)

Links for reference:
* https://www.openssl.org/docs/man1.1.1/man3/RAND_DRBG_new.html
* https://www.openssl.org/docs/man1.1.1/man3/RAND_DRBG_generate.html
* https://www.openssl.org/docs/man1.1.1/man3/RAND_DRBG_get0_private.html (`RAND_DRBG_get0_private` after properly configuring the default in [our library initialization](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/csrc/loader.cpp#L28) may be completely sufficient. This strategy may also address #121 by removing some `ThreadLocal` state.)
* https://www.openssl.org/docs/man1.1.1/man3/RAND_DRBG_set_reseed_defaults.html (Both for defaults and to implement `SecureRandom.setSeed()` ([Default values appear to be every 256 calls and hour](https://github.com/openssl/openssl/blob/15f54941bd4c0324f1c8d7773db56004183a9cb7/providers/implementations/rands/drbg_local.h#L30-L31). This is likely fine for our use and will result in us not needing to override it. Number of calls is definitely fine. I might lower the time interval to a small number of minutes.)

PR #179, opened 2022-03-04
Add LibCryptoRng SecureRandom Implementation and remove AesCtrDrbg

alexw91:
*Issue #, if available:*
Prerequisite work needed for https://github.com/corretto/amazon-corretto-crypto-provider/issues/139

*Description of changes:*
This change adds a new SecureRandom implementation to ACCP that internally just calls AWS-LC's `RAND_bytes()` API. This RNG is seeded from an internal CPU jitter implementation on first use, and mixes in entropy from RDRAND on x86 platforms when available, or from system OS entropy when not. 


src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -84,6 +84,10 @@ private void buildServiceMap() {
                     .setSelfTest(AesCtrDrbg.SPI.SELF_TEST);
         }
 
+        addService("SecureRandom", "LibCryptoRng", "LibCryptoRng$SPI",
+                singletonMap("ThreadSafe", "true"), "LibCryptoRng")
+                .setSelfTest(LibCryptoRng.SPI.SELF_TEST);
+
```
SalusaSecondus, line 90: Can we also add an alias of `DEFAULT` to whatever our current default implementation is so that is will remain safe for people to always request the `DEFAULT` algorithm from our provider and know that they'll get something good?

src/com/amazon/corretto/crypto/provider/LibCryptoRng.java
```diff
@@ -0,0 +1,96 @@
...
+            return new SelfTestResult(SelfTestStatus.FAILED);
+        }
+
+        private static class NativeDrgbState extends NativeResource {
+            private NativeDrgbState(long ptr) {
+                super(ptr, LibCryptoRng::releaseState);
+            }
+        }
+
+        private static final ThreadLocal<NativeDrgbState> state_ = new ThreadLocal<NativeDrgbState>() {
```
SalusaSecondus, line 56: Rather than doing this with an anonymous inner class could you please use [ThreadLocal.withInitial()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ThreadLocal.html#withInitial(java.util.function.Supplier)) instead? My understanding is that it is nicer to some systems with multiple classloaders.

tst/com/amazon/corretto/crypto/provider/test/ServiceSelfTestMetaTest.java
```diff
@@ -88,6 +88,9 @@ public void whenDRBGSelfTestsFail_nothingIsVended() throws Throwable {
         sneakyInvoke(test, "forceFailure");
 
         for (Provider.Service service : accp.getServices()) {
+            if (service.getType().equals("SecureRandom")) {
+                continue;
+            }
```
SalusaSecondus, line 93: Why do we allow an instance of `SecureRandom` to be advertised when that is the exact thing which failed?

csrc/libcrypto_rng.h
```diff
@@ -0,0 +1,35 @@
...
+#include <openssl/evp.h>
+#include <openssl/rand.h>
+#include <vector>
+#include "generated-headers.h"
+#include "util.h"
+#include "compiler.h"
+
+namespace AmazonCorrettoCryptoProvider {
+
+class libcrypto_rng {
```
SalusaSecondus, line 16: We clearly don't need to track state anymore, so rather than have the complexity of a class and then needing to track this state on the Java side (which can be expensive), let's remove *all* native state and just call directly into `RAND_bytes`.

SalusaSecondus (changes requested):

src/com/amazon/corretto/crypto/provider/LibCryptoRng.java
```diff
@@ -0,0 +1,96 @@
...
+                }
+            }
+            return new SelfTestResult(SelfTestStatus.FAILED);
+        }
+
+        private static class NativeDrgbState extends NativeResource {
+            private NativeDrgbState(long ptr) {
+                super(ptr, LibCryptoRng::releaseState);
+            }
+        }
```
SalusaSecondus, line 54: This class is no longer needed as there is no native state to track anymore. Let's get rid of it and everything which uses it.

tst/com/amazon/corretto/crypto/provider/test/ServiceSelfTestMetaTest.java
alexw91: I'm not sure I follow? This test looks to me like it uses reflection to test the case where AesCtrDrbg fails it's self tests, and is asserting that all other ACCP services refuse to start up (because they presumably have an implicit dependency on the AesCtrDrbg entropy source?). Let me know if I'm misunderstanding.

However, the new LibCryptoRng is self-sufficient, and doesn't have such a dependency on AesCtrDrbg, and will successfully instantiate itself even if AesCtrDrbg fails. So this test was updated to to not require SecureRandom services fail to instantiate themselves if AesCtrDrbg fails.

src/com/amazon/corretto/crypto/provider/LibCryptoRng.java
alexw91: Fixed

csrc/libcrypto_rng.h
alexw91: Fixed

src/com/amazon/corretto/crypto/provider/LibCryptoRng.java
alexw91: Fixed

tst/com/amazon/corretto/crypto/provider/test/LibCryptoRngTest.java
```diff
@@ -0,0 +1,147 @@
...
+        rnd.setSeed(new byte[16]);
+        rnd.setSeed(new byte[20]);
+        rnd.setSeed(new byte[24]);
+        rnd.setSeed(new byte[32]);
+        rnd.setSeed(new byte[48]);
+        rnd.setSeed(new byte[64]);
+    }
+
+    @Test
+    public void largeRequest() {
```
torben-hansen, line 123: Assuming you want to test re-seeding in the underlying DRBG.

This is unlikely to cause a re-seed in AWS-LC. There are two relevant numbers:
1) max number of `generate` calls between reseeds, and
2) max number of bytes per `generate` call.

If (1) is reached, AWS-LC will re-seed the CTR-DRBG. For (2), if you request output larger than the maximum, AWS-LC will split the output generation into several calls to `generate`.

(1):
* non-fips `2^12`
* fips `2^24`
* https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/rand/rand.c#L62.

(2):
* `2^16` bytes https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/rand/internal.h#L86
* Splitting logic https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/rand/rand.c#L421

To exercise the different code-paths, assuming AWS-LC and CTR-DRBG, can modify test to:
* Re-seed: `x * 2^12` calls to `RAND_bytes()`, for some scaling factor `x`.
* Splitting: input size requests (strictly) larger than `2^16` bytes.

torben-hansen: The latter doesn't seem relevant though, since you already have your own splitting logic for much smaller requests
```
private static final int MAX_SINGLE_REQUEST = 8192;
```

csrc/libcrypto_rng.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+using namespace AmazonCorrettoCryptoProvider;
+
+
+bool libCryptoRngGenerateRandomBytes(uint8_t *buf, int len) noexcept {
+    /**
+     * AWS LibCrypto provides a thread local, lazily initialized, FIPS Validated DRBG that is seeded with CPU Jitter
+     * entropy on first use. This API also mixes in more entropy from the fastest available source after every call.
+     * If available it will use x86 RDRAND instruction, or otherwise use the OS system entropy (Eg /dev/urandom/) to
+     * keep prediction resistance.
```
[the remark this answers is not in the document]
alexw91: Fixed

tst/com/amazon/corretto/crypto/provider/test/LibCryptoRngTest.java
```diff
@@ -0,0 +1,147 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+import org.junit.jupiter.api.parallel.ResourceAccessMode;
+import org.junit.jupiter.api.parallel.ResourceLock;
+
+@ExtendWith(TestResultLogger.class)
+@Execution(ExecutionMode.CONCURRENT)
+@ResourceLock(value = TestUtil.RESOURCE_GLOBAL, mode = ResourceAccessMode.READ)
+public class LibCryptoRngTest {
```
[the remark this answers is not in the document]
alexw91: Fixed

csrc/libcrypto_rng.cpp
```diff
@@ -0,0 +1,127 @@
...
+     * AWS-LC will lazily initialize itself on first use. There is no API to force the RNG to initialize itself.
+     *
+     * Keep this method stub in case different initialization logic is needed for different LibCrypto's.
+     */
+    initialized = true;
+}
+
+bool libcrypto_rng::generateRandomBytes(uint8_t *buf, int len) noexcept {
+    /**
+     * AWS LibCrypto provides a thread local, lazily initialized, FIPS Validated DRBG that is seeded with CPU Jitter
```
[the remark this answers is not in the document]
alexw91: Fixed

tst/com/amazon/corretto/crypto/provider/test/LibCryptoRngTest.java
```diff
@@ -0,0 +1,147 @@
...
+                        "Check array size " + size + " position " + x + " is equal to zero");
+            }
+        }
+    }
+
+    // A common mistake it when filling arrays to not do it properly and leave
+    // zero gaps. To detect this, we'll generate arrays of different lengths
+    // and check to see if certain bytes are always zero
+    @Test
+    public void testGenerateSeed() {
```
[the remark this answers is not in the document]
alexw91: I'm going to leave that as-is since I didn't write that test. It's copied from the other DRBG test in ACCP. 
 - https://github.com/corretto/amazon-corretto-crypto-provider/blob/80b3fa918a4787dd5d1d4e3b2451f7e0e5ae5aad/tst/com/amazon/corretto/crypto/provider/test/AesCtrDrbgTest.java#L122

torben-hansen (approved):
lgtm.

Dunno if I should approve anything - but sending a nominally thumbs up at least.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -84,6 +84,10 @@ private void buildServiceMap() {
                     .setSelfTest(AesCtrDrbg.SPI.SELF_TEST);
         }
 
+        addService("SecureRandom", "LibCryptoRng", "LibCryptoRng$SPI",
+                singletonMap("ThreadSafe", "true"), "DEFAULT")
+                .setSelfTest(LibCryptoRng.SPI.SELF_TEST);
+
```
SalusaSecondus, line 90: FYI, but putting this at the end it has "NIST800-90A/AES-CTR-256" be the default algorithm for `SecureRandom`. However, since we're planning on removing that anyway later, I think it's fine.

tst/com/amazon/corretto/crypto/provider/test/LibCryptoRngTest.java
```diff
@@ -0,0 +1,209 @@
...
+
+        // Ensure that we can successfully reseed the RNG a few times
+        int RESEED_COUNT = 3;
+        for(long i = 0; i < (RESEED_COUNT * MAX_REQUESTS_BEFORE_RESEED) + 1; i++) {
+            rng.nextBytes(rngOutputBuffer);
+        }
+    }
+
+    @Test
+    public void forkTest() throws Throwable {
```
SalusaSecondus, line 166: This isn't technically forking as that isn't possible from within Java. This is just multi-threading.

Also, as both `sharedRngOutputs` and `invididualRngOutputs` are not volatile (or arrays of volatile longs) I'm not certain this test gives us the confidence we need.

However, you should create new versions of the following targets in [CMakeLists.txt](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/CMakeLists.txt) which does relatively aggressive multi-threading logic to stress test implementations as well throws the `dieharder` suites at our RNGs:
* [run-dieharder](https://github.com/corretto/amazon-corretto-crypto-provider/blob/80b3fa918a4787dd5d1d4e3b2451f7e0e5ae5aad/CMakeLists.txt)
* [run-dieharder-tail](https://github.com/corretto/amazon-corretto-crypto-provider/blob/80b3fa918a4787dd5d1d4e3b2451f7e0e5ae5aad/CMakeLists.txt#L808-L815)
* [run-dieharder-threads](https://github.com/corretto/amazon-corretto-crypto-provider/blob/80b3fa918a4787dd5d1d4e3b2451f7e0e5ae5aad/CMakeLists.txt#L835-L842)
* [run-dieharder-threads-tail](https://github.com/corretto/amazon-corretto-crypto-provider/blob/80b3fa918a4787dd5d1d4e3b2451f7e0e5ae5aad/CMakeLists.txt#L844-L851)

The new tasks would be added to [`check-dieharder` and `check-dieharder-threads`](https://github.com/corretto/amazon-corretto-crypto-provider/blob/80b3fa918a4787dd5d1d4e3b2451f7e0e5ae5aad/CMakeLists.txt#L871-L895). While this will significantly increase the runtime of `overkill`, that will only be temporary until we can remove the older random number generator.

SalusaSecondus, line 166: This isn't technically forking as that isn't possible from within Java. This is just multi-threading.

Also, as both `sharedRngOutputs` and `invididualRngOutputs` are not volatile (or arrays of volatile longs) I'm not certain this test gives us the confidence we need.

However, you should create new versions of the following targets in [CMakeLists.txt](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/CMakeLists.txt) which does relatively aggressive multi-threading logic to stress test implementations as well throws the `dieharder` suites at our RNGs:
* [run-dieharder](https://github.com/corretto/amazon-corretto-crypto-provider/blob/80b3fa918a4787dd5d1d4e3b2451f7e0e5ae5aad/CMakeLists.txt)
* [run-dieharder-tail](https://github.com/corretto/amazon-corretto-crypto-provider/blob/80b3fa918a4787dd5d1d4e3b2451f7e0e5ae5aad/CMakeLists.txt#L808-L815)
* [run-dieharder-threads](https://github.com/corretto/amazon-corretto-crypto-provider/blob/80b3fa918a4787dd5d1d4e3b2451f7e0e5ae5aad/CMakeLists.txt#L835-L842)
* [run-dieharder-threads-tail](https://github.com/corretto/amazon-corretto-crypto-provider/blob/80b3fa918a4787dd5d1d4e3b2451f7e0e5ae5aad/CMakeLists.txt#L844-L851)

The new tasks would be added to [`check-dieharder` and `check-dieharder-threads`](https://github.com/corretto/amazon-corretto-crypto-provider/blob/80b3fa918a4787dd5d1d4e3b2451f7e0e5ae5aad/CMakeLists.txt#L871-L895). While this will significantly increase the runtime of `overkill`, that will only be temporary until we can remove the older random number generator.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
alexw91: Fixed in latest commit

tst/com/amazon/corretto/crypto/provider/test/LibCryptoRngTest.java
alexw91: Fixed in latest commit

alexw91: Fixed in latest commit

SalusaSecondus (changes requested):
I'm seeing lots of failures in unit tests of the form:
```
  JUnit Jupiter:EvpSignatureTest:verifyExtendedMessage(TestParams):[323] SHA384withECDSA length 2048. Read-only: false, Sliced: true
    MethodSource [className = 'com.amazon.corretto.crypto.provider.test.EvpSignatureTest', methodName = 'verifyExtendedMessage', methodParameterTypes = 'com.amazon.corretto.crypto.provider.test.EvpSignatureTest$TestParams']
    => java.security.SignatureException: error:1a000064:ECDSA routines:OPENSSL_internal:BAD_SIGNATURE
       com.amazon.corretto.crypto.provider.EvpSignature.verifyFinish(Native Method)
       com.amazon.corretto.crypto.provider.EvpSignature.lambda$engineVerify$20(EvpSignature.java:545)
       com.amazon.corretto.crypto.provider.NativeResource$Cell.use(NativeResource.java:88)
       com.amazon.corretto.crypto.provider.NativeResource.use(NativeResource.java:127)
       com.amazon.corretto.crypto.provider.EvpSignature.lambda$engineVerify$21(EvpSignature.java:545)
       com.amazon.corretto.crypto.provider.InputBuffer.doFinal(InputBuffer.java:601)
       com.amazon.corretto.crypto.provider.EvpSignature.engineVerify(EvpSignature.java:574)
       com.amazon.corretto.crypto.provider.EvpSignature.engineVerify(EvpSignature.java:520)
       java.base/java.security.Signature$Delegate.engineVerify(Signature.java:1416)
       java.base/java.security.Signature.verify(Signature.java:790)
       [...]
```
It looks like [this line](https://github.com/corretto/amazon-corretto-crypto-provider/blob/fd967b00839fad0651d4cae77625cb5ff85d9896/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureTest.java#L395) is throwing an exception when (for the message being wrong but everything else being correct) it should definitely just be returning `false`.

SalusaSecondus:
Oh, it looks like the dieharder tests are now timing out. Since we're removing the old `SecureRandom` before we actually cut a release, please just remove the existing dieharder tests for the old instance of secure random (to bring our runtimes back down to a reasonable level).

tst/com/amazon/corretto/crypto/provider/test/LibCryptoRngTest.java
```diff
@@ -41,6 +47,21 @@ public void teardown() {
...
+                return;
+            }
+        }
+        fail("Failed to find a different value");
+    }
+
+    @Test
+    public void testGetInstance() throws NoSuchAlgorithmException {
+        ensureRngGeneratesUniqueValues(SecureRandom.getInstance("DEFAULT", AmazonCorrettoCryptoProvider.INSTANCE));
+        ensureRngGeneratesUniqueValues(SecureRandom.getInstance("LibCryptoRng", AmazonCorrettoCryptoProvider.INSTANCE));
```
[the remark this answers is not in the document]
alexw91: Fixed

tst/com/amazon/corretto/crypto/provider/test/LibCryptoRngTest.java
```diff
@@ -0,0 +1,215 @@
...
+
+@ExtendWith(TestResultLogger.class)
+@Execution(ExecutionMode.CONCURRENT)
+@ResourceLock(value = TestUtil.RESOURCE_GLOBAL, mode = ResourceAccessMode.READ)
+public class LibCryptoRngTest {
+    private LibCryptoRng rnd;
+
+    @BeforeEach
+    public void setup() {
+        rnd = new LibCryptoRng();
```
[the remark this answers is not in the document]
alexw91: Fixed

WillChilds-Klein (approved):

src/com/amazon/corretto/crypto/provider/LibCryptoRng.java
```diff
@@ -0,0 +1,72 @@
...
+            final byte[] seed = new byte[numBytes];
+            engineNextBytes(seed);
+            return seed;
+        }
+
+        @Override
+        protected void engineNextBytes(byte[] bytes) {
+            int offset = 0;
+            while (offset < bytes.length) {
+                final int toGenerate = Math.min(MAX_SINGLE_REQUEST, bytes.length - offset);
```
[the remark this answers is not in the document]
alexw91: This value was taken from [the previous AesCtrDrbg implementation](https://github.com/corretto/amazon-corretto-crypto-provider/blob/80b3fa918a4787dd5d1d4e3b2451f7e0e5ae5aad/src/com/amazon/corretto/crypto/provider/AesCtrDrbg.java#L22), which I think was taken from [page 49 of the NIST SP800-90A document](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-90Ar1.pdf#page=58). I kept the 8k limit to be on the safe side, but it may not be needed since this RNG doesn't internally use AES anymore.

src/com/amazon/corretto/crypto/provider/LibCryptoRng.java
```diff
@@ -0,0 +1,72 @@
...
+    static class SPI extends SecureRandomSpi {
+        private static final long serialVersionUID = 1L;
+        static final SelfTestSuite.SelfTest SELF_TEST = new SelfTestSuite.SelfTest(ALGORITHM_NAME, LibCryptoRng.SPI::runSelfTest);
+
+        static SelfTestResult runSelfTest() {
+            LibCryptoRng rnd = new LibCryptoRng();
+
+            /* Basic RNG self-test, do we generate different random numbers on repeat calls? */
+            final long initialLong = rnd.nextLong();
+            for (int trial = 0; trial < 10; trial++) {
```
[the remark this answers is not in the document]
alexw91: This self-test was taken from the existing AesCtrDrbg test suite [here](https://github.com/corretto/amazon-corretto-crypto-provider/blob/80b3fa918a4787dd5d1d4e3b2451f7e0e5ae5aad/tst/com/amazon/corretto/crypto/provider/test/AesCtrDrbgTest.java#L155-L160) which used 10 trail rounds before asserting the RNG had failed. 

I agree that 10 rounds is excessive, updated PR with a lower value of 3 trail rounds.

SalusaSecondus (approved):
LGTM

PR #181, opened 2022-03-21
Switch to OpenSSL 1.1.1n

SalusaSecondus:
*Description of changes:* Switch to OpenSSL 1.1.1n

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 1.7.0 (Unreleased)
 ### Improvements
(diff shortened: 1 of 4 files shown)


WillChilds-Klein (approved):

SalusaSecondus:
Dieharder tests are (by their nature) flaky. The failure seen above isn't worrisome and matches the type of flaky issue we've seen before. *If* it is real then we'll also see failures in `overkill` and later when we try to merge things into `master.

After out-of-band discussion with @WillChilds-Klein, we will proceed with merging.

PR #183, opened 2022-03-23
Crt params

WillChilds-Klein:
# Status

This PR will remain in "draft" status until [AWS-LC PR #436](https://github.com/awslabs/aws-lc/pull/436) is merged.

# Disabling RSA Blinding on Stripped Kyes

By default AWS-LC enables RSA blinding, and forces callers to set a flag in order to disable it. Since RSA blinding is impossible with stripped keys, this commit sets that flag where appropriate when we detect stripped RSA keys.

# Dockerfile.dev Improvements

Corretto 11 installed per [these instructions][1]. We also add comments
around development guidance and adds 2 development "dependencies" that
aren't strictly required for building/running ACCP, but facilitate ACCP
and AWS-LC development.

[1]: https://docs.aws.amazon.com/corretto/latest/corretto-11-ug/generic-linux-install.html

# Issues
- CryptoAlg-1099
- CryptoAlg-1022

---


SalusaSecondus:
Do we also need to add this logic here? https://github.com/corretto/amazon-corretto-crypto-provider/blob/aws-lc/csrc/keyutils.cpp#L43-L85

(This might replace the logic in `sign.cpp`)

WillChilds-Klein:
ah, nice. yeah that's a better place for the EVP blinding logic. i'll move it over there from `sign.cpp`.

SalusaSecondus:
Looks good. I expect that I'll be able to quickly approve this once it's ready for review.

PR #186, opened 2022-03-28
Fix static linkage

SalusaSecondus:
*Description of changes:*
The current aws-lc branch doesn't properly do static linking but doesn't break dynamic linking enough for tests to find it. (We should fix that but I'm not sure how yet.) This fixes static link so that it all works again.

When we switch to dynamic linking for FIPS, we'll need to approach it more intentionally.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -290,7 +290,7 @@ set(OLD_CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS}")
 ## command line while building a _shared_ library. This works only on certain
 ## platforms, but if successful it gives us a private copy of libcrypto that we
 ## can be sure will not interfere with any other openssl users.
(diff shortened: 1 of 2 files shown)


WillChilds-Klein:
2 of the integration tests are failing, which i don't believe we've seen with the other PRs on this branch:

```
Unit Jupiter:ExternalHTTPSIntegrationTest:testHTTPSConnectivity(boolean, String):HTTPS integration test: URL=https://rsa8192.badssl.com/ BCEnabled=false
MethodSource [className = 'com.amazon.corretto.crypto.provider.test.integration.ExternalHTTPSIntegrationTest', methodName = 'testHTTPSConnectivity', methodParameterTypes = 'boolean, java.lang.String']
javax.net.ssl.SSLHandshakeException: PKIX path building failed: 
sun.security.provider.certpath.SunCertPathBuilderException: unable to 
find valid certification path to requested target</span></span>
```

i'm assuming this is due to `badssl.com` being bad (unsurprisingly) and allowing the cert to expire, as you pointed out in slack earlier today. PR approved.

WillChilds-Klein (approved):

SalusaSecondus:
Yup, https://rsa8192.badssl.com/ has an expired certificate. That said, they just refreshed the certificate so I'm going to close and re-open to retrigger the workflow.

PR #188, opened 2022-04-11
Changed generateEcKeyFromSpec to use EVP APIs

dkostic:
*Issue #, if available:*
N/A

*Description of changes:*
Function `generateEcKeyFromSpec` is modified to use libcrypto's EVP APIs.

### DIFF
--- a/csrc/ec_gen.cpp
+++ b/csrc/ec_gen.cpp
@@ -159,11 +159,14 @@ JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_EcGen_generateEc
   jbyteArray yArr,
   jbyteArray sArr)
 {
-    // TODO, figure out how to do this with EVP
     std::vector<uint8_t, SecureAlloc<uint8_t> > derBuf;
-    EC_KEY* ecParams = NULL;
+    EC_KEY *ecParams = NULL;
     EC_KEY *key = NULL;
     const EC_GROUP* group;
+
+    EVP_PKEY_CTX *keyCtx = NULL;
+    EVP_PKEY_CTX *tmpCtx = NULL;
+    EVP_PKEY *pkey = NULL;
 
     try {
         raii_env env(pEnv);
@@ -183,16 +186,17 @@ JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_EcGen_generateEc
         // Now that we have the params, extract the group from them (all we care about)
         CHECK_OPENSSL(group = EC_KEY_get0_group(ecParams));
 
-        // Build the structure which will hold our result
-        if(!likely(key = EC_KEY_new())) {
-            throw java_ex(EX_OOM, "Out of memory");
-        }
+        // Setup the temporary PKEY context with the extracted parameters.
+        CHECK_OPENSSL(tmpCtx = EVP_PKEY_CTX_new_id(EVP_PKEY_EC, NULL));
+        CHECK_OPENSSL(EVP_PKEY_paramgen_init(tmpCtx) == 1);
+        CHECK_OPENSSL(EVP_PKEY_CTX_set_ec_paramgen_curve_nid(tmpCtx, EC_GROUP_get_curve_name(group)) == 1);
+        CHECK_OPENSSL(EVP_PKEY_paramgen(tmpCtx, &pkey) == 1);
 
-        CHECK_OPENSSL(EC_KEY_set_group(key, group));
-
-        if (!likely(EC_KEY_generate_key(key))) {
-            throw_openssl("Unable to generate key");
-        }
+        // Generate the actual EC key.
+        CHECK_OPENSSL(keyCtx = EVP_PKEY_CTX_new(pkey, NULL));
+        CHECK_OPENSSL(EVP_PKEY_keygen_init(keyCtx) == 1);
+        CHECK_OPENSSL(EVP_PKEY_keygen(keyCtx, &pkey) == 1);
+        CHECK_OPENSSL(key = EVP_PKEY_get1_EC_KEY(pkey));
 
         if (checkConsistency) {
             CHECK_OPENSSL(EC_KEY_check_key(key) == 1);
@@ -203,6 +207,9 @@ JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_EcGen_generateEc
         ex.throw_to_java(pEnv);
     }
 
-    EC_KEY_free(ecParams);
     EC_KEY_free(key);
+    EC_KEY_free(ecParams);
+    EVP_PKEY_free(pkey);
+    EVP_PKEY_CTX_free(tmpCtx);
+    EVP_PKEY_CTX_free(keyCtx);
 }


csrc/ec_gen.cpp
```diff
@@ -183,16 +186,17 @@ JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_EcGen_generateEc
...
-
-        CHECK_OPENSSL(EC_KEY_set_group(key, group));
+        // Setup the temporary PKEY context with the extracted parameters.
+        CHECK_OPENSSL(tmpCtx = EVP_PKEY_CTX_new_id(EVP_PKEY_EC, NULL));
+        CHECK_OPENSSL(EVP_PKEY_paramgen_init(tmpCtx) == 1);
+        CHECK_OPENSSL(EVP_PKEY_CTX_set_ec_paramgen_curve_nid(tmpCtx, EC_GROUP_get_curve_name(group)) == 1);
+        CHECK_OPENSSL(EVP_PKEY_paramgen(tmpCtx, &pkey) == 1);
 
-        if (!likely(EC_KEY_generate_key(key))) {
-            throw_openssl("Unable to generate key");
```
WillChilds-Klein, line 194: previously, on malloc failure and keygen failure, we would throw an exception up to the java code. what happens when a call to `CHECK_OPENSSL` fails? will that also result in an exception being thrown, or will the failure manifest in some other way?

csrc/ec_gen.cpp
```diff
@@ -159,12 +159,15 @@ JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_EcGen_generateEc
...
-    // TODO, figure out how to do this with EVP
     std::vector<uint8_t, SecureAlloc<uint8_t> > derBuf;
-    EC_KEY* ecParams = NULL;
+    EC_KEY *ecParams = NULL;
     EC_KEY *key = NULL;
     const EC_GROUP* group;
 
+    EVP_PKEY_CTX *keyCtx = NULL;
+    EVP_PKEY_CTX *tmpCtx = NULL;
+    EVP_PKEY *pkey = NULL;
```
SalusaSecondus, line 169: When possible we prefer to use [EvpKeyContext](https://github.com/corretto/amazon-corretto-crypto-provider/blob/fd967b00839fad0651d4cae77625cb5ff85d9896/csrc/keyutils.h#L14) which allows us to track these structures on the stack and know that they will be automatically freed as soon as the tracking object goes out of scope. This will allow you to remove all of the free statements except for `EC_KEY_free()` at the bottom.

Also, if you were to rename `tmpCtx` to `paramCtx`, it would be much clearer.

So I'd probably do:
```suggestion
EvpKeyContext keyCtx;
EvpKeyContext paramCtx;
```

WillChilds-Klein:
it looks like this PR is [forked](https://github.com/dkostic/amazon-corretto-crypto-provider/tree/generateEcKeyFromSpec-with-EVP-APIs) from the `develop` branch, which currently [uses OpenSSL's libcrypto](https://github.com/dkostic/amazon-corretto-crypto-provider/blob/generateEcKeyFromSpec-with-EVP-APIs/build.gradle#L40). we have [a separate `aws-lc` branch](https://github.com/corretto/amazon-corretto-crypto-provider/tree/aws-lc) that links against AWS-LC. i think it would be fine to merge this PR against `develop` (at which point i can cherry-pick it over to `aws-lc`), but i wanted to confirm -- we don't expect any behavioral difference in the EC EVP APIs between OpenSSL and AWS-LC, right?

csrc/ec_gen.cpp
```diff
@@ -183,16 +186,17 @@ JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_EcGen_generateEc
...
+        CHECK_OPENSSL(EVP_PKEY_paramgen(tmpCtx, &pkey) == 1);
 
-        if (!likely(EC_KEY_generate_key(key))) {
-            throw_openssl("Unable to generate key");
-        }
+        // Generate the actual EC key.
+        CHECK_OPENSSL(keyCtx = EVP_PKEY_CTX_new(pkey, NULL));
+        CHECK_OPENSSL(EVP_PKEY_keygen_init(keyCtx) == 1);
+        CHECK_OPENSSL(EVP_PKEY_keygen(keyCtx, &pkey) == 1);
+        CHECK_OPENSSL(key = EVP_PKEY_get1_EC_KEY(pkey));
```
SalusaSecondus, line 199: Can we switch to `EVP_PKEY_get0_EC_KEY()` which removes the need to free the returned `EC_KEY` object? (That API was introduced after ACCP launched on 1.0.x, so wasn't available originally.)

csrc/ec_gen.cpp
```diff
@@ -203,6 +207,9 @@ JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_EcGen_generateEc
         ex.throw_to_java(pEnv);
     }
 
-    EC_KEY_free(ecParams);
     EC_KEY_free(key);
+    EC_KEY_free(ecParams);
```
SalusaSecondus, line 211: Please remove the comment related to freeing this object. It's definitely out of date and doesn't make sense to me anymore.

SalusaSecondus (changes requested):

csrc/ec_gen.cpp
```diff
@@ -159,12 +159,15 @@ JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_EcGen_generateEc
   jbyteArray yArr,
   jbyteArray sArr)
 {
-    // TODO, figure out how to do this with EVP
```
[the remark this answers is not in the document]
SalusaSecondus: EVP *might* be slightly slower due to more levels of memory management and pointers, but it is unlikely to matter much relative to the cost of generating these keys.

The motivation is general code quality concerns. Moving off of the old APIs to the newer ones makes future upgrades and improvements easier (as the new APIs are the ones getting all of the attention).

Also, when this comment was made it was unclear if the direct APIs would be permitted when running against the FIPS-validated module and we were certain that the EVP APIs would be.

Finally, #132 moves all asymmetric cryptography over to the EVP APIs (for performance reasons) and avoided extra conversions between the two will be good.

dkostic:
> it looks like this PR is [forked](https://github.com/dkostic/amazon-corretto-crypto-provider/tree/generateEcKeyFromSpec-with-EVP-APIs) from the develop branch, which currently [uses OpenSSL's libcrypto](https://github.com/dkostic/amazon-corretto-crypto-provider/blob/generateEcKeyFromSpec-with-EVP-APIs/build.gradle#L40). we have [a separate aws-lc branch](https://github.com/corretto/amazon-corretto-crypto-provider/tree/aws-lc) that links against AWS-LC. i think it would be fine to merge this PR against develop (at which point i can cherry-pick it over to aws-lc), but i wanted to confirm -- we don't expect any behavioral difference in the EC EVP APIs between OpenSSL and AWS-LC, right?

yeah, sorry about that, I actually made the change on `aws-lc` branch and cherry-picked it to `develop` :)
I'd rather re-submit the PR to the `aws-lc` branch because that's where I tested the change and also I know which APIs aws-lc offers and how they work (for OpenSSL I can't guarantee anything because I haven't worked much with it). So if you don't mind, I'll address your comments and submit a new PR (or force push if that's possible)?

dkostic:
As we agreed, I opened a new PR, closing this one in favor of #199

PR #190, opened 2022-04-12
Ignore expired certs in tests

SalusaSecondus:
*Description of changes:*
External integration tests now skip certificate validation for expired certificates. This is to work around external sites which may have allowed their certificates to expire.


WillChilds-Klein:
so, how do we notice when tests are skipping due to cert expiration? this feels like it could open the door to regressions creeping in around cert validations...

SalusaSecondus:
The logs are how we'd notice it (ideally).

We're still using the certificates which is where the most interesting cryptography is. Validating the signature on them is relatively boring.

PR #195, opened 2022-04-19
Add invalid ecdsa r,s value tests

SalusaSecondus:
*Description of changes:*

Add explicit tests for ECDSA signatures containing `r` and `s` values outside of the valid range `[1, order)`.

### DIFF
--- a/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
@@ -10,7 +10,10 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
+import java.util.List;
+import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
 import java.security.GeneralSecurityException;
 import java.security.InvalidAlgorithmParameterException;
@@ -24,24 +27,31 @@ import java.security.Provider;
 import java.security.PublicKey;
 import java.security.Signature;
 import java.security.SignatureException;
+import java.security.interfaces.ECPublicKey;
 import java.security.interfaces.RSAPrivateKey;
 import java.security.spec.AlgorithmParameterSpec;
 import java.security.spec.ECGenParameterSpec;
 import java.security.spec.PSSParameterSpec;
 import java.security.spec.RSAKeyGenParameterSpec;
 import java.security.spec.RSAPrivateKeySpec;
+import java.util.ArrayList;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider;
 import org.bouncycastle.jce.provider.BouncyCastleProvider;
+import org.bouncycastle.util.Arrays;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.api.parallel.Execution;
 import org.junit.jupiter.api.parallel.ExecutionMode;
 import org.junit.jupiter.api.parallel.ResourceAccessMode;
 import org.junit.jupiter.api.parallel.ResourceLock;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 import javax.crypto.AEADBadTagException;
 import javax.crypto.Cipher;
@@ -54,11 +64,11 @@ import javax.crypto.spec.SecretKeySpec;
 @ExtendWith(TestResultLogger.class)
 @Execution(ExecutionMode.CONCURRENT)
 @ResourceLock(value = TestUtil.RESOURCE_GLOBAL, mode = ResourceAccessMode.READ)
-public class EvpSignatureSpecificTest {
+public final class EvpSignatureSpecificTest {
     private static final BouncyCastleProvider BOUNCYCASTLE_PROVIDER = new BouncyCastleProvider();
     private static final byte[] MESSAGE = new byte[513];
-    private final static KeyPair RSA_PAIR;
-    private final static KeyPair ECDSA_PAIR;
+    private static final KeyPair RSA_PAIR;
+    private static final KeyPair ECDSA_PAIR;
 
     static {
         for (int x = 0; x < MESSAGE.length; x++) {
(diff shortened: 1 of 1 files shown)


tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -98,6 +138,48 @@ private static void testKeyTypeMismatch(final String algorithm, final String bas
...
+    @MethodSource("ecdsaBadSignatures")
+    @SuppressWarnings("unchecked")
+    public void testBadEcdsaSignature(final String algorithm, final BigInteger r, final BigInteger s)
+            throws GeneralSecurityException {
+        final Signature verifier = Signature.getInstance(algorithm, NATIVE_PROVIDER);
+        // We always use the largest key to avoid hash truncation confusion
+        final ECPublicKey pubKey = (ECPublicKey) ECDSA_PAIR.getPublic();
+        final byte[] signature;
+        byte[] rArr = r.toByteArray();
+        if (rArr[0] == 0) { // Trim leading zero byte if present
```
WillChilds-Klein, line 151: why does this (sometimes?) have a leading zero? is that an ASN.1 thing?

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -98,6 +138,48 @@ private static void testKeyTypeMismatch(final String algorithm, final String bas
...
+        }
+
+        if (algorithm.endsWith("inP1363Format")) {
+            // Just zeros of the appropriate length
+            final int elementLength = (pubKey.getParams().getOrder().bitLength() + 7) / 8;
+            signature = new byte[elementLength * 2];
+            System.arraycopy(rArr, 0, signature, elementLength - rArr.length, rArr.length);
+            System.arraycopy(sArr, 0, signature, 2 * elementLength - sArr.length, sArr.length);
+        } else {
+            // It isn't worth building full ASN.1 logic here, so we cover the easy cases of ones and zeros
```
WillChilds-Klein, line 166: doesn't BouncyCastle have [facilities](https://people.eecs.berkeley.edu/~jonah/bc/org/bouncycastle/asn1/package-summary.html) for this?

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -98,6 +138,48 @@ private static void testKeyTypeMismatch(final String algorithm, final String bas
...
+            // and leave the complicated cases of the full orders to the P1363 tests
+            // (which cover the same implementations) under the hood
+            assumeTrue(rArr.length == 1 && sArr.length == 1, "This test only supports {0,1} for non-P1363 sigs.");
+            signature = new byte[]{0x30, 0x06, 0x02, 0x01, 0x00, 0x02, 0x01, 0x00};
```
WillChilds-Klein, line 170: what are these magic values?

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -98,6 +138,48 @@ private static void testKeyTypeMismatch(final String algorithm, final String bas
...
+            signature[4] = rArr[0];
+            signature[7] = sArr[0];
+
+        }
+        verifier.initVerify(pubKey);
+        try {
+            assertFalse(verifier.verify(signature));
+        } catch (final SignatureException ex) {
+            // We also allow a SignatureException.
```
WillChilds-Klein, line 179: for an invalid signature (as opposed to a mismatched but otherwise valid signature), shouldn't we always be throwing SignatureException? from [oracle's javadoc](https://docs.oracle.com/javase/7/docs/api/java/security/Signature.html#verify(byte[])):

> Returns:
>     true if the signature was verified, false if not.
> Throws:
>     SignatureException - if this signature object is not initialized properly, the passed-in signature is improperly encoded or of the wrong type, if this signature algorithm is unable to process the input data provided, etc.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
SalusaSecondus: `BigInteger` returns signed values and these values are all positive, so must have a signed zero bit. Thus if the order of the curve fits exactly into a multiple of 8 bits, the leading one bit of the order with be the first bit of a byte, and thus when we add a sign bit of 0, it overflows to a new byte.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
SalusaSecondus: Yeah, and we do have a dependency of BC for our tests, but I really just use it for cryptography. The math is the same between these two algorithms and only the signature parsing varies. I don't think it is worth the complexity here.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
SalusaSecondus: That's an ASN.1 encoded sequence of two zero values.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
SalusaSecondus: The exact lines of what is an invalid signature are unclear and not well specified. In fact you can see that [JDK15+ returns false](https://github.com/corretto/corretto-17/commit/2d4103a3d929e05edca98e7703e0869077966be7#diff-af656439ff8511c928a2107a4806c6b669c8851db55d30d9092cdd96db84be56R227) rather than throws an exception in this case. All that matters is that we reject it. How is far less important.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
WillChilds-Klein: fair enough.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
WillChilds-Klein: makes sense, thanks.  last piece of the puzzle for me was this [javadoc](https://docs.oracle.com/javase/7/docs/api/java/math/BigInteger.html#toByteArray()) stating that `toByteArray` always returns big-endian, 2's complement byte array.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
WillChilds-Klein: I see, thanks for the context. Because the JCA interfaces go to the trouble of stipulating different failure modes (throwing vs. returning), I would have expected that they precisely specify which error/failure cases correspond to throwing vs. returning. I suppose it's reasonable to rely on a reference implementation absent something more formal.

WillChilds-Klein (approved):

PR #199, opened 2022-05-05
Changed generateEcKeyFromSpec to use EVP APIs

dkostic:
*Issue #, if available:*
N/A

*Description of changes:*
Changed `generateEcKeyFromSpec` function to use higher level
EVP APIs, instead of calling directly into EC_KEY functions.

### DIFF
--- a/csrc/ec_gen.cpp
+++ b/csrc/ec_gen.cpp
@@ -10,6 +10,7 @@
 #include "env.h"
 #include "bn.h"
 #include "buffer.h"
+#include "keyutils.h"
 
 using namespace AmazonCorrettoCryptoProvider;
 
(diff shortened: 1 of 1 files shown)


dkostic:
@WillChilds-Klein to answer the question you left on the previous PR regarding `CHECK_OPENSSL` -- the macro is defined to throw an exception on error (https://github.com/corretto/amazon-corretto-crypto-provider/blob/fd967b00839fad0651d4cae77625cb5ff85d9896/csrc/env.h#L127)

dkostic:
@SalusaSecondus I changed the implementation to use `EvpKeyContext` as you suggested. Let me know if that's what you meant or I completely missed the point.

csrc/ec_gen.cpp
```diff
@@ -183,16 +180,29 @@ JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_EcGen_generateEc
...
+        // Generate the actual EC key.
+        if (!keyCtx.setKeyCtx(EVP_PKEY_CTX_new(paramCtx.getKey(), NULL))) {
+            throw_openssl("Unable to create new evp key context");
         }
+        CHECK_OPENSSL(EVP_PKEY_keygen_init(keyCtx.getKeyCtx()) == 1);
+        // We need to define a temporary pointer to an EVP_PKEY
+        // to be able to call EVP_PKEY_keygen function.
+        EVP_PKEY *tmpKey1 = NULL;
+        CHECK_OPENSSL(EVP_PKEY_keygen(keyCtx.getKeyCtx(), &tmpKey1) == 1);
+        keyCtx.setKey(tmpKey1);
```
WillChilds-Klein, line 204: why do we not need to free `tmpKey0` or `tmpKey1`? does `setKey` transfer ownership to the EvpKeyContext struct?

dkostic: yes, I think that's the reason Greg prefers to use `EvpKeyContext`

dkostic:
not sure if I'm missing something, but it looks to me that the latest change merged in `develop` makes this PR redundant. Would you agree with that @SalusaSecondus ?

SalusaSecondus:
Agreed. This is now redundant.

dkostic:
thanks Greg, closing the PR.

PR #202, opened 2022-05-13
 Support PSS padding for RSA signatures

WillChilds-Klein:
*Issue #, if available:*

CryptoAlg-1036

*Description of changes:*

# Status
- PR ready to review
- the (many) new parameterized test cases in EvpSignatureTest are causing CodeBuild timeouts in the legacy `accp_github_pr` CodeBuild job that lives in corretto's account
    - this job runs the tests 3 times back-to-back, and will be removed soon in favor of the `accp-ci-pr-*` CodeBuild Batchs that provide the same coverage across a greater number of platforms and CPU architectures.
    - we'll be deleting this check soon, so reviewers can ignore it for the purposes of this review if it's still failing.

# Testing
- new parameterized unit test dimensions added for PSS changes
- for JUnit version bump, I reverted the threadsafety change and was able to reproduce randomly failing EvpSignatureTests under JUnit 5.8. Upon un-reverting the thread safety changes and tunning the tests under JUnit 5.8, the all pass.

# Description

This commit adds a [new flavor of RSA Signatures][1]: `RSASSA-PSS`.
PSS-padded RSA is described in [RFC 8017 section 8.1][2], with relevant
input parameter validation explained in [RFC 3447 section 9.1.1][3]. As
with OAEP RSA cipher padding, we mostly rely on AWS-LC to validate
message digest support, and explicity forbid anything outside of the SHA
family.

PSS allows users to specify the length (in bytes) of the randomly
generated salt. This value may not be less than 0 (JCE doesn't allow
negative salt lengths at all, and AWS-LC reserved them as special flag
values), so we enforce a minimum salt length of 0. The maximum is
specified in [RFC 3447 section 9.1.1][3], but relies on the key modulus
size to determine the maximum length. Since we don't always know the
key size when setParameter is called (key is specified on `init*`, and we
don't enforce ordering between calls to `setParameter` and `init*`), we
assume 2048 bit field size if no key is present. 2048 was selected
because it is the smallest field size that is still considered "secure",
and if a caller ends up using a smaller key, AWS-LC will detect this and
throw an error.

We switch the JCE provider for RSASSA-PSS parameterized tests in
EvpSignatureTest from BouncyCastle (BC) to [the SunRsaSign provider][6],
primarily due to differences in how they handle PSS parameters.
SunRsaSign [allows PSS message digest algorithm and MGF digest
algorithms to differ][4], while BC [does not][5]. So, in order to test
all possible combinations of PSS and MGF digest algorithms, we need to
migrate those tests to use SuncJCE as the JCE provider.

To increase compatibility accross providers, we allow callers to specify
`null` params in calls to `setParameter` (this is [consistent with
BC][7], but [not with SunRsaSign][8]. Similarly, we don't require
`setParameter` to be called on `RSASSA-PSS` instances before `init*` is
called, defaulting to digest algorithms of SHA1 unless otherwise
specified.

One point where we do diverge from SunRsaSign (but _not_ from BC) is
that we throw if callers attempt to update PSS parameters while
building or verifying a signature. We forbid this behavior to avoid
signature corruption due to digest algorithms being changed
while the signature/digest is still being updated.

Finally, @jakemas noted [a criticism][9] of PSS described by Peter
Gutman, the crux of which is:

> Now let's look at what an attacker can do with the above.  The
> immediately obvious, trivial attack is a hash-substitution attack.
> Unlike PKCS #1, PSS doesn't encode the hash algorithm used anywhere in
> the signature. [...] By changing hashAlgorithm to whatever you like, for
> example an algorithm of the same size that you know how to break, you
> can forge the signature.  Change the hash from SHA-2/256 to Streebog,
> Blake2, Keccak, CRC-256, XOR256, whatever you like, PSS won't detect the
> change, making it only as strong as the weakest hash algorithm of the
> same bit width that the victim will accept.

Unfortunately, because PSS is already standardized, there's not much we
can do to mitigate this in our impelementation. The main mitigation we
employ is to explicitly disallow non-SHA hash functions (e.g. MD5). All
SHA-family hash functions are currently believed to be secure for use in
PSS (indeed, the oldest, SHA1, is the default for Java) so no immediate
threat remains. If SHA1 is deemed insecure for use in PSS at some point
in the future, however, we will need to restrict its specification as a
PSS parameter.

[1]: https://docs.oracle.com/en/java/javase/11/docs/specs/security/standard-names.html#signature-algorithms
[2]: https://datatracker.ietf.org/doc/html/rfc8017#section-8.1
[3]: https://datatracker.ietf.org/doc/html/rfc3447#section-9.1.1
[4]: https://github.com/corretto/corretto-11/blob/f2c5b39cdb2afc9a9101620d14cf34c7448f826d/src/java.base/share/classes/sun/security/rsa/RSAPSSSignature.java#L290-L333
[5]: https://github.com/bcgit/bc-java/blob/bc3b92f1f0e78b82e2584c5fb4b226a13e7f8b3b/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/rsa/PSSSignatureSpi.java#L239-L242
[6]: https://docs.oracle.com/javase/7/docs/technotes/guides/security/SunProviders.html#SunRsaSignProvider
[7]: https://github.com/bcgit/bc-java/blob/bc3b92f1f0e78b82e2584c5fb4b226a13e7f8b3b/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/rsa/PSSSignatureSpi.java#L199-L207
[8]: https://github.com/corretto/corretto-11/blob/f2c5b39cdb2afc9a9101620d14cf34c7448f826d/src/java.base/share/classes/sun/security/rsa/RSAPSSSignature.java#L292-L294
[9]: https://marc.info/?l=cryptography&m=157345122116408&w=2

# ToDo
- [x] relax PSS salt lenght restrictions -- neither [the spec](https://datatracker.ietf.org/doc/html/rfc3447#appendix-A.2.3) nor [awslc](https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/rsa/padding.c#L649-L662) appear to enforce a cap on salt len beyond `mLen - hLen -2`
- [x] document above spec/impl. links, as well as [the `RSASSA-PSS` algorithm name required by JCE](https://docs.oracle.com/en/java/javase/11/docs/specs/security/standard-names.html)
- [x] clean up other ToDos
- [x] add `minimal_clean` gradle target that does a `rm build/cmake-coverage/*.jar build/lib/* build/awslc/bin/lib/libcrypto.so build/cmake/*.jar`
- [x] determine if/which KATs we want to add
- [x] parameterize PSS test case to test all SHA digest algorithms, salt lengths, etc.
- [x] add test case described [here](https://github.com/corretto/amazon-corretto-crypto-provider/pull/200/files/2f18616771b35ccde463e38363d06726c7556bea#r873938942)
- [x] Fix readme for both OAEP and PSS algorithms.

- [x] add test case to ensure we disallow MD5 and other "valid" but undesirable hash algorithms
- [x] parameterize more extant tests with new PSS params
- [x] address potential concern from @jakemas around [PSS criticism](https://marc.info/?l=cryptography&m=157345122116408&w=2) of not hashing/authenticating digest algorithm used in the signature. hwo might that affect which digest algorithms we'd support? should SHA-1 be rejected automatically (probably not as it seems to be the default).

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/EvpSignature.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpSignature.java
@@ -333,11 +333,12 @@ class EvpSignature extends EvpSignatureBase {
     }
 
     private InputBuffer<byte[], EvpContext, RuntimeException> getSigningBuffer() {
-        final String pssMgfMd = pssParams_ != null
-            ? Utils.jceDigestNameToAwsLcName(
-                ((MGF1ParameterSpec) pssParams_.getMGFParameters()).getDigestAlgorithm()
-            )
-            : null;
+        final String pssMgfMd;
+        if (pssParams_ != null) {
+            pssMgfMd = Utils.jceDigestNameToAwsLcName(((MGF1ParameterSpec) pssParams_.getMGFParameters()).getDigestAlgorithm());
+        } else {
+            pssMgfMd = null;
+        }
         final int pssSaltLen = pssParams_ != null ? pssParams_.getSaltLength() : 0;
         return new InputBuffer<byte[], EvpContext, RuntimeException>(1024)
             .withInitialUpdater((src, offset, length) ->
@@ -369,11 +370,14 @@ class EvpSignature extends EvpSignatureBase {
     }
 
     private InputBuffer<Boolean, EvpContext, SignatureException> getVerifyingBuffer() {
-        final String pssMgfMd = pssParams_ != null
-            ? Utils.jceDigestNameToAwsLcName(
+        final String pssMgfMd;
+        if (pssParams_ != null) {
+            pssMgfMd = Utils.jceDigestNameToAwsLcName(
                 ((MGF1ParameterSpec) pssParams_.getMGFParameters()).getDigestAlgorithm()
-            )
-            : null;
+            );
+        } else {
+            pssMgfMd = null;
+        }
         final int pssSaltLen = pssParams_ != null ? pssParams_.getSaltLength() : 0;
         return new InputBuffer<Boolean, EvpContext, SignatureException>(1024)
             .withInitialUpdater((src, offset, length) ->
--- a/src/com/amazon/corretto/crypto/provider/EvpSignature.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpSignature.java
@@ -4,8 +4,12 @@
 package com.amazon.corretto.crypto.provider;
 
 import java.nio.ByteBuffer;
+import java.security.InvalidAlgorithmParameterException;
 import java.security.SignatureException;
+import java.security.spec.AlgorithmParameterSpec;
 import java.security.spec.PKCS8EncodedKeySpec;
+import java.security.spec.MGF1ParameterSpec;
+import java.security.spec.PSSParameterSpec;
 import java.security.spec.X509EncodedKeySpec;
 
 class EvpSignature extends EvpSignatureBase {
@@ -301,10 +305,10 @@ class EvpSignature extends EvpSignatureBase {
      */
     private static native boolean verifyFinish(long ctx, byte[] signature, int sigOff, int sigLen) throws SignatureException;
 
-    private final String digestName_;
+    private String digestName_;
     private byte[] oneByteArray_ = null;
-    private final InputBuffer<byte[], EvpContext, RuntimeException> signingBuffer;
-    private final InputBuffer<Boolean, EvpContext, SignatureException> verifyingBuffer;
+    private InputBuffer<byte[], EvpContext, RuntimeException> signingBuffer;
+    private InputBuffer<Boolean, EvpContext, SignatureException> verifyingBuffer;
 
     /**
      * Creates a new instances of this class.
@@ -324,17 +328,28 @@ class EvpSignature extends EvpSignatureBase {
         Loader.checkNativeLibraryAvailability();
         digestName_ = digestName;
 
-        signingBuffer = new InputBuffer<byte[], EvpContext, RuntimeException>(1024)
+        signingBuffer = getSigningBuffer();
+        verifyingBuffer = getVerifyingBuffer();
+    }
+
+    private InputBuffer<byte[], EvpContext, RuntimeException> getSigningBuffer() {
+        final String pssMgfMd = pssParams_ != null
+            ? Utils.jceDigestNameToAwsLcName(
+                ((MGF1ParameterSpec) pssParams_.getMGFParameters()).getDigestAlgorithm()
+            )
+            : null;
+        final int pssSaltLen = pssParams_ != null ? pssParams_.getSaltLength() : 0;
+        return new InputBuffer<byte[], EvpContext, RuntimeException>(1024)
             .withInitialUpdater((src, offset, length) ->
                 new EvpContext(key_.use(ptr ->
                     signStart(ptr,
-                        digestName_, paddingType_, null, 0,
+                        digestName_, paddingType_, pssMgfMd, pssSaltLen,
                         src, offset, length)))
             )
             .withInitialUpdater((src) ->
                 new EvpContext(key_.use(ptr ->
                     signStartBuffer(ptr,
-                        digestName_, paddingType_, null, 0, src)))
+                        digestName_, paddingType_, pssMgfMd, pssSaltLen, src)))
             )
             .withUpdater((ctx, src, offset, length) ->
                 ctx.useVoid(ptr -> signUpdate(ptr, src, offset, length))
@@ -348,20 +363,29 @@ class EvpSignature extends EvpSignatureBase {
             .withSinglePass((src, offset, length) ->
                 key_.use(ptr ->
                     sign(ptr,
-                            digestName_, paddingType, null, 0,
+                            digestName_, paddingType_, pssMgfMd, pssSaltLen,
                             src, offset, length))
             );
-        verifyingBuffer = new InputBuffer<Boolean, EvpContext, SignatureException>(1024)
+    }
+
+    private InputBuffer<Boolean, EvpContext, SignatureException> getVerifyingBuffer() {
+        final String pssMgfMd = pssParams_ != null
+            ? Utils.jceDigestNameToAwsLcName(
+                ((MGF1ParameterSpec) pssParams_.getMGFParameters()).getDigestAlgorithm()
+            )
+            : null;
+        final int pssSaltLen = pssParams_ != null ? pssParams_.getSaltLength() : 0;
+        return new InputBuffer<Boolean, EvpContext, SignatureException>(1024)
             .withInitialUpdater((src, offset, length) ->
                 new EvpContext(key_.use(ptr ->
                     verifyStart(ptr, digestName_,
-                        paddingType_, null, 0,
+                        paddingType_, pssMgfMd, pssSaltLen,
                         src, offset, length)))
             )
             .withInitialUpdater((src) ->
                 new EvpContext(key_.use(ptr ->
                     verifyStartBuffer(ptr,
-                        digestName_, paddingType_, null, 0, src)))
+                        digestName_, paddingType_, pssMgfMd, pssSaltLen, src)))
             )
             .withUpdater((ctx, src, offset, length) ->
                 ctx.useVoid(ptr -> verifyUpdate(ptr, src, offset, length))
@@ -376,6 +400,23 @@ class EvpSignature extends EvpSignatureBase {
     protected synchronized void engineReset() {
         signingBuffer.reset();
         verifyingBuffer.reset();
+    }
+
+    @Override
+    protected synchronized void engineSetParameter(final AlgorithmParameterSpec params)
+            throws InvalidAlgorithmParameterException {
+        super.engineSetParameter(params);
+        if (params instanceof PSSParameterSpec) {
+            final String jceDigestName = ((PSSParameterSpec) params).getDigestAlgorithm();
+            digestName_ = Utils.jceDigestNameToAwsLcName(jceDigestName);
+            // referesh signing and verifying buffer closures now that we've updated PSS params
+            signingBuffer = getSigningBuffer();
+            verifyingBuffer = getVerifyingBuffer();
+        }
+    }
+
+    protected boolean isBufferEmpty() {
+        return signingBuffer.size() == 0 && verifyingBuffer.size() == 0;
     }
 
     @Override
@@ -441,6 +482,12 @@ class EvpSignature extends EvpSignatureBase {
             finalOff = off;
             finalLen = len;
         }
+        final String pssMgfMd = pssParams_ != null
+            ? Utils.jceDigestNameToAwsLcName(
+                ((MGF1ParameterSpec) pssParams_.getMGFParameters()).getDigestAlgorithm()
+            )
+            : null;
+        final int pssSaltLen = pssParams_ != null ? pssParams_.getSaltLength() : 0;
         try {
             return verifyingBuffer
                 .withDoFinal((ctx) ->
@@ -448,7 +495,7 @@ class EvpSignature extends EvpSignatureBase {
                 )
                 .withSinglePass((src, offset, length) ->
                     key_.use(ptr -> verify(ptr,
-                        digestName_, paddingType_, null, 0,
+                        digestName_, paddingType_, pssMgfMd, pssSaltLen,
                         src, offset, length, finalSigBytes, finalOff, finalLen))
                 ).doFinal();
         } finally {
@@ -485,6 +532,12 @@ class EvpSignature extends EvpSignatureBase {
     static final class SHA512withRSA extends EvpSignature {
         SHA512withRSA(AmazonCorrettoCryptoProvider provider) {
             super(provider, EvpKeyType.RSA, RSA_PKCS1_PADDING, "sha512");
+        }
+    }
+
+    static final class RSASSA_PSS extends EvpSignature {
+        RSASSA_PSS(AmazonCorrettoCryptoProvider provider) {
+            super(provider, EvpKeyType.RSA, RSA_PKCS1_PSS_PADDING, null);
         }
     }
 
--- a/src/com/amazon/corretto/crypto/provider/EvpSignature.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpSignature.java
@@ -46,7 +46,7 @@ class EvpSignature extends EvpSignatureBase {
      *      href="https://www.openssl.org/docs/man1.1.0/crypto/EVP_get_digestbyname.html">EVP_get_digestbyname</a>
      */
     private static native byte[] sign(long privateKey, String digestName, int paddingType, String mgfMd,
-            int saltLen, byte[] message, int offset, int length);
+            int saltLen, byte[] message, int offset, int length) throws SignatureException;
 
 
     /**
@@ -286,7 +286,7 @@ class EvpSignature extends EvpSignatureBase {
      *            {@link #signStart(byte[], int, String, int, String, int, byte[], int, int)} or
      *            {@link #signStartBuffer(byte[], int, String, int, String, int, ByteBuffer)}.
      */
-    private static native byte[] signFinish(long ctx);
+    private static native byte[] signFinish(long ctx) throws SignatureException;
 
     /**
      * Verifies the signature and <em>destroys the context</em>.
@@ -307,7 +307,7 @@ class EvpSignature extends EvpSignatureBase {
 
     private String digestName_;
     private byte[] oneByteArray_ = null;
-    private InputBuffer<byte[], EvpContext, RuntimeException> signingBuffer;
+    private InputBuffer<byte[], EvpContext, SignatureException> signingBuffer;
     private InputBuffer<Boolean, EvpContext, SignatureException> verifyingBuffer;
 
     /**
@@ -332,7 +332,7 @@ class EvpSignature extends EvpSignatureBase {
         verifyingBuffer = getVerifyingBuffer();
     }
 
-    private InputBuffer<byte[], EvpContext, RuntimeException> getSigningBuffer() {
+    private InputBuffer<byte[], EvpContext, SignatureException> getSigningBuffer() {
         final String pssMgfMd;
         if (pssParams_ != null) {
             pssMgfMd = Utils.jceDigestNameToAwsLcName(((MGF1ParameterSpec) pssParams_.getMGFParameters()).getDigestAlgorithm());
@@ -340,7 +340,7 @@ class EvpSignature extends EvpSignatureBase {
             pssMgfMd = null;
         }
         final int pssSaltLen = pssParams_ != null ? pssParams_.getSaltLength() : 0;
-        return new InputBuffer<byte[], EvpContext, RuntimeException>(1024)
+        return new InputBuffer<byte[], EvpContext, SignatureException>(1024)
             .withInitialUpdater((src, offset, length) ->
                 new EvpContext(key_.use(ptr ->
                     signStart(ptr,
--- a/src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
@@ -169,7 +169,8 @@ abstract class EvpSignatureBase extends SignatureSpi {
             final int saltLen = pssParams.getSaltLength();
             final int mdLen = Utils.getMdLen(Utils.getMdPtr(pssParams.getDigestAlgorithm()));
             // If key is not yet set, assume it has a 2048-bit modulus. Even if a smaller key ends up being
-            // used, AWS-LC will detect this and throw an error.
+            // used, AWS-LC will detect this and throw an error here:
+            // https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/rsa/padding.c#L661
             final int emLen = key_ != null
                 ? (((RSAKey) key_).getModulus().bitLength() + 7) / 8
                 : 2048 / 8;
--- a/src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
@@ -5,6 +5,7 @@ package com.amazon.corretto.crypto.provider;
 
 import java.math.BigInteger;
 import java.security.AlgorithmParameters;
+import java.security.GeneralSecurityException;
 import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
 import java.security.InvalidParameterException;
@@ -14,7 +15,10 @@ import java.security.PublicKey;
 import java.security.SignatureException;
 import java.security.SignatureSpi;
 import java.security.interfaces.ECKey;
+import java.security.interfaces.RSAKey;
 import java.security.spec.AlgorithmParameterSpec;
+import java.security.spec.MGF1ParameterSpec;
+import java.security.spec.PSSParameterSpec;
 import java.util.Arrays;
 import java.util.Base64;
 
@@ -22,14 +26,16 @@ abstract class EvpSignatureBase extends SignatureSpi {
     // Package visible so main Provider can use it
     static final String P1363_FORMAT_SUFFIX = "inP1363Format";
     protected static final int RSA_PKCS1_PADDING = 1;
+    protected static final int RSA_PKCS1_PSS_PADDING = 6;
     protected final AmazonCorrettoCryptoProvider provider_;
     protected final EvpKeyType keyType_;
-    protected final int paddingType_;
+    protected int paddingType_;
     protected Key untranslatedKey_ = null;
     protected EvpKey key_ = null;
     protected boolean signMode;
     protected int keyUsageCount_ = 0;
     protected String algorithmName_ = null;
+    protected PSSParameterSpec pssParams_ = null;
 
     EvpSignatureBase(
             final AmazonCorrettoCryptoProvider provider,
@@ -39,6 +45,9 @@ abstract class EvpSignatureBase extends SignatureSpi {
         provider_ = provider;
         keyType_ = keyType;
         paddingType_ = paddingType;
+        if (paddingType_ == RSA_PKCS1_PSS_PADDING) {
+            pssParams_ = PSSParameterSpec.DEFAULT;
+        }
     }
 
     protected abstract void engineReset();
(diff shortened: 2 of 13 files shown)


SalusaSecondus:
Note: Fix readme for both OAEP and PSS algorithms.

WillChilds-Klein:
closing/reopening to prod the CI. die harder failure looks transient.

WillChilds-Klein:
also, a note to reviewers -- the [libaccp plumbing](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/csrc/sign.cpp#L40) for PSS has been in place [for a while now](https://github.com/corretto/amazon-corretto-crypto-provider/commit/1c7201392deada85cad5853684c8154b3d24e94c), but appears to have been unused. i reviewed the relevant portion of `sign.cpp` and didn't identify any issues upon inspection, but i'd also request that reviewers also take a look at that code for potential issues.

SalusaSecondus:
**NOTE: We must ensure `EvpSignatureTest` is thread-safe as part of this change.**

src/com/amazon/corretto/crypto/provider/EvpSignature.java
```diff
@@ -324,17 +328,28 @@ private EvpSignature(AmazonCorrettoCryptoProvider provider, final EvpKeyType key
...
+        signingBuffer = getSigningBuffer();
+        verifyingBuffer = getVerifyingBuffer();
+    }
+
+    private InputBuffer<byte[], EvpContext, RuntimeException> getSigningBuffer() {
+        final String pssMgfMd = pssParams_ != null
+            ? Utils.jceDigestNameToAwsLcName(
+                ((MGF1ParameterSpec) pssParams_.getMGFParameters()).getDigestAlgorithm()
+            )
+            : null;
```
SalusaSecondus, line 340: I think the following might be more readable.
```suggestion
        final String pssMgfMd;
        if (pssParams != null) {
            pssMgfMd = Utils.jceDigestNameToAwsLcName(((MGF1ParameterSpec) pssParams_.getMGFParameters()).getDigestAlgorithm());
        } else {
            pssMgfMd = null;
        }
```

src/com/amazon/corretto/crypto/provider/EvpSignature.java
```diff
@@ -348,20 +363,29 @@ private EvpSignature(AmazonCorrettoCryptoProvider provider, final EvpKeyType key
...
                     sign(ptr,
-                            digestName_, paddingType, null, 0,
+                            digestName_, paddingType_, pssMgfMd, pssSaltLen,
                             src, offset, length))
             );
-        verifyingBuffer = new InputBuffer<Boolean, EvpContext, SignatureException>(1024)
+    }
+
+    private InputBuffer<Boolean, EvpContext, SignatureException> getVerifyingBuffer() {
+        final String pssMgfMd = pssParams_ != null
```
SalusaSecondus, line 372: Ditto here on thinking the ternary operator makes things hard to read.

src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
```diff
@@ -113,13 +122,81 @@ protected void engineSetParameter(final String param, final Object value) throws
...
-        if (params != null) {
-            throw new InvalidAlgorithmParameterException("No parameters supported by this algorithm");
+        if (params instanceof PSSParameterSpec) {
+            if (!isBufferEmpty()) {
+                throw new IllegalStateException("Cannot update PSS parameters with buffered data, reset Signature.");
+            }
+            final PSSParameterSpec pssParams = (PSSParameterSpec) params;
+            if (keyType_ != EvpKeyType.RSA || paddingType_ != RSA_PKCS1_PSS_PADDING) {
+                throw new InvalidAlgorithmParameterException("PSS params only supported for RSASSA-PSS signatures");
+            }
```
SalusaSecondus, line 132: Move this check before checking for an empty buffer. This is the more important failure.

src/com/amazon/corretto/crypto/provider/Utils.java
```diff
@@ -59,6 +64,29 @@ private Utils() {
...
+        return jceName.replace("-", "").replace("/", "-").toUpperCase();
+    }
+
+    static long getMdPtr(String digestName) {
+        final String name = jceDigestNameToAwsLcName(digestName);
+        // NOTE: it's a little awkward to throw an unchecked exception here, but the lambda
+        // invoking the native Utils.getEvpMdFromName can also throw an unchecked exception,
+        // so callers already need to handle this.
+        if (!name.startsWith("SHA")) {
+            throw new RuntimeException("Unsupported digest algorithm");
```
SalusaSecondus, line 81: This should be an `IllegalArgumentException`

src/com/amazon/corretto/crypto/provider/Utils.java
```diff
@@ -59,6 +64,29 @@ private Utils() {
...
+        }
+        return digestPtrByName.computeIfAbsent(name, n -> Utils.getEvpMdFromName(n));
```
SalusaSecondus, line 83:
```suggestion
        return digestPtrByName.computeIfAbsent(name, Utils::getEvpMdFromName);
```

src/com/amazon/corretto/crypto/provider/Utils.java
```diff
@@ -59,6 +64,29 @@ private Utils() {
...
+    }
+
+    static int getMdLen(long mdPtr) {
+        return digestLengthByPtr.computeIfAbsent(mdPtr, p -> Utils.getDigestLength(p));
```
SalusaSecondus, line 87:
```suggestion
        return digestLengthByPtr.computeIfAbsent(mdPtr, Utils::getDigestLength);
```

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -364,6 +368,168 @@ public void testRsaWithoutCrtParams() throws Exception {
...
+        signer.initSign(pair.getPrivate());
+        verifier.initVerify(pair.getPublic());
+        signer.setParameter(spec1);
+        verifier.setParameter(spec1);
+
+        // Assert that set values persisted
+        assertEquals(
+            spec1.getDigestAlgorithm(),
+            signer.getParameters().getParameterSpec(PSSParameterSpec.class).getDigestAlgorithm()
+        );
```
SalusaSecondus, line 460: You might find it helpful to create a new `assertPssParamsEquals()` method.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -364,6 +368,168 @@ public void testRsaWithoutCrtParams() throws Exception {
...
+
+        // Update message digest, assert that we can't upate params, and that signature verifies.
+        signer.update(MESSAGE);
+        verifier.update(MESSAGE);
+        assertThrows(IllegalStateException.class, () -> signer.setParameter(spec2));
+        assertThrows(IllegalStateException.class, () -> verifier.setParameter(spec2));
+        assertTrue(verifier.verify(signer.sign()));
+
+        // Finally, assert that we can set params after reinit (verify doesn't need explicit re-init)
```
SalusaSecondus, line 481: You shouldn't need to explicitly re-init signer either (as calling `.sign()` implicitly re-inits it).

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -364,6 +368,168 @@ public void testRsaWithoutCrtParams() throws Exception {
...
+        // Now, initialize the signature with a smaller key
+        final KeyPairGenerator kg = KeyPairGenerator.getInstance("RSA", AmazonCorrettoCryptoProvider.INSTANCE);
+        kg.initialize(smallKeySize);
+        KeyPair pair = kg.generateKeyPair();
+        signer.initSign(pair.getPrivate());
+
+        // Assert that doFinal (i.e. the native method called in signer.sign()) detects overly large salt len and throws
+        byte[] shortMessage = new byte[] { (byte) 0xDE, (byte) 0xAD, (byte) 0xBE, (byte) 0xEF };
+        signer.update(shortMessage);
+        assertThrows(RuntimeCryptoException.class, () -> signer.sign());
```
SalusaSecondus, line 523: Is this really the correct exception? We can live with it but I'd prefer to see a more specific exception when we call `initSign`

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -460,25 +625,52 @@ public void simpleCorrectnessAllAlgorithms() throws Throwable {
...
+        //       SuncJCE in EvpSignatureTest. enforce min key size of 2048 bits, consistent with ffSize above.
+        for (int shaVersion : new int[] {1, 224, 256, 384, 512}) {
+            final String algorithm = "RSASSA-PSS";
+            final String mdName = String.format("SHA-%d", shaVersion);
+            final int mdLen = MessageDigest.getInstance(mdName).getDigestLength();
+            final int keySize = mdLen * 8 < 2048 ? 2048 : mdLen * 8;
+            final KeyPairGenerator kg = KeyPairGenerator.getInstance("RSA");
+            kg.initialize(keySize);
+            final KeyPair pair = kg.generateKeyPair();
+            final Signature nativeSig = Signature.getInstance(algorithm, AmazonCorrettoCryptoProvider.INSTANCE);
```
SalusaSecondus, line 642: Have this use `NATIVE_PROVIDER`.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -460,25 +625,52 @@ public void simpleCorrectnessAllAlgorithms() throws Throwable {
             final Signature nativeSig = Signature.getInstance(algorithm, AmazonCorrettoCryptoProvider.INSTANCE);
```
SalusaSecondus, line 625: Have this use `NATIVE_PROVIDER`.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -460,25 +625,52 @@ public void simpleCorrectnessAllAlgorithms() throws Throwable {
...
-                throw new AssertionError(algorithm, ex);
-            }
+            simpleCorrectnessSignVerify(algorithm, pair, bcSig, nativeSig);
+        }
+
+        // NOTE: for RSASSA-PSS, test supported digest lengths, but  keep PSS and MGF1 digest lengths equal because
+        //       BouncyCastle doesn't support differing lengths. we test differing lengths exhaustively against
+        //       SuncJCE in EvpSignatureTest. enforce min key size of 2048 bits, consistent with ffSize above.
+        for (int shaVersion : new int[] {1, 224, 256, 384, 512}) {
+            final String algorithm = "RSASSA-PSS";
```
SalusaSecondus, line 635: Add a check which skips over these tests for versions < 2.0.0

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureTest.java
```diff
@@ -62,19 +71,29 @@ private static class TestParams {
...
             this.keyPair = keyPair;
+            this.paramSpec = paramSpec;
 
             signer = getNativeSigner();
+            if (paramSpec != null) {
+                signer.setParameter(paramSpec);
+            }
             signer.initSign(keyPair.getPrivate());
             verifier = getNativeSigner();
+            verifier.setParameter(paramSpec);
```
SalusaSecondus, line 90: Why does `signer.setParameter()` need to be in an if-block but not for the `verifier`?

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureTest.java
```diff
@@ -84,18 +103,33 @@ public TestParams(String base, String algorithm, int length, boolean readOnly, b
...
+            final String pssParams = pssParamSpec == null
+                ? null
+                : String.format(
+                    "{PSS md: %s, MGF1 md: %s, saltLen: %d}",
+                    pssParamSpec.getDigestAlgorithm(),
+                    pssParamSpec.getMGFParameters() == null
+                        ? null
+                        : ((MGF1ParameterSpec) pssParamSpec.getMGFParameters()).getDigestAlgorithm(),
+                    pssParamSpec.getSaltLength()
+                );
```
SalusaSecondus, line 126: Please be cautious with the ternary operator. It's easy for it to make things really hard to read.

You don't need the second null-check at all because we can assume that the MGFParameters aren't null as that would be invalid (and we're creating them ourselves)

```suggestion
            String pssParams = null;
            if (pssParamSpec != null) {
                pssParams = String.format(
                    "{PSS md: %s, MGF1 md: %s, saltLen: %d}",
                    pssParamSpec.getDigestAlgorithm(),
                    ((MGF1ParameterSpec) pssParamSpec.getMGFParameters()).getDigestAlgorithm(),
                    pssParamSpec.getSaltLength()
                );
            }
```

SalusaSecondus (commented):
Looking good. Just minor things

src/com/amazon/corretto/crypto/provider/EvpSignature.java
WillChilds-Klein: sure.

src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
WillChilds-Klein: done.

src/com/amazon/corretto/crypto/provider/Utils.java
WillChilds-Klein: ah, yep. agreed. done.

src/com/amazon/corretto/crypto/provider/Utils.java
WillChilds-Klein: yep, good call.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
WillChilds-Klein: yeah, might as well.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
WillChilds-Klein: done.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
WillChilds-Klein: i went ahead and changed all references to AmazonCorrettoCryptoProvider.INSTANCE to NATIVE_INSTANCE in this test file.

src/com/amazon/corretto/crypto/provider/EvpSignature.java
```diff
@@ -324,17 +328,29 @@ private EvpSignature(AmazonCorrettoCryptoProvider provider, final EvpKeyType key
...
         digestName_ = digestName;
 
-        signingBuffer = new InputBuffer<byte[], EvpContext, RuntimeException>(1024)
+        signingBuffer = getSigningBuffer();
+        verifyingBuffer = getVerifyingBuffer();
+    }
+
+    private InputBuffer<byte[], EvpContext, RuntimeException> getSigningBuffer() {
+        final String pssMgfMd;
+        if (pssParams != null) {
```
SalusaSecondus, line 337: FYI, this is the line which is failing the CI right now.
```suggestion
        if (pssParams_ != null) {
```

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureTest.java
WillChilds-Klein: readability concerns understood, i'll update to use conventional conditional blocks. and sure, i can remove the second null check.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
WillChilds-Klein: reinit removed.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
WillChilds-Klein: added here and in EvpSignatureTest parameter generation.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
WillChilds-Klein: hm, this test case is looking for the native provider to throw on signing finalization, not initialization. we could `catch` [this call](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/src/com/amazon/corretto/crypto/provider/EvpSignature.java#L385) in `engineSign` and wrap/rethrow with another exception if you think that's more appropriate than RuntimeCryptoException.

which exception were you thinking? `InvalidAlgorithmParameterException`?

WillChilds-Klein:
the failing CI check is due to test flakiness, and appears to be unrelated to this change. we'll remove the legacy `accp_github_*` checks soon, anyway

Screen Shot 2022-06-13 at 12 18 17 PM

WillChilds-Klein:
one of the CI tests failed with:


```
JUnit Jupiter:EvpSignatureTest:verifySubByteBufferWrap(TestParams)
--
MethodSource [className = 'com.amazon.corretto.crypto.provider.test.EvpSignatureTest', methodName = 'verifySubByteBufferWrap', methodParameterTypes = 'com.amazon.corretto.crypto.provider.test.EvpSignatureTest$TestParams']
=> java.util.ConcurrentModificationException
java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1135)
com.amazon.corretto.crypto.provider.Utils.getMdLen(Utils.java:87)
com.amazon.corretto.crypto.provider.EvpSignatureBase.engineSetParameter(EvpSignatureBase.java:170)
com.amazon.corretto.crypto.provider.EvpSignature.engineSetParameter(EvpSignature.java:412)
java.base/java.security.Signature$Delegate.engineSetParameter(Signature.java:1434)
```

so, it looks like while we managed to make EvpSignatureTest threadsafe, the digest ptr caches used in our application code wasn't threadsafe. solution here should be simple -- i'll swap out the caches' HashMap for a [ConcurrentHashMap](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html).

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
SalusaSecondus: Can't we detect this failure the moment that either `setParameters()` or `init()` is called? The moment that the salt is incompatible with the key we should detect this and throw and appropriate exception (relative to the call).

SalusaSecondus:
Now that we have simple ways to manage the `EVP_MD*` values as `long`s in Java, we should change [initializeContext()](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/csrc/sign.cpp#L61/) (and thus [configurePadding()](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/csrc/sign.cpp#L31), [signStart()](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/src/com/amazon/corretto/crypto/provider/EvpSignature.java#L121), etc.) to no longer take a `String` containing the name of the message digest but rather a `long` containing the value of `EVP_MD*`.

You'd change the type of [digestName_](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/src/com/amazon/corretto/crypto/provider/EvpSignature.java#L304) to a `long` and also introduce a new field for the MGF MD `long` value.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
WillChilds-Klein: we can only detect this in `setParameters()` if `init()` has already been called. for greater compatibility, this implementation does not enforce an order between those two calls being made (neither does [Sun](https://github.com/corretto/corretto-11/blob/f2c5b39cdb2afc9a9101620d14cf34c7448f826d/src/java.base/share/classes/sun/security/rsa/RSAPSSSignature.java#L181-L184) nor [BouncyCastle](https://github.com/bcgit/bc-java/blob/bc3b92f1f0e78b82e2584c5fb4b226a13e7f8b3b/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/rsa/PSSSignatureSpi.java#L259-L270)). from the commit message:

> To increase compatibility accross providers, we allow callers to specify
> null params in calls to setParameter (this is consistent with
> BC, but not with SunRsaSign. **Similarly, we don't require
> `setParameter` to be called on RSASSA-PSS instances before `init*` is
> called**, defaulting to digest algorithms of SHA1 unless otherwise
> specified.

i could put checks in each of `init()` and `setParameters()` and throw when either detects an incompatible size. if this sounds like a good approach, is `InvalidAlgorithmParameterException` the correct exception to throw?

SalusaSecondus: You know what, I think the current solution is better. Otherwise you need to update the key and parameters in just the correct order and it only matters once they update or do the final call.

Technically we could throw as soon as either update or final is called _but_ I don't think it's worth it. Keep the exception on the final call for simplicity.

However, we should probably throw a [SignatureException](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/security/SignatureException.html) in that case.

SalusaSecondus: 1. Is this only possible on the signing flow and not the verification flow? (I only see the new check on the signing flow.)
2. As `SignatureException` is a checked exception, please declare it on the relevant native methods in Java as well.

WillChilds-Klein: > 1. Is this only possible on the signing flow and not the verification flow? (I only see the new check on the signing flow.)

I believe this is only possible in the signing flow. The condition this test checks is for when `salt length + digest length + message length` exceeds the maximum total message size (referred to as "EM" in the RFC), which is the size of the modulus. Since such a signature can't be generated, there's nothing to verify.

In local testing, no exception was thrown when i tried to verify an empty signature byte array with the oversized salt length + small key parameter set. Signature.verify() simply returned `false`.

> 2. As `SignatureException` is a checked exception, please declare it on the relevant native methods in Java as well.

Ah, thanks for the reminder. done.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -364,21 +369,173 @@ public void testRsaWithoutCrtParams() throws Exception {
...
+        kg.initialize(minimallySecureKeyLen);
+        pair = kg.generateKeyPair();
+        signer.initSign(pair.getPrivate());
+        signer.update(shortMessage);
+        final byte[] signature = signer.sign();
+
+        // For completeness, also verify the data and signature. We don't test verification with the smaller key
+        // size because generating a siganture with oversized salt length + undersized key is not possible. If no
+        // signature can be generated, there's nothing nothing to verify.
+        final Signature verifier = Signature.getInstance("RSASSA-PSS", NATIVE_PROVIDER);
```
SalusaSecondus, line 516: What happens when you *try* to validate the signature though? Yeah, it cannot actually exist. So generate a valid signature (with reasonable salt) and then configure your validation logic to use it invalid salt. Do you get a `false` (because it isn't valid), a `SignatureException` (correctly detecting there is something wrong with the configuration/setup) or a `CryptoRuntimeException` because it fails in an unexpected way?

WillChilds-Klein: Slightly different case than what you describe above, but same result as [mentioned in our prior thread](https://github.com/corretto/amazon-corretto-crypto-provider/pull/202#discussion_r898268843). It simply returns false. To be totally explicit, with this diff applied to the PR's [current HEAD commit](https://github.com/corretto/amazon-corretto-crypto-provider/commit/ed7ec940e9d8a56ba8075a966930e1757c2109ea):

```
diff --git a/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java b/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
index 3c8c612..82d4e13 100644
--- a/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
@@ -518,6 +518,23 @@ public final class EvpSignatureSpecificTest {
         verifier.initVerify(pair.getPublic());
         verifier.update(shortMessage);
         assertTrue(verifier.verify(signature));
+
+        // Generate a valid signature with an undersized key but minimal salt length. Then, bump the salt length
+        // for the verifier instance and try to verify the signature. The verification attempt will return false
+        // instead of throwing. Apparently AWS-LC doesn't perform relevant validation upon verification.
+        PSSParameterSpec spec2 = new PSSParameterSpec("SHA-1", "MGF1", MGF1ParameterSpec.SHA1, /*saltLen*/20, 1);
+        kg.initialize(smallKeySize);
+        pair = kg.generateKeyPair();
+        signer.setParameter(spec2);
+        signer.initSign(pair.getPrivate());
+        signer.update(shortMessage);
+        final byte[] signature2 = signer.sign();
+
+        spec2 = new PSSParameterSpec("SHA-1", "MGF1", MGF1ParameterSpec.SHA1, largeSaltLen, 1);
+        verifier.setParameter(spec2);
+        verifier.initVerify(pair.getPublic());
+        verifier.update(shortMessage);
+        assertFalse(verifier.verify(signature));
     }
 
 

```

All tests pass:

```
root@79bf3ce255b6:/accp# ./gradlew cmake_clean && ./gradlew single_test -DSINGLE_TEST=com.amazon.corretto.crypto.provider.test.EvpSignatureSpecificTest
...
[         0 tests failed          ]

[100%] Built target check-junit-single

BUILD SUCCESSFUL in 2m 25s
```

SalusaSecondus (approved):
LGTM

WillChilds-Klein:
correctness checks pass, merging.

PR #203, opened 2022-05-16
Reduce unneeded copies in aes-gcm decryption

SalusaSecondus:
*Description of changes:*

This reduces (by one) the number of data copies when AES-GCM decryption is performed in a single `doFinal` call. This should help improve our speed in this case.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
@@ -197,6 +197,7 @@ final class AesGcmSpi extends CipherSpi {
     private NativeResource context = null;
     private SecretKey jceKey;
     private byte[] iv, key;
+    /** GCM tag length in bytes */
     private int tagLength;
     private int opMode = -1;
     private boolean hasConsumedData = false;
@@ -204,7 +205,8 @@ final class AesGcmSpi extends CipherSpi {
     private int keyUsageCount = 0;
     private boolean contextInitialized = false;
 
-    private AccessibleByteArrayOutputStream decryptInputBuf, decryptAADBuf;
+    private final AccessibleByteArrayOutputStream decryptInputBuf = new AccessibleByteArrayOutputStream();
+    private final AccessibleByteArrayOutputStream decryptAADBuf = new AccessibleByteArrayOutputStream();
 
     AesGcmSpi(AmazonCorrettoCryptoProvider provider) {
         this.provider = provider;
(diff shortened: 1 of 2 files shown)


src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -555,16 +557,34 @@ protected synchronized int engineDoFinal(byte[] bytes, int offset, int length, b
...
-                final int tempBufferLength = decryptInputBuf.size();
-
-                if (tempBufferLength < tagLength) {
+                // We cannot write directly to the output array because that could release
+                // possibly unauthenticated plaintext.
+                // If we have already processed some amount of ciphertext, then we can use our existing input buffer
+                // for both input and output and then copy it over to the real output upon success.
+                // If we have not already processed some ciphertext, then we can use the input buffer as is
+                // and allocate temporary output buffer.
+                if (decryptInputBuf.size() + length < tagLength) {
```
WillChilds-Klein, line 566: isn't this inconsistent with [how we validate tagLen in our C++ shim](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/csrc/aes_gcm.cpp#L395)?

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -555,16 +557,34 @@ protected synchronized int engineDoFinal(byte[] bytes, int offset, int length, b
...
+                final byte[] workingOutputArray;
+                final int workingInputOffset;
+                final int workingInputLength;
+                if (decryptInputBuf.size() > 0) {
+                    // We've already processed data so need to merge this in
+                    engineUpdate(bytes, offset, length); // Decrypt mode never generates output for updates
+                    workingInputArray = decryptInputBuf.getDataBuffer();
+                    workingInputLength = decryptInputBuf.size();
+                    workingInputOffset = 0;
+                    workingOutputArray = workingInputArray; // We can decrypt in place
```
WillChilds-Klein, line 580: just to check my understanding -- this is where you're saving a copy, right? the idea is that instead of making [each call to EVP_CipherUpdate](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/csrc/aes_gcm.cpp#L81) do a bunch of extra loads/stores for some array elsewhere in memory, it's already got the (consolidated in/out) buffer loaded and ready to go?

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
SalusaSecondus: No, this is the same logic. However, this is prior to us merging data in `decryptInputBuf` and `bytes` (whose length is represented by `length`). So, after we merge the two, `inlen` (in C++) is equal to `decryptInputBuf.size() + length` (in Java).

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
WillChilds-Klein: synced up offline -- my confusion was that I missed the [underlying copy](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java#L425) in the call to `engineUpdate` on line 576. in the one-shot case (i.e. the `else` branch below), we simply reference the input array and pass that down to AWS-LC instead of copying `bytes`'s contents into our instance buffer.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
WillChilds-Klein: ah got it. i think the root cause of my misunderstanding here was also missing the fact that L576's call to `engineUpdate` [modifies `decryptInputBuf`](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java#L425) and increases its size.

WillChilds-Klein (approved):

PR #204, opened 2022-05-17
Optimize GCM more and fix some linting

SalusaSecondus:
*Description of changes:*

This should even better optimize AES-GCM logic by separating out encrypt and decrypt flows. Decryption always requires creating a temporary buffer for the plaintext (so, we can just return it sometimes, but cannot avoid creating it.) Encryption doesn't always need a temporary output buffer so we can often write directly to the buffer passed in by the caller.

I've also done minor clean up to reduce noise from the linter.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
@@ -46,9 +46,6 @@ final class AesGcmSpi extends CipherSpi {
      * this is because we cannot safely return only plaintext until we have validated
      * the tag at the end of the ciphertext.
      * Additionally, this matches JCE behavior.
-     *
-     * For non-one-shot encryption, to avoid any security issues related to thread safety concerns, we mark all methods
-     * which might eventually interact with the native context buffer as being synchronized.
      */
 
     private static final int NATIVE_MODE_ENCRYPT = 1;
@@ -256,7 +253,7 @@ final class AesGcmSpi extends CipherSpi {
     /* Returns the maximum amount of data that could be returned from an update (not doFinal) operation.
      * Not exposed via the Cipher API, but used internally to allocate buffers to return to the caller.
      */
-    private synchronized int getUpdateOutputSize(final int inputLen) {
+    private int getUpdateOutputSize(final int inputLen) {
         switch (opMode) {
             case NATIVE_MODE_ENCRYPT:
                 return inputLen;
@@ -285,7 +282,7 @@ final class AesGcmSpi extends CipherSpi {
     }
 
     @Override
-    protected synchronized void engineInit(final int opMode, final Key key, final SecureRandom secureRandom)
+    protected void engineInit(final int opMode, final Key key, final SecureRandom secureRandom)
             throws InvalidKeyException {
         if (opMode != Cipher.ENCRYPT_MODE && opMode != Cipher.WRAP_MODE) {
             throw new InvalidKeyException("IV required for decrypt");
@@ -302,7 +299,7 @@ final class AesGcmSpi extends CipherSpi {
     }
 
     @Override
-    protected synchronized void engineInit(
+    protected void engineInit(
         final int jceOpMode,
         final Key key,
         final AlgorithmParameterSpec algorithmParameterSpec,
@@ -402,7 +399,7 @@ final class AesGcmSpi extends CipherSpi {
     }
 
     @Override
-    protected synchronized void engineInit(
+    protected void engineInit(
         int opMode, Key key, AlgorithmParameters algorithmParameters, SecureRandom secureRandom
     ) throws InvalidKeyException, InvalidAlgorithmParameterException {
         try {
@@ -413,7 +410,7 @@ final class AesGcmSpi extends CipherSpi {
     }
 
     @Override
-    protected synchronized byte[] engineUpdate(byte[] bytes, int offset, int length) {
+    protected byte[] engineUpdate(byte[] bytes, int offset, int length) {
         byte[] buf = new byte[getUpdateOutputSize(length)];
 
         int actualLength;
@@ -431,7 +428,7 @@ final class AesGcmSpi extends CipherSpi {
     }
 
     @Override
-    protected synchronized int engineUpdate(byte[] bytes, int offset, int length, byte[] output, int outputOffset) throws ShortBufferException {
+    protected int engineUpdate(byte[] bytes, int offset, int length, byte[] output, int outputOffset) throws ShortBufferException {
         checkArrayLimits(bytes, offset, length);
 
         hasConsumedData = true;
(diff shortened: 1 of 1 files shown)


src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -924,6 +981,10 @@ private void checkNeedReset() {
 
     // @GuardedBy("this") // Restore once replacement for JSR-305 available
     private void stateReset() {
+        // While this shouldn't happen, we cover this case to ensure we return to a good state.
+        if (context != null && context.isReleased()) {
+            context = null;
+        }
         decryptInputBuf.reset();
         decryptAADBuf.reset();
```
WillChilds-Klein, line 989: not particularly pertinent to this change, but do we define a max size for our AccessibleByteArrayOutputStream?

if so, and the max is reasonably small, would it be advantageous for performance to back that with a stack-allocated array, and then just maintain a "high watermark" value to track the current size of the buffer (or similarly, track start/end indices and use it like a ring buffer)? way we could avoid allocations on resize as well as the cost of `Arrays.fill(0)` on calls to `reset()` -- just set the watermark or head/tail to 0. i suppose we wouldn't want to leak e.g. plaintext from prior invocations of the cipher, but we could prevent that by ensuring we don't expose the buffer reference anywhere.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -739,6 +643,153 @@ protected synchronized int engineDoFinal(byte[] bytes, int offset, int length, b
...
+            try {
+                actualLength = engineEncryptFinal(bytes, offset, length, buf, 0);
+            } catch (ShortBufferException e) {
+                throw new AssertionError(e);
+            }
+
+            if (actualLength == buf.length) {
+                return buf;
+            } else {
+                return Arrays.copyOf(buf, actualLength);
```
WillChilds-Klein, line 763: what would cause `engineGetOutputSize` to be incorrect in the encryption case? isn't it just len(pt) + padding + tagLen?

WillChilds-Klein (commented):
overall looks good, comments are mostly just questions for my own edification.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -739,6 +643,153 @@ protected synchronized int engineDoFinal(byte[] bytes, int offset, int length, b
...
+            } else {
+                throw new RuntimeCryptoException("Output longer than expected.");
+            }
+        } finally {
+            stateReset();
+        }
+    }
+
+    @Override
+    protected byte[] engineDoFinal(byte[] bytes, int offset, int length)
```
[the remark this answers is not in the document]
SalusaSecondus: I'll add synchronization. Thank you.

We do support multi-shot here, or am I missing something?

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
SalusaSecondus: It should never happen but libcrypto is technically allowed to do this. A prior version had a comment describing this. I'll re-add it.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -924,6 +981,10 @@ private void checkNeedReset() {
 
     // @GuardedBy("this") // Restore once replacement for JSR-305 available
     private void stateReset() {
+        // While this shouldn't happen, we cover this case to ensure we return to a good state.
+        if (context != null && context.isReleased()) {
```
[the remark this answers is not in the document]
SalusaSecondus: No. But while debugging something else I found a hypothetical (but probably impossible flow) which would have resulted in a non-null released context at the end of a crypto operation. I couldn't even figure out how to trigger it as it would have probably required things breaking within libcrypto. But, I spent so long trying to figure out if it was possible that it seemed easier to simply cover the case so that even if it were to happen we'd still recover properly.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
SalusaSecondus: It might be possible, but we don't have a max size in general (some cases we do). I'm not sure how useful a stack allocated array would be. We already maintain a high water mark and thus avoid extra resizes.

There is an interesting experiment I made in a private branch where I implemented an [ArrayCache](https://github.com/SalusaSecondus/amazon-corretto-crypto-provider/blob/buff_pool/src/com/amazon/corretto/crypto/provider/ArrayCache.java) to specifically reuse arrays of common sizes to avoid allocation/GC costs. While I think it is probably useful in the grand scheme of things, it wasn't sufficiently valuable *then* (as there were bigger/easier improvements to be had).

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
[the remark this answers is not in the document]
WillChilds-Klein: > We do support multi-shot here, or am I missing something?

nope, that's my understanding as well. i was calling that out this could be used multi-shot in case you had skipped the `synchronized` qualifier thinking this would only be called in a single-shot manner (and thus not requiring serialization with calls to `engineUpdate`).

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -860,8 +920,8 @@ protected synchronized int engineUpdate(ByteBuffer input, ByteBuffer output) thr
     }
 
     @Override
-    protected int engineDoFinal(ByteBuffer input, ByteBuffer output) throws ShortBufferException,
-        IllegalBlockSizeException, BadPaddingException
+    protected int engineDoFinal(final ByteBuffer input, final ByteBuffer output)
```
WillChilds-Klein, line 923: sorry, missed this earlier. should this be `synchronized` as well?

SalusaSecondus: Doesn't actually matter because it calls into other methods which synchronize.

I think that I'm going to go through and neaten up the synchronization logic since we've evolved so much since it was first written.

SalusaSecondus: I've gone back to the *original* version of this file and found why the synchronization is there. It predates the existence of `NativeContext` (with built in locking). Since we lock at the context level now in all cases (and that's the only state we keep around) we can remove all other synchronization from this class.

Incorrect multi-threaded code is very slightly more likely to result in an exception now (rather than undefined behavior which isn't an exception) but that code was never safe or correct to begin with.

WillChilds-Klein: so we don't need to worry about e.g. race conditions around `decryptInputBuf.write` in multi-threaded [calls to `engineUpdate`](https://github.com/corretto/amazon-corretto-crypto-provider/pull/204/files/adc1300861f02dec5011d5edbc45e34f3d271b0d..60fc618c8e96682329db9c94719137375f3b0883#diff-b98fb0cf29aa3c6687048b3876d672e7d6d3840dd3200b6a2bd775b6eb84785eR439)? the caller is responsible for that synchronization?

WillChilds-Klein (approved):

PR #207, opened 2022-05-23
Change EvpKey to not use EvpKeyContext and become stateless

SalusaSecondus:
*Description of changes:*

Java `EvpKey` is currently backed by a native `EvpKeyContext` object. These contain `EVP_PKEY_CTX` and `EVP_MD_CTX` structures which are stateful, mutable, and thus not thread-safe. Thus, the current implementation of `EvpKey` is not thread safe. This changes `EvpKey` to be backed by `EVP_PKEY` directly (which is never mutated in our code and thus thread safe).

This also has minor related refactors to move us off of `EvpKeyContext` more generally and move `EvpKeyContext` onto our new `*_auto` classes to simplify our pointer management logic.

This is a fast follow for #200 where this issue was identified.

### DIFF
--- a/csrc/keyutils.h
+++ b/csrc/keyutils.h
@@ -12,61 +12,44 @@
 #include <openssl/x509.h>
 #include "util.h"
 #include "env.h"
+#include "auto_free.h"
 
 namespace AmazonCorrettoCryptoProvider {
 
+// This class should generally not be used for new development
+// as it has been replaced by the *_auto classes in auto_free.h
+// The only time this class should be used is when you *need* to keep various EVP objects together.
+// The only currently known *good* use for this class is tracking state when signing/verifying data.
+//
+// Since all but EVP_PKEY are stateful and likely to mutate while being used, this class is not threadsafe.
 class EvpKeyContext {
 public:
-    EvpKeyContext()
-        : digestCtx_(NULL), keyCtx_(NULL), key_(NULL) {}
-
-    virtual ~EvpKeyContext() {
-        if (digestCtx_) {
-            EVP_MD_CTX_destroy(digestCtx_);
-        }
-        if (keyCtx_) {
-            EVP_PKEY_CTX_free(keyCtx_);
-        }
-        if (key_) {
-            EVP_PKEY_free(key_);
-        }
-    }
-
-    EVP_MD_CTX* getDigestCtx() const { return digestCtx_; }
-    EVP_PKEY_CTX* getKeyCtx() const { return keyCtx_; }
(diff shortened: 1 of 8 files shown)


csrc/keyutils.h
```diff
@@ -12,83 +12,63 @@
...
 #include "util.h"
 #include "env.h"
+#include "auto_free.h"
 
 namespace AmazonCorrettoCryptoProvider {
 
+// This class should generally not be used for new development
+// as it has been replaced by the *_auto classes in auto_free.h
+// The only time this class should be used is when you *need* to keep various EVP objects together.
+// The only currently known *good* use for this class is tracking state when signing/verifying data.
```
WillChilds-Klein, line 22: is that something we'd want to (eventually) create special-purpose `auto_`-wrapped buffer for?

csrc/rsa_cipher.cpp
```diff
@@ -61,15 +62,17 @@ JNIEXPORT jint JNICALL Java_com_amazon_corretto_crypto_provider_RsaCipher_cipher
...
 
-        EvpKeyContext *ctx = reinterpret_cast<EvpKeyContext *>(ctxHandle);
-
-        EVP_PKEY_CTX *keyCtx = ctx->getKeyCtx();
-        if (keyCtx == nullptr) {
-            keyCtx = ctx->setKeyCtx(EVP_PKEY_CTX_new(ctx->getKey(), /*engine*/nullptr));
-        }
+        EVP_PKEY* key = reinterpret_cast<EVP_PKEY*>(ctxHandle);
+        // It is unclear if allocating a new EVP_PKEY_CTX for each operation is expensive.
+        // If profiling shows that this is a significant cost when in a hot-path we should figure out a way to
```
WillChilds-Klein, line 67: out of curiosity, what profiling tools do we use for the native portions of ACCP? i.e. what tool would we use to determine whether or not this is a hotspot?

WillChilds-Klein (approved):
nice. this is a lot cleaner than i was expecting.

csrc/keyutils.h
SalusaSecondus: Doesn't seem likely. That wrapper is to make using things from libcrypto easier. This bundles a few of them together so they live together.

csrc/rsa_cipher.cpp
SalusaSecondus: A combination of any benchmark tools we have and [async-profiler](https://github.com/jvm-profiling-tools/async-profiler/) and [perf](https://en.wikipedia.org/wiki/Perf_%28Linux%29).

PR #209, opened 2022-05-24
Fix JAVA_HOME config in Dockerfile.dev

alex-chew:
Dockerfile.dev previously didn't set JAVA_HOME correctly,
so CMake would complain with a "Could NOT find JNI" error when trying to run `./gradlew release`.

*Description of changes:*

* Changes the JAVA_HOME variable in Dockerfile.dev from a build-time argument to an environment variable that persists in the image.
* Adds the JAVA_HOME variable to `.bashrc` instead of `.bash_rc`.


WillChilds-Klein (approved):
thanks for fixing this up alex.

PR #213, opened 2022-05-26
Second attempt to get a FIPS build working

SalusaSecondus:
*Description of changes:*

This is a second approach at getting the FIPS build working (first was in #211). The primary difference is that this PR is broken into smaller self-contained commits to make it easier to reason about and review.

My change doesn't touch the CI to add new tests because I'm not sure the best way to do that. I have manually verified that the tests work with: `./gradlew -DFIPS=true release`


csrc/accp_lc_loader.cpp
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        // We need to load it now
+        jni_string path(env, libPath);
+        void* libPtr = dlopen(path, RTLD_LAZY | RTLD_GLOBAL);
+        if (libPtr == nullptr) {
+            throw_java_ex(EX_RUNTIME_CRYPTO, "Unable to load libcrypto");
+        }
+        dlerror(); // Clear errors
+        dlErrorReturned = dlerror(); // Did the lookup succeed?
```
WillChilds-Klein, line 35: don't you want to check the first invocation of `dlerror()` here?  the second will always be null

> The function dlerror() returns a human readable string describing the most recent error that occurred from dlopen(), dlsym() or dlclose() since the last call to dlerror(). It returns NULL if no errors have occurred since initialization or since it was last called.

https://linux.die.net/man/3/dlerror

csrc/accp_lc_loader.cpp
```diff
@@ -0,0 +1,45 @@
...
+JNIEXPORT jboolean JNICALL Java_com_amazon_corretto_crypto_provider_Loader_loadLibCrypto(
+    JNIEnv* pEnv, jclass, jstring libPath)
+{
+    try {
+        raii_env env(pEnv);
+
+        // Have we already loaded a properly mangled version of AWS-LC?
+        dlerror(); // Clear errors
+        dlsym(RTLD_DEFAULT, CANARY_FUNCTION_NAME); // Return value doesn't matter
+        char* dlErrorReturned = dlerror(); // Did the lookup succeed?
```
WillChilds-Klein, line 22: if i understand this correctly, this block will tell us whether _some library_ presenting the symbol `CRYPTO_library_init` has been loaded by the process, but not necessarily that it's the exact library we desire (AWS-LC and BoringSSL export many identical symbol names). perhaps we want to be more precise in our checks? it  looks like we can use `dlopen` with the `RTLD_NOLOAD` flag set to test whether the specific library we want has already been loaded:

> RTLD_NOLOAD (since glibc 2.2)
Don't load the library. This can be used to test if the library is already resident (dlopen() returns NULL if it is not, or the library's handle if it is resident). This flag can also be used to promote the flags on a library that is already loaded. For example, a library that was previously loaded with RTLD_LOCAL can be reopened with RTLD_NOLOAD | RTLD_GLOBAL. This flag is not specified in POSIX.1-2001.

https://linux.die.net/man/3/dlerror

csrc/accp_lc_loader.cpp
```diff
@@ -0,0 +1,45 @@
...
+        if (dlErrorReturned == nullptr) {
+            // Already loaded
+            return JNI_FALSE; // We didn't do anything
+        }
+
+        // We need to load it now
+        jni_string path(env, libPath);
+        void* libPtr = dlopen(path, RTLD_LAZY | RTLD_GLOBAL);
```
WillChilds-Klein, line 30: do we want this to be global? what libraries are we planning on subsequently loading?

> RTLD_GLOBAL
> The symbols defined by this library will be made available for symbol resolution of subsequently loaded libraries.

src/com/amazon/corretto/crypto/provider/DebugFlag.java
```diff
@@ -23,7 +23,12 @@ enum DebugFlag {
      * Increases the verbosity of logs.
      * May still need to be combined with increasing the log level of your configured logger.
      */
-    VERBOSELOGS;
+    VERBOSELOGS,
+    /**
+     * By default ACCP attempts to delete the native libraries it uses immediately after loading.
+     * When this value is set it does not delete them until the JVM is done running.
```
WillChilds-Klein, line 29: as in, unlink them on the filesystem? if the JVM process is restarted, wouldn't the libraries be gone?

also, why would we want to leave them around in the debug case?

src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -243,7 +295,7 @@ private static void assertVersionMatch() {
      * /dev/urandom. Clearly this won't work when we start supporting windows systems. We are intentionally taking
      * as few dependencies here as possible.
      */
-    private synchronized static Path createTmpFile(final String prefix, final String suffix) throws IOException {
+    private static synchronized Path createTmpFile(final String prefix, final String suffix) throws IOException {
```
WillChilds-Klein, line 298: can't this still be `static`? your addition on L344 doesn't reference any instance variables.

src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -121,63 +127,32 @@ static String getProperty(String propertyName, String def) {
...
-                    } catch (final Exception realException) {
-                        loadingException = realException;
-                    } catch (final Throwable realError) {
-                        loadingException = new RuntimeCryptoException("Unable to load native library", realError);
-                    } finally {
-                        Files.delete(libPath);
-                    }
-                } else {
-                    loadingException = new RuntimeCryptoException("Skipping bundled library null mapped name");
+                final Path libCryptoPath = writeResourceToTemporaryFile(System.mapLibraryName(LIBCRYPTO_NAME));
```
WillChilds-Klein, line 130: why copy this to a temp file and then load from the temp file instead of just loading the original directly?

src/com/amazon/corretto/crypto/provider/DebugFlag.java
SalusaSecondus: Yes, as in unlink them from the filesystem.

Without the flag the flow looks like this:
1. Write files to disk
2. Load libraries from files
3. Try to delete files from disk (but don't worry if you fail)
4. (JVM Shutdown) Automatically delete files if still thre

This flag causes step 3 above to be skipped so that the files live until the JVM terminates. This makes it easier to do *native* debugging (say, with GDB) because your debugger can actually see the files which contain the running code.

src/com/amazon/corretto/crypto/provider/Loader.java
SalusaSecondus: You can only load libraries from actual files. This library is stored in our JAR and thus must be extracted and written to the filesystem before it can be loaded.

src/com/amazon/corretto/crypto/provider/Loader.java
SalusaSecondus: It is still static. I just had the order of `static` and `synchronized` backwards for best Java style.

SalusaSecondus:
The `accp-ci-overkill-dieharder` failure was a dieharder-threads failure (and these are flakey). For the following reasons, I do not believe we need to worry:
1. No changes specific to the RNG have been introduced since the last successful build. (e.g., If these changes broke something, they'd break many more things than just this test.)
2. dieharder-threads passed in the legacy `accp_github_overkill) job.

build.gradle
```diff
@@ -63,9 +64,13 @@ task buildAwsLc {
             workingDir awslcSrcPath
             executable 'cmake'
             args "-B${cMakeBuildDir}"
-            args '-DBUILD_SHARED_LIBS=OFF'
+            args '-DBUILD_SHARED_LIBS=ON'
```
justsmth, line 67: Should building the shared library be contingent upon `isFIps`.  Is there a reason to prefer static over shared libraries in this context?

SalusaSecondus: Moving some conversation into this thread:

> I'm removing all static-linking logic as part of the FIPS work. This is for the following reasons:
> 1. We must support dynamic linking of libcrypto for FIPS
> 2. Supporting more than one linking mode increases complexity
> 3. The new CI (and potentially other systems) breaks statically linking an object into a shared object. The tests should be failing but are not. This is because ACCP (and some of our tests) are too robust. I'll fix this.

As follow ups from that:
1. I "fixed" the tests so that they properly fail when linking (static or dynamic) isn't don't quite right.
2. I think that I figured out why the new CI tests were not doing proper static linking but since it simplifies things to consolidate on a single linking strategy, that is not mot.

Static linking has two benefits and two drawbacks.
**Pros:**
* Smaller binary as we can remove unused symbols
* Easier loading story as everything ends up within a single shared object.

**Cons:**
* Cannot be used with AWS-LC FIPS
* More complicated (and fragile) build process.

Since we *must* support dynamic linking, it makes sense for us to reduce complexity by just doing that one mode instead of both.

src/com/amazon/corretto/crypto/provider/Loader.java
WillChilds-Klein: ah, whoops. misread that.

src/com/amazon/corretto/crypto/provider/Loader.java
WillChilds-Klein: got it. i was missing that this is reading directly from the JAR archive rather than unzipped contents on the filesystem.

csrc/accp_lc_loader.cpp
WillChilds-Klein: fixed this by consolidating logic into `libraryLoadError`

WillChilds-Klein (approved):

src/com/amazon/corretto/crypto/provider/Loader.java
alexw91: I wanted to make sure that I'm correctly understanding everything that is happening here, since this line is doing several things at once.

1. `System.mapLibraryName(LIBCRYPTO_NAME)` returns the platform specific resource name of the crypto library ACCP has a dependency on. (On Linux this would be `libcrypto.so`)
2. `writeResourceToTemporaryFile()` takes that library name, and calls Java's `Loader.class.getResourceAsStream()` which takes that resource name, and **looks inside the ACCP JAR file** for a resource with that name first.  
3. A `libcrypto.so` resource is found inside `AmazonCorrettoCryptoProvider.jar`, and is streamed to a temp file that is written to disk
4. The path to this temp file is returned, and is later passed down `loadLibCrypto()` where from C we load that crypto library into the current process.

Steps 2 and 3 were not immediately clear from reading just this line, and I wasn't sure if we were searching for currently installed resources on the platform (Eg searching the `/lib/x86_64-linux-gnu/` directory on Linux) or searching the resources inside the ACCP Jar file.

PR #216, opened 2022-06-02
Split CI into different jdk versions

samuel40791765:
*Issue #, if available:*

Resolves `CryptoAlg-1102`

*Description of changes:*

This PR spits the JDK tests across different dimensions according to its version. The `TEST_JAVA_HOME` variable is added to each build to specify the path to the java version we want to test upon. The way AWS-LC's CI deals with this is to use a specific docker image for each version/compiler, but this method should suit ACCP better. ACCP still needs a JDK version > 10 for building, and JDK versions lower than that are only used for testing.


samuel40791765:
Tests in new CI are failing because of the new test script behavior. 
PR in local fork https://github.com/samuel40791765/amazon-corretto-crypto-provider/pull/6 shows that this works as expected.

tests/ci/run_accp_basic_tests.sh
```diff
@@ -3,11 +3,26 @@ set -exo pipefail
...
+# Parse and check which JDK version we're testing upon.
+version=$($TEST_JAVA_HOME/bin/java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1)
 
-echo "Testing ACCP standard release with corretto 8."
-./gradlew -DTEST_JAVA_HOME=/usr/lib/jvm/java-1.8.0-amazon-corretto test
+# The JDK version should be least 10 for a regular ACCP build. We can
+# still test on older versions with the TEST_JAVA_HOME property.
+if (( "$version" <= "10" )); then
+	./gradlew -DTEST_JAVA_HOME=$TEST_JAVA_HOME test
+	exit 0
```
WillChilds-Klein, line 13: i don't see `-e` set at the top of this script... shouldn't you `exit $?` to pick up the result of your gradle invocation?

tests/ci/run_accp_basic_tests.sh
```diff
@@ -3,11 +3,26 @@ set -exo pipefail
...
+# Assign the JDK version we're testing as the system's default JDK and
+# assign JAVA_HOME variable to the path. Otherwise, Ubuntu will
+# default to the newest version of Java on the system.
+export JAVA_HOME=$TEST_JAVA_HOME
+export PATH=$JAVA_HOME/bin:$PATH
+
+if (( "$version" >= "17" )); then
+	# This flag is necessary in Java17+ for certain unit tests to
+	# perform deep reflection on nonpublic members.
+    ./gradlew -DTEST_JAVA_MAJOR_VERSION=$version release
```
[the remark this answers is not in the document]
samuel40791765: Seems like this part does a version check so I was overengineering it
https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/CMakeLists.txt#L623-L632

tests/ci/run_accp_basic_tests.sh
samuel40791765: `-e` is set above the copyright comments, but either way, using `exit $?` is much cleaner.

WillChilds-Klein (approved):

tests/ci/run_accp_basic_tests.sh
WillChilds-Klein: ah whoops, i forgot to expand the diff boundary. thanks.

PR #217, opened 2022-06-03
MacOS CI in ACCP with Github Actions

samuel40791765:
*Issue #, if available:*
`CryptoAlg-1103`

*Description of changes:*
Sets up the MacOS CI framework for ACCP. Functionality will be disabled for now, and should be enabled when support is added. Corretto 8 hasn't been included in the workflow file because retrieving and saving the `JAVA_HOME` variable installed by corretto8 will result in added complications to the workflow. We can look to add it if it's absolutely needed.

### DIFF
--- a/.github/workflows/macos_ci.yml
+++ b/.github/workflows/macos_ci.yml
@@ -21,38 +21,28 @@ jobs:
       with:
         submodules: true
 
+    # Test on Corretto 11.
     - name: Set up corretto 11
       uses: actions/setup-java@v3
       with:
         java-version: '11'
         distribution: 'corretto'
-    - name: Build ${{ env.PACKAGE_NAME }} in corretto 11
+    - name: Build and run tests for ${{ env.PACKAGE_NAME }} in corretto 11
+      env:
+        TEST_JAVA_HOME: ${{ env.JAVA_HOME }}
       run: |
         ./tests/ci/run_accp_basic_tests.sh
-    - name: Run integration tests for ${{ env.PACKAGE_NAME }} in corretto 11
-      run: |
         ./tests/ci/run_accp_test_integration.sh
 
-    - name: Set up corretto 8
-      uses: actions/setup-java@v3
-      with:
-        java-version: '8'
-        distribution: 'corretto'
-    - name: Build ${{ env.PACKAGE_NAME }} in corretto 8
-      run: |
-        ./tests/ci/run_accp_basic_tests.sh
-    - name: Run integration tests for ${{ env.PACKAGE_NAME }} in corretto 8
-      run: |
-        ./tests/ci/run_accp_test_integration.sh
-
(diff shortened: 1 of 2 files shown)


samuel40791765:
Activation may need this PR to be merged first, then changed in the Actions tab configuration. This file works in https://github.com/samuel40791765/amazon-corretto-crypto-provider/pull/10

WillChilds-Klein (commented):
> Corretto 8 hasn't been included in the workflow file because retrieving and saving the JAVA_HOME variable installed by corretto8 will result in added complications to the workflow.

what's the nature of these complications? java8 is pretty ubiquitous... why can't we use `java-version`, `distribution` etc. directives to access `env.JAVA_HOME` correctly?

samuel40791765:
> what's the nature of these complications? java8 is pretty ubiquitous... why can't we use java-version, distribution etc. directives to access env.JAVA_HOME correctly?

My main concern was that the JDK version had to be at least 10 for a full ACCP build. Installing java8 overwrites the `JAVA_HOME` variable in Github Actions and there initially didn't seem to be a way to save the java8 path from the environment to be used later on. 
I've just learned that there is a way to [save runtime variables in the workflow](https://microsoft.github.io/code-with-engineering-playbook/continuous-delivery/devops-provider-recipes/github-actions/runtime-variables/runtime-variables/), so I've added the corretto8 testing dimension as well now :)

.github/workflows/macos_ci.yml
```diff
@@ -0,0 +1,69 @@
...
+        distribution: 'corretto'
+    - name: Build and run tests for ${{ env.PACKAGE_NAME }} in corretto 11
+      env:
+        TEST_JAVA_HOME: ${{ env.JAVA_HOME }}
+      run: |
+        ./tests/ci/run_accp_basic_tests.sh
+        ./tests/ci/run_accp_test_integration.sh
+
+    # Test on Corretto 8.
+    - name: Build and run tests for ${{ env.PACKAGE_NAME }} in corretto 8
```
WillChilds-Klein, line 51: nit: move this Corrtto 8 testing up closer to the setup?

also, this is a lack of GitHub CI familiarity on my part, but why do you need to break out the various setup and test tasks into separate jobs? also, is each `name`d job run sequentially or concurrently?

WillChilds-Klein (dismissed):

.github/workflows/macos_ci.yml
samuel40791765: Moved closer 👍 
`macOS` is actually the only job that's being run. Each job is points to a single Github Action Webhook. The `name`s actually refer to each `step` within the `job`, so they're run sequentially. I've looked into trying to run them concurrently, but `1 job=1 webhook`, so it bloats up the CI.

WillChilds-Klein (approved):

PR #218, opened 2022-06-07
Add FIPS support to ACCP CI

samuel40791765:
*Issue #, if available:*
`CryptoAlg-1105`

*Description of changes:*
After merging https://github.com/corretto/amazon-corretto-crypto-provider/commit/084363e964d33d00b4e5a52e1043c7550e300365, we now have support for FIPS in ACCP. This PR adds FIPS support to our CI to ensure we have continuous integration tests running on the FIPS dimension. CI Readme has also been updated to reflect the newest state.

Example of run here: https://github.com/samuel40791765/amazon-corretto-crypto-provider/pull/13
Links to build can be found under the SIM ticket.

### DIFF
--- a/tests/ci/run_accp_dieharder.sh
+++ b/tests/ci/run_accp_dieharder.sh
@@ -9,7 +9,7 @@ while [[ $# -gt 0 ]]; do
     case ${1} in
     --fips)
       testing_fips=${2}
-	  shift
+      shift
       ;;
     *)
       echo "${1} is not supported."
--- a/tests/ci/run_accp_dieharder.sh
+++ b/tests/ci/run_accp_dieharder.sh
@@ -8,8 +8,7 @@ testing_fips=false
 while [[ $# -gt 0 ]]; do
     case ${1} in
     --fips)
-      testing_fips=${2}
-      shift
+      testing_fips=true
       ;;
     *)
       echo "${1} is not supported."
--- a/tests/ci/run_accp_dieharder.sh
+++ b/tests/ci/run_accp_dieharder.sh
@@ -3,5 +3,11 @@ set -exo pipefail
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+# Testing non-FIPS is the default.
+testing_fips=false
+if [[ "${1}" == "--fips" ]]; then
+	testing_fips=true
+fi
+
 echo "Testing ACCP dieharder tests."
-./gradlew dieharder
+./gradlew -DFIPS=$testing_fips dieharder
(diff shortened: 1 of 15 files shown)


WillChilds-Klein (commented):
nice!

1. overall, changes look good to me. only sticking point is questions about go versions
2. can we also hold off on merging this until MacOS CI dimensions get merged? would be great to have FIPS dimension for MacOS as well

tests/ci/run_accp_dieharder.sh
```diff
@@ -3,5 +3,11 @@ set -exo pipefail
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+# Testing non-FIPS is the default.
+testing_fips=false
+if [[ "${1}" == "--fips" ]]; then
```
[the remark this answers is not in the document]
samuel40791765: `getopts` only supports one character, but I've added support for what you're looking for. I think it's a valid concern, it also paves way to easily add support for testing new flags.

samuel40791765 (commented):
> 2. can we also hold off on merging this until MacOS CI dimensions get merged? would be great to have FIPS dimension for MacOS as well

Sure, I'll add a MacOS FIPS framework dimension in this PR when that gets merged. Just briefly addressed your other comments.

samuel40791765:
Added the FIPS Mac CI. Also adjusted the settings of the repository so that only workflows that we created can be spun up in PR (same as AWS-LC).

tests/ci/run_accp_dieharder.sh
```diff
@@ -3,5 +3,22 @@ set -exo pipefail
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+# Testing non-FIPS is the default.
+testing_fips=false
+while [[ $# -gt 0 ]]; do
+    case ${1} in
+    --fips)
+      testing_fips=${2}
+      shift
```
WillChilds-Klein, line 12: don't you need to call `shift` twice -- one for the `--fips` one for the `true`/`false`?

related nit: do you really need the `true`/`false` argument to the `--fips` flag? conventionally, boolean values are passed by mere presence of absence of a `--boolean-flag`. to simplify it a bit, you could do

```
testing_fips=false
while [[ $# -gt 0 ]]; do
    case ${1} in
    --fips)
      testing_fips=true
      shift
      ;;
    *)
      echo "${1} is not supported."
      exit 1
      ;;
    esac
    # Check next option -- key/value.
    shift
done

echo "Testing ACCP dieharder tests."
./gradlew dieharder
./gradlew -DFIPS=$testing_fips dieharder
```

samuel40791765: > don't you need to call shift twice -- one for the --fips one for the true/false?

The `shift` before the done is second `shift.

I think the `--boolean-flag` makes perfect sense. Adjusted it accordingly.

WillChilds-Klein: ah, whoops. sorry i missed the `shift` on line 20.

PR #223, opened 2022-06-16
Enable build badges for CI

WillChilds-Klein:
*Issue #, if available:* n/a

*Description of changes:*

Enable build badges for CI

I'm posting the first revision of this PR for review, once this change is reviewed, I'll add a commit to this PR updating the README.


WillChilds-Klein:
ok, i spun up a CI stack in my dev account, and the build badge being enabled is visible in the lower right corner of this screenshot:

Screen Shot 2022-06-16 at 7 46 20 PM

if this looks OK, I'll go ahead and update the CI stack in ACCP's account and update this PR with another commit pointing the README.md to the new badge URLs.

**REVIEWERS:** please either like this comment to give the go-ahead to update CI in the ACCP account.

samuel40791765 (approved):

tests/ci/cdk/cdk/accp_github_ci_stack.py
```diff
@@ -27,6 +27,8 @@ def __init__(self,
             fetch_submodules=True,
             webhook_filters=[
                 codebuild.FilterGroup.in_event_of(
+                    codebuild.EventAction.PUSH,
+                    codebuild.EventAction.PULL_REQUEST_MERGED,
```
samuel40791765, line 31: Are both of these necessary, or would `PUSH` be enough to suffice?

tests/ci/cdk/cdk.context.json
```diff
@@ -1 +0,0 @@
-{}
```
samuel40791765, line 1: I know we talked about not updating this file. But will the cdk run work with it being deleted, or will it be automatically generated again?

tests/ci/cdk/cdk/accp_github_ci_stack.py
WillChilds-Klein: i'm not 100% sure, but the fact that they bothered to break them out into separate fields indicates to me that GitHub differentiates between a "push" action and a "merge form PR" action. in either case, i feel there's no potential harm in including both.

tests/ci/cdk/cdk.context.json
WillChilds-Klein: i should have called this out in the README. i tested that by deleting this file, then running the `update-ci` action, which worked. it even re-created this file, which should subsequently be `.gitignore`d after this is merged.

samuel40791765 (approved):

tests/ci/cdk/cdk.context.json
samuel40791765: That's actually pretty neat, looks like I should apply to AWS-LC as well.

PR #224, opened 2022-06-21
Dynamically load ACCP with AWS-LC using RPath

alexw91:
**Issue #, if available:** N/A

**Description of changes:**

Updates how ACCP dynamically loads it's dependencies. The changes in this PR include:
 1. `libamazonCorrettoCryptoProvider.so` now declares a shared library dependency on `libcrypto.so`
 2. `libamazonCorrettoCryptoProvider.so` now contains an `RPATH` value of `$ORIGIN`, which tells the runtime dynamic loader to look in the same directory as `libamazonCorrettoCryptoProvider.so` for any shared library dependencies before looking in system shared library directories
 3.  ACCP's `Loader.java` class now creates a temporary directory, copies `libamazonCorrettoCryptoProvider.so` and `libcrypto.so` into that temp directory, and calls `System.load()` on that libACCP that was just copied.
 4. Various refactorings in `Loader.java` to break up large functions into smaller ones.
 5. Deletion of previous `accpLcLoader` shared library that loaded AWS-LC libcrypto symbols into the global object group, which would break any other JNI objects that required the use of a different version of libcrypto
 6. Improved checks for ensuring that the native ACCP libraries are loaded before attempting to use that functionality
 7. Equivalent RPath changes for test executables that depend on libACCP.


src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -177,63 +162,112 @@ static String getProperty(String propertyName, String def) {
...
-            while (read >= 0) {
-                os.write(buffer, 0, read);
-                read = is.read(buffer);
-            }
-            os.flush();
-        }
-        return libPath;
+    private static void tryLoadLibraryFromSystem() {
+        /**
+         * Attempt to load library using system's default shared library lookup paths
```
SalusaSecondus, line 243: Should we still call `validateLibcryptoExactVersionMatch();`? After all, if the versions don't match, how do you know we'll work and not crash horribly?

Do we need a fuzzier match for the system library case?

src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -289,20 +324,78 @@ private static void assertVersionMatch() {
...
+    private static byte[] bootstrapRng(int numBytes) throws IOException {
         final Path urandomPath = Paths.get("/dev/urandom");
         if (!Files.exists(urandomPath)) {
             throw new AssertionError("/dev/urandom must exist for bootstrapping");
         }
-        final Path tmpDir = Paths.get(System.getProperty("java.io.tmpdir"));
-        if (!Files.isDirectory(tmpDir)) {
-            throw new AssertionError("java.io.tmpdir is not valid: " + tmpDir);
+
+        final byte[] rndBytes = new byte[numBytes]; // Default java tmp files use this much entropy
```
SalusaSecondus, line 376: Comment is no longer accurate. Please remove.

src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -289,20 +324,78 @@ private static void assertVersionMatch() {
...
+        }
+        maybeDelete(tmpDirectory);
+    }
+
+    private static Path writeJarResourceToTemporaryFile(Path tempDirectory, final String resourceFileName) throws IOException {
+        final Path tempResourceFilePath = tempDirectory.resolve(resourceFileName);
+
+        try (InputStream is = Loader.class.getResourceAsStream(resourceFileName);
+             OutputStream os = Files.newOutputStream(tempResourceFilePath, StandardOpenOption.CREATE,
+                     StandardOpenOption.WRITE, StandardOpenOption.TRUNCATE_EXISTING)) {
```
SalusaSecondus, line 350: We need to ensure that file permissions are properly set at creation time (like with the temporary directory).

We also need to fail if an a file already exists with this name and not overwrite it.

alexw91: Fixed

src/com/amazon/corretto/crypto/provider/Loader.java
alexw91: I'm assuming that if someone is manually loading ACCP from system libraries, then the person installing the libraries into the system directories will be ensuring that ACCP and LibCrypto work together, so I added the fuzzier version check if loading ACCP from system libraries. 

In practice however, AWS-LC is not regularly updating `OPENSSL_VERSION_NUMBER` so this check may not provide as much value as I initially intended. I've opened a PR to AWS-LC to provide an API to retrieve their `AWSLC_API_VERSION` value at runtime

src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -289,68 +342,116 @@ private static void assertVersionMatch() {
...
+    }
+
+    private static Path writeJarResourceToTemporaryFile(Path tempDirectory, final String resourceFileName) throws IOException {
+        final Path tempResourceFilePath = tempDirectory.resolve(resourceFileName);
+
+        // Create a new temporary file. If one already exists this will throw FileAlreadyExistsException
+        Files.createFile(tempResourceFilePath, SELF_OWNER_FILE_PERMISSIONS);
+
+        try (InputStream is = Loader.class.getResourceAsStream(resourceFileName);
+             OutputStream os = Files.newOutputStream(tempResourceFilePath, StandardOpenOption.CREATE,
```
SalusaSecondus, line 370: Let's remove `CREATE` as it's no longer required (and thus misleading).

src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -289,68 +342,116 @@ private static void assertVersionMatch() {
...
+            String resourceFileName = System.mapLibraryName(jarResource);
+            maybeDelete(tmpDirectory.resolve(resourceFileName));
+        }
+        maybeDelete(tmpDirectory);
+    }
+
+    private static Path writeJarResourceToTemporaryFile(Path tempDirectory, final String resourceFileName) throws IOException {
+        final Path tempResourceFilePath = tempDirectory.resolve(resourceFileName);
+
+        // Create a new temporary file. If one already exists this will throw FileAlreadyExistsException
```
SalusaSecondus, line 366: Please expand this comment to explain why it is important that we create first:
We need to ensure that we create the file and set permissions atomically so that an adversary cannot put a symlink or file here which would cause us to write (or read) to an arbitrary attacker controlled location.

SalusaSecondus (commented):
Very minor tweaks.

csrc/loader.cpp
```diff
@@ -60,3 +64,47 @@ JNIEXPORT jstring JNICALL Java_com_amazon_corretto_crypto_provider_Loader_getNat
...
+
+        return JNI_TRUE;
+    } catch (java_ex &ex) {
+        ex.throw_to_java(pEnv);
+    }
+
+    return JNI_FALSE;
+}
+
+JNIEXPORT jboolean JNICALL Java_com_amazon_corretto_crypto_provider_Loader_validateLibcryptoFuzzyVersionMatch(JNIEnv* pEnv, jclass)
```
WillChilds-Klein, line 90: minor style nit: decent amount of code duplicated between fuzzy/exact matchers. you could pull out most of each of these functions into a shared helper that takes compile-time and runtime versions and returns a boolean.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -266,15 +266,16 @@ public AmazonCorrettoCryptoProvider() {
 
         Utils.optionsFromProperty(ExtraCheck.class, extraChecks, "extrachecks");
 
-        if (!Loader.IS_AVAILABLE && DebugFlag.VERBOSELOGS.isEnabled()) {
```
WillChilds-Klein, line 269: good catch. was this a bug? i.e. we wouldn't have returned early if loader was unavailable but verbose logging hadn't been enabled?

src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -289,68 +342,116 @@ private static void assertVersionMatch() {
...
+        final StringBuilder privateTempDir = new StringBuilder(prefix);
+
+        for (byte b : rndBytes) {
+            // We convert to an unsigned integer first to avoid sign-bit extension when converting to hex.
+            String hexByte = Integer.toHexString(Byte.toUnsignedInt(b));
+            if (hexByte.length() == 1) {
+                privateTempDir.append('0');
+            }
+            privateTempDir.append(hexByte);
+        }
```
WillChilds-Klein, line 432: do we care about uniqueness, unpredictability, or both here?

if all we care about is uniqueness, we could do something stupid like current millis time + JVM pid or something instead of taking the dependency on `/dev/urandom`

src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -289,68 +342,116 @@ private static void assertVersionMatch() {
...
+            attempt++;
+            try {
+                final Path result = Files.createDirectory(privateDirFullPath, SELF_OWNER_FILE_PERMISSIONS);
+                if (DebugFlag.VERBOSELOGS.isEnabled()) {
+                    LOG.log(Level.FINE, "Created temporary library directory");
                 }
+                result.toFile().deleteOnExit();
+                return result;
+            } catch (final FileAlreadyExistsException ex) {
+                // We ignore and retry this exception
```
WillChilds-Klein, line 448: why retry this without taking any action? if the file already exists, we'd need to delete it before we'd expect our directory creation to work, right?

src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -289,68 +342,116 @@ private static void assertVersionMatch() {
...
+            int read = is.read(buffer);
+            while (read >= 0) {
+                os.write(buffer, 0, read);
+                read = is.read(buffer);
+            }
+            os.flush();
+        }
+
+        // Ensure we delete any temp files on exit
+        tempResourceFilePath.toFile().deleteOnExit();
```
WillChilds-Klein, line 382: here and in the directory creation routine, should we guard the `deleteOnExit()` call with `!DebugFlag.PRESERVE_NATIVE_LIBRARIES.isEnabled()`?

src/com/amazon/corretto/crypto/provider/Loader.java
SalusaSecondus: Unpredictabity is better and we should do as best we can. In the windows world we will probably need to settle on current millis combined with PID and nanotime (though how to combine, I don't know).

src/com/amazon/corretto/crypto/provider/Loader.java
SalusaSecondus: Random generation *should* be in the retry loop so it goes for different names. If not then that's a bug.

src/com/amazon/corretto/crypto/provider/Loader.java
SalusaSecondus: No. We currently always want to delete on exit. That flag simply prevents immediate deletion (rather than deletion when the JVM terminates).

CMakeLists.txt
```diff
@@ -300,6 +295,11 @@ add_library(
         ${JNI_HEADER_DIR}/generated-headers.h
 )
 
+# We MUST set ACCP's RPATH to "$ORIGIN" so that the runtime dynamic loader will look in the same directory as ACCP for
+# any shared library dependencies BEFORE looking for a system-installed depencencies (such as libcrypto). If RPath is
+# not set, the loader may attempt to load an incompatible system-installed libcrypto and cause runtime crashes.
```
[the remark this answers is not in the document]
alexw91: I made that change locally, but reverted it since it didn't work as expected. The `libcrypto.so` file embeds a `SONAME` attribute within the ELF format. This `SONAME` attribute is what is actually used by the dynamic linker and loader, and the filename is ignored. Renaming `libcrypto.so` to `libamazonCorrettoCrypto.so` before the ACCP build began did not have an effect on the final libAACP shared library, which still declared a shared library dependency on `libcrypto.so`.

```
$ readelf -d libcrypto.so | grep SONAME
6: 0x000000000000000e (SONAME)             Library soname: [libcrypto.so]
```

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
alexw91: Yes, this was a bug that I ran into when debugging partially completed versions of this PR that did not load the native libraries correctly. In practice this hasn't been an issue since ACCP was always loading native libraries correctly.

src/com/amazon/corretto/crypto/provider/Loader.java
alexw91: We should aim for cryptographically strong unpredictability. My understanding is that we should be resilient against an attacker that:
1. Is on the same host, with the privileges of a normal user
2. Can read/write to the same system temporary directory (`/tmp/*`) that ACCP is using
3. Can time their reads/writes with very high precision and throughput

If we only use currentTime + PID we don't have unpredictability, then an attacker could theoretically still mount a DDOS attack on AACP by creating empty directories with the same names that ACCP is attempting to create.

My idea (that I'm not sure would work), would be to have `bootstrapRng()` attempt to use the SunJCE SecureRandom implementation for bootstrapping, but I'm not sure if we'll run into issues where ACCP is being initialized and installed before the SunJCE is ready or if it is not available.

CMakeLists.txt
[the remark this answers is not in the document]
WillChilds-Klein: ah, interesting. that makes sense. thank you for the explanation.

src/com/amazon/corretto/crypto/provider/Loader.java
SalusaSecondus: For context, when [conscrypt writes out its shared object](https://github.com/google/conscrypt/blob/ca6485b689219323e91d3bbe0410715b5b2d687f/openjdk/src/main/java/org/conscrypt/NativeLibraryLoader.java#L189-L199) its temporary files are [based entirely on the current time](https://github.com/google/conscrypt/blob/ca6485b689219323e91d3bbe0410715b5b2d687f/openjdk/src/main/java/org/conscrypt/Platform.java#L115-L144). While I believe that we can do better, they're solution is likely sufficient.

src/com/amazon/corretto/crypto/provider/Loader.java
alexw91: Nice catch! I don't think we should delete the directory since if it already exists then it was created by something else and may be a collision. We already call `result.toFile().deleteOnExit();` so we will automatically delete our own temp directory. 

I've moved the call to `bootstrapRng()` to be inside the retry loop so that a different directory name is generated on each retry.

src/com/amazon/corretto/crypto/provider/Loader.java
alexw91: Thanks! Added that check

alexw91: Done.

src/com/amazon/corretto/crypto/provider/Loader.java
alexw91: I'll leave the question for what to do when we don't have `/dev/urandom` for a later time. We don't need to answer it until we pick up https://github.com/corretto/amazon-corretto-crypto-provider/issues/48

WillChilds-Klein (dismissed):

csrc/loader.cpp
```diff
@@ -25,6 +25,11 @@
 // 0x1010107f == v1.1.1g release
 #define LIBCRYPTO_MAJOR_MINOR_VERSION_MASK 0xFFF00000
 
+#define LIBCRYPTO_EXACT_VERSION_MATCH       0
+#define LIBCRYPTO_FUZZY_VERSION_MATCH       1
```
SalusaSecondus, line 29: NitPick: Why not?
```suggestion
#define LIBCRYPTO_EXACT_VERSION_MATCH       false
#define LIBCRYPTO_FUZZY_VERSION_MATCH       true
```

csrc/loader.cpp
```diff
@@ -65,20 +70,28 @@ JNIEXPORT jstring JNICALL Java_com_amazon_corretto_crypto_provider_Loader_getNat
...
+    if (libcrypto_compiletime_version != libcrypto_runtime_version) {
+        memset(accp_loader_exception_msg, 0, sizeof(accp_loader_exception_msg));
+        snprintf(accp_loader_exception_msg, sizeof(accp_loader_exception_msg),
+                "Runtime libcrypto version does not match compile-time version. Expected: 0x%08lX , Actual: 0x%08lX",
+                libcrypto_compiletime_version, libcrypto_runtime_version);
+        throw java_ex(EX_RUNTIME_CRYPTO, accp_loader_exception_msg);
+    }
+}
 
+JNIEXPORT jboolean JNICALL Java_com_amazon_corretto_crypto_provider_Loader_validateLibcryptoExactVersionMatch(JNIEnv* pEnv, jclass)
```
SalusaSecondus, line 91: NitPick: I'd just expose a single native method and take in a boolean as a parameter.

csrc/loader.cpp
```diff
@@ -25,6 +25,11 @@
 // 0x1010107f == v1.1.1g release
 #define LIBCRYPTO_MAJOR_MINOR_VERSION_MASK 0xFFF00000
 
+#define LIBCRYPTO_EXACT_VERSION_MATCH       0
+#define LIBCRYPTO_FUZZY_VERSION_MATCH       1
+
+static char accp_loader_exception_msg[256] = {0};
```
SalusaSecondus, line 31: This is technically not thread safe anymore. Could you move this back (non statically) inside `accpValidateLibcryptoVersion` please?

SalusaSecondus (commented):
One (minor) thread safety issue and some nitpicks you can choose to ignore.

csrc/loader.cpp
WillChilds-Klein: does throwing an exception result in copying exception data over to the JVM heap or some other location? if not, and we stack-allocate the message buffer within the scope of `accpValidateLibcryptoVersion` method, de-referencing the buffer outside of `accpValidateLibcryptoVersion`'s scope would violate the buffer's lifetime.

SalusaSecondus: Oh dear. You're right.

Hypothetically this could be a bigger issue. We should probably rewrite `java_ex` to always (and immediately) copy the string into local memory (see `env.h`, it should be easy). Can we pull that into this change as there really isn't a safe path forward without that?

alexw91: Fixed. `java_ex` now keeps a local copy of the message passed to it.

csrc/env.h
```diff
@@ -50,8 +51,7 @@ class java_ex {
         jthrowable m_java_exception;
 
         const char *m_java_classname;
-        const std::string m_message;
-        const char *m_message_cstr;
```
SalusaSecondus, line 54: We didn't previously have a maximum message length. While this works why not just always store the message in `m_message` which is an actual C++ object and thus will properly manage its memory as we need?

csrc/env.h
```diff
@@ -61,16 +61,22 @@ class java_ex {
...
         java_ex(const char *java_classname, const char *message) COLD
-            : m_java_exception(nullptr), m_java_classname(java_classname), m_message(), m_message_cstr(message)
-        { capture_trace(); }
+            : m_java_exception(nullptr), m_java_classname(java_classname) {
+
+            capture_trace();
+
+            if (message != NULL) {
+                size_t amount_to_copy = strnlen(message, (sizeof(m_message_cstr) - 1));
+                std::strncpy(m_message_cstr, message, amount_to_copy);
```
SalusaSecondus, line 74: While I believe that we should be doing this all through `m_message` rather than `m_message_cstr`, why is this not the simpler:
```c++
std::strncpy(m_message_cstr, message, sizeof(m_message_cstr) - 1);
```

csrc/loader.cpp
```diff
@@ -70,40 +65,26 @@ JNIEXPORT jstring JNICALL Java_com_amazon_corretto_crypto_provider_Loader_getNat
...
-JNIEXPORT jboolean JNICALL Java_com_amazon_corretto_crypto_provider_Loader_validateLibcryptoExactVersionMatch(JNIEnv* pEnv, jclass)
+JNIEXPORT jboolean JNICALL Java_com_amazon_corretto_crypto_provider_Loader_validateLibcryptoVersion(JNIEnv* pEnv, jclass, jboolean jFuzzyMatch)
 {
-    try {
-        accpValidateLibcryptoVersion(LIBCRYPTO_EXACT_VERSION_MATCH);
-        return JNI_TRUE;
-    } catch (java_ex &ex) {
-        ex.throw_to_java(pEnv);
-    }
+    bool fuzzyMatch = (jFuzzyMatch == JNI_TRUE);
```
SalusaSecondus, line 70: Nice!

We might (later) want to add the following to one of our common header files:
```c
inline bool jToBool(jboolean value) {
    return value == JNI_TRUE;
}


inline jboolean boolToJ(bool value) {
    return value ? JNI_TRUE : JNI_FALSE;
}

```

csrc/env.h
alexw91: Done. Switched this exception handling to use `std::string` rather than plain C strings.

csrc/env.h
alexw91: This was a misunderstanding on my part of the Eclipse IDE `strncpy` documentation that states `This function is similar to strcpy but always copies exactly size characters into to.` which made it sound like if `sizeof(m_message_cstr) - 1` was passed in that strncpy might read past the end of the from buffer. Other documentation sources are more clear that this isn't the case.

src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -289,68 +335,122 @@ private static void assertVersionMatch() {
...
+        }
+        throw new AssertionError("Unable to read enough entropy");
+    }
+
+    /**
+     * Unfortunately, we cannot actually use Files.createTempFile, because that internally depends on
+     * SecureRandom, which results in a circular dependency.
+     */
+    private static synchronized Path createPrivateTmpDir(final String prefix) throws IOException {
+        final Path systemTempDir = Paths.get(System.getProperty("java.io.tmpdir"));
```
WillChilds-Klein, line 414: is this property always guaranteed to be present? if not, does `Files.isDirectory` handle `null` well or does it blow up w/ NPE?

alexw91: `Paths.get()` will return `java.nio.file.InvalidPathException` if `java.io.tmpdir` is not set. However this was the previously existing behavior before this PR, so I think it is acceptable.

WillChilds-Klein: ah, yep looking at tip-of-`develop`, i see that we'd already assumed the presence of that property. fair enough.

WillChilds-Klein (approved):

SalusaSecondus (approved):

PR #226, opened 2022-06-25
Add configuration security file for JDKs newer than 11

### LINKED ISSUE #175: TLS SSLContext not available when using JDK 15
Hi, I am using ACCP in OpenSearch 1.2.4 and get an error when configure `Djava.security.properties=/mypath/amazon-corretto-crypto-provider.security`

> Feb 11 02:09:50 ip-10-1-40-70 opensearch[1787]: [2022-02-11T02:09:50,196][ERROR][o.o.s.s.DefaultSecurityKeyStore] [ip-10-1-40-70] Unable to determine supported ciphers due to
Feb 11 02:09:50 ip-10-1-40-70 opensearch[1787]: java.security.NoSuchAlgorithmException: TLS SSLContext not available
Feb 11 02:09:50 ip-10-1-40-70 opensearch[1787]:         at sun.security.jca.GetInstance.getInstance(GetInstance.java:159) ~[?:?]
Feb 11 02:09:50 ip-10-1-40-70 opensearch[1787]:         at javax.net.ssl.SSLContext.getInstance(SSLContext.java:184) ~[?:?]
Feb 11 02:09:50 ip-10-1-40-70 opensearch[1787]:         at org.opensearch.security.ssl.DefaultSecurityKeyStore.initEnabledSSLCiphers(DefaultSecurityKeyStore.java:805) [opensearch-security-1.2.4.0.jar:1.2.4.0]

I traced back into DefaultSecurityKeyStore.java:805 in https://github.com/opensearch-project/security/blob/7ebb2ca045e6ff259c5c8084762eee5656820e33/src/main/java/org/opensearch/security/ssl/DefaultSecurityKeyStore.java#L805 and its just ` final SSLContext serverContext = SSLContext.getInstance("TLS");`

But when I switch back to JDK 11, it is working.

So is there any issue with JDK 15 and above or anything I need configure manually?

SalusaSecondus:
*Issue:* Fixes #175

*Description of changes:*
General fixes to existing `java.security` configuration file. Introduces a new version of JDKs newer than 11 (tested on 17). Updates tests to use the new file as well as introduces a regression test which detects if we break TLS in the future.

### DIFF
--- a/etc/amazon-corretto-crypto-provider-jdk15.security
+++ b/etc/amazon-corretto-crypto-provider-jdk15.security
@@ -1,0 +1,36 @@
+#
+# List of providers and their preference orders
+
+# Note that we can't just insert ours at the top - we need to restate all
+# providers that are being shifted down to a different index (which is to say,
+# all of them).
+#
+# To ensure this file works on all possible platforms we do include platform specific providers.
+
+security.provider.1=com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider
+security.provider.2=SUN
+security.provider.3=SunRsaSign
+security.provider.4=SunEC
+security.provider.5=SunJSSE
+security.provider.6=SunJCE
+security.provider.7=SunJGSS
+security.provider.8=SunSASL
+security.provider.9=XMLDSig
+security.provider.10=SunPCSC
+security.provider.11=JdkLDAP
+security.provider.12=JdkSASL
+security.provider.13=SunMSCAPI
+security.provider.14=Apple
+security.provider.15=SunPKCS11
+
+#
+# A list of known strong SecureRandom implementations.
+#
+# To help guide applications in selecting a suitable strong
+# java.security.SecureRandom implementation, Java distributions should
+# indicate a list of known strong implementations using the property.
+#
+# This is a comma-separated list of algorithm and/or algorithm:provider
+# entries.
+#
+securerandom.strongAlgorithms=NIST800-90A/AES-CTR-256:AmazonCorrettoCryptoProvider,NativePRNGBlocking:SUN,DRBG:SUN
--- a/etc/amazon-corretto-crypto-provider-jdk15.security
+++ b/etc/amazon-corretto-crypto-provider-jdk15.security
@@ -33,4 +33,4 @@ security.provider.15=SunPKCS11
 # This is a comma-separated list of algorithm and/or algorithm:provider
 # entries.
 #
-securerandom.strongAlgorithms=NIST800-90A/AES-CTR-256:AmazonCorrettoCryptoProvider,NativePRNGBlocking:SUN,DRBG:SUN
+securerandom.strongAlgorithms=DEFAULT:AmazonCorrettoCryptoProvider,NativePRNGBlocking:SUN,DRBG:SUN
--- a/tests/ci/run_accp_test_integration.sh
+++ b/tests/ci/run_accp_test_integration.sh
@@ -25,7 +25,7 @@ version=$($TEST_JAVA_HOME/bin/java -version 2>&1 | head -1 | cut -d'"' -f2 | sed
 # The JDK version should be least 10 for a regular ACCP build. We can
 # still test on older versions with the TEST_JAVA_HOME property.
 if (( "$version" <= "10" )); then
-	./gradlew -DTEST_JAVA_HOME=$TEST_JAVA_HOME -DFIPS=$testing_fips test_integration
+	./gradlew -DTEST_JAVA_HOME=$TEST_JAVA_HOME -DTEST_JAVA_MAJOR_VERSION=$version -DFIPS=$testing_fips test_integration
 	exit $?
 fi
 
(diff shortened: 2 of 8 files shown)


etc/amazon-corretto-crypto-provider-jdk15.security
```diff
@@ -0,0 +1,36 @@
...
+# A list of known strong SecureRandom implementations.
+#
+# To help guide applications in selecting a suitable strong
+# java.security.SecureRandom implementation, Java distributions should
+# indicate a list of known strong implementations using the property.
+#
+# This is a comma-separated list of algorithm and/or algorithm:provider
+# entries.
+#
+securerandom.strongAlgorithms=NIST800-90A/AES-CTR-256:AmazonCorrettoCryptoProvider,NativePRNGBlocking:SUN,DRBG:SUN
```
WillChilds-Klein, line 36: i see this is copied over from the other security properties file, but is it still accurate? i believe [we removed the `NIST800-90A/AES-CTR-256` service](https://github.com/corretto/amazon-corretto-crypto-provider/commit/e15bca8ad1cfd57d9a74239683ff1d7b86212a19#diff-36c87648dfaf4bdfe0f2238610f780a9dfef4ad8532944b522a19d7250643c84L82) a few months ago.

etc/amazon-corretto-crypto-provider-jdk15.security
```diff
@@ -0,0 +1,36 @@
...
+security.provider.6=SunJCE
+security.provider.7=SunJGSS
+security.provider.8=SunSASL
+security.provider.9=XMLDSig
+security.provider.10=SunPCSC
+security.provider.11=JdkLDAP
+security.provider.12=JdkSASL
+security.provider.13=SunMSCAPI
+security.provider.14=Apple
+security.provider.15=SunPKCS11
```
WillChilds-Klein, line 24: out of curiosity, what's this list based on? how are the priorities decided?

tests/ci/run_accp_test_integration.sh
```diff
@@ -25,7 +25,7 @@ version=$($TEST_JAVA_HOME/bin/java -version 2>&1 | head -1 | cut -d'"' -f2 | sed
 # The JDK version should be least 10 for a regular ACCP build. We can
 # still test on older versions with the TEST_JAVA_HOME property.
 if (( "$version" <= "10" )); then
-	./gradlew -DTEST_JAVA_HOME=$TEST_JAVA_HOME -DFIPS=$testing_fips test_integration
+	./gradlew -DTEST_JAVA_HOME=$TEST_JAVA_HOME -DTEST_JAVA_MAJOR_VERSION=$version -DFIPS=$testing_fips test_integration
```
WillChilds-Klein, line 28: why is this now necessary? doesn't -DTEST_JAVA_HOME tell which JAVA_HOME to use for running test, and thus which version to use?

etc/amazon-corretto-crypto-provider-jdk15.security
SalusaSecondus: Oops! Good point. I'll fix that now.

etc/amazon-corretto-crypto-provider-jdk15.security
SalusaSecondus: This is based on the current provider list in Corretto-17

tests/ci/run_accp_test_integration.sh
SalusaSecondus: We don't (currently) extract the major version information from the TEST_JAVA_HOME path and need that value to gate some logic in cmake. So, we need to pass it in explicitly. (Technically, it should already have been on this command-line, but previously we only cared if the version was sufficiently old, not if it was sufficiently new.

WillChilds-Klein: i see, that makes sense. thanks.

tst/com/amazon/corretto/crypto/provider/test/SecurityPropertyTester.java
```diff
@@ -26,11 +29,19 @@ public static void main(String[] args) throws Exception {
...
+
     // We know that Java has the SunEC provider which can generate EC keys.
     // We try to grab it to show that the nothing interfered with proper provider loading.
     @SuppressWarnings("unused")
     KeyPairGenerator kpg = KeyPairGenerator.getInstance("EC", "SunEC");
 
+    // Ensure we properly configured ourselves as "strong" instance of SecureRandom
+    // Applications should never use getInstanceStrong as it is an anti-pattern.
+    final SecureRandom strongRng = SecureRandom.getInstanceStrong();
+    assertEquals(NATIVE_PROVIDER.getName(), strongRng.getProvider().getName());
```
WillChilds-Klein, line 43: nit: if we're going to assert that ACCP is the provider here, maybe we should remove the 2 `SUN`  algs from `securerandom.strongAlgorithms`?

WillChilds-Klein (approved):

tst/com/amazon/corretto/crypto/provider/test/SecurityPropertyTester.java
SalusaSecondus: While we *know* in our tests that ACCP is on a supported system and so the fallbacks aren't needed, we want this configuration to work even when loaded on an unsupported system. Sure, ACCP won't work, but everything else should. So we leave the two SUN algorithms there as fallbacks for that case.

PR #220, opened 2022-06-16
Enforce write lock on MiscSingleThreadedTests

WillChilds-Klein:
*Issue #, if available:* n/a

*Description of changes:*

We've been seeing some intermittent CI failures (both internally and
externally). These failures manifest in a few ways: assertions that keys
translated by different providers aren't the same, buffer lengths
differing from expected size, etc. These failures are consistent with
unexpected providers performing intended operations. One way this can
happen is a race in our tests. Some of our tests install ACCP as a
JVM-global provider, while the rest of our tests assume that ACCP is not
installed. Because we run our JUnit tests in parallel within the same
JVM process, if one of the installing tests wins the race, subsequent
losing tests will have their uninstalled-ACCP assumptions violated.

Taking a look at which tests install ACCP:

```
$ ag -l '\.install\(\)' tst/
tst/com/amazon/corretto/crypto/provider/test/RecursiveInitializationTest.java
tst/com/amazon/corretto/crypto/provider/test/integration/TestCertificateGenerator.java
tst/com/amazon/corretto/crypto/provider/test/MiscSingleThreadedTests.java
tst/com/amazon/corretto/crypto/provider/test/SecureRandomBench.java
tst/com/amazon/corretto/crypto/provider/test/SecureRandomGenerator.java
tst/com/amazon/corretto/crypto/provider/test/SecurityPropertyRecursiveTester.java
tst/com/amazon/corretto/crypto/provider/test/TestProviderInstallation.java
tst/com/amazon/corretto/crypto/provider/test/integration/TestHTTPSServer.java
```

All installing tests other than `TestProviderInstallation.java` and
`MiscSingleThreadedTests.java` are run as standalone tests in a separate
process from our tests run by the JUnit test runner.
`TestProviderInstallation.java` already had a write lock installed, but
`MiscSingleThreadedTests.java` hadn't. So, we add a write lock to ensure
that it does not run concurrently (and thus potentially interfere) with
any other tests.

Finally, we add `@AfterClass` methods to installing JUnit tests to
ensure that ACCP is not installed if these tests are run before any
others.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #227, opened 2022-06-28
Skip tests that rely on JCE RSASSA-PSS for JDKs that don't support it

WillChilds-Klein:
*Issue #, if available:* n/a

*Description of changes:*

# Notes
## Skip tests that rely on JCE RSASSA-PSS for JDKs that don't support it

JDK10 doesn't seem to ship with RSASSA-PSS support by default, so skip
tests that rely on comparisons against JCE PSS implementations for JDK
versions that don't support them.

We also adjust EvpSignature.engineGetParameters to throw an
InvalidArgumentException (from the java.security package) when PSS
parameters are requested on a JDK platform that doesn't support them.

## Use copies of static member vars in EvpKeyFactoryTest

We've been seeing a number of intermittent test EvpKeyFactoryTest
failures in dryrun builds (such as [this one][1]) that appear to be due
to thread safety issues in the test runner. This commit returns _copies_
of static hashmap entries rather than shared references for
parameterized test parameters, so concurrent test case threads
accessing/using/potentially overwriting those values should be totally
independent. Confidence in this change was bolstered by two successful
dryruns:

- [AmazonCorrettoCryptoProvider/import][2]
- [AWSCryptoTools/native-jce-bindings][3]
  + NB: KMSCryptoCommon is failing in this dryrun for [known reasons][4]
    and can be ignored for the purposes of this commit.

[1]: https://build.amazon.com/6026895824
[2]: https://build.amazon.com/6026902091
[3]: https://build.amazon.com/6026903374
[4]: https://sim.amazon.com/issues/f81ed0df-6cc5-4c61-a385-d0f699f40576?selectedConversation=b126847d-41b6-4c5c-a9f8-ba28cb4479d7

# Testing
See CI checks for external validation. For internal validation, in a local brazil workspace I ran:

```
$ bb clean && bb release
...
BUILD SUCCEEDED
```

Dry runs against a few versionsets:

- [AmazonCorrettoCryptoProvider/import](https://build.amazon.com/6027001760)
- [AWSCryptoTools/njce-pre-sign](https://build.amazon.com/6027002969)
   + NB: KMSCryptoCommon is failing in this dryrun for [known reasons][4]
    and can be ignored for the purposes of this commit.
---

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
@@ -10,6 +10,7 @@ import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
 import java.security.InvalidParameterException;
 import java.security.Key;
+import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.security.SignatureException;
@@ -223,6 +224,10 @@ abstract class EvpSignatureBase extends SignatureSpi {
                 final AlgorithmParameters params = AlgorithmParameters.getInstance("RSASSA-PSS");
                 params.init(pssParams_);
                 return params;
+            } catch (final NoSuchAlgorithmException ex) {
+                // NOTE: this method can only throw unchecked exceptions, and UnsupportedOperationException
+                // seems like the most appropriate for JDK platforms that don't support RSASSA-PSS.
+                throw new UnsupportedOperationException("RSASSA-PSS unsupported.", ex);
             } catch (final GeneralSecurityException ex) {
                 throw new AssertionError(ex);
             }
--- a/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
@@ -17,6 +17,7 @@ import static org.junit.jupiter.api.Assumptions.assumeTrue;
 import java.util.List;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
+import java.security.AlgorithmParameters;
 import java.security.GeneralSecurityException;
 import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
@@ -419,13 +420,13 @@ public final class EvpSignatureSpecificTest {
 
         signature = Signature.getInstance("RSASSA-PSS", NATIVE_PROVIDER);
         signature.initSign(pair.getPrivate());
-        spec = signature.getParameters().getParameterSpec(PSSParameterSpec.class);
+        spec = getPssParams(signature);
         assertEquals("SHA-1", spec.getDigestAlgorithm());
         assertEquals("SHA-1", ((MGF1ParameterSpec) spec.getMGFParameters()).getDigestAlgorithm());
 
         signature = Signature.getInstance("RSASSA-PSS", NATIVE_PROVIDER);
         signature.initVerify(pair.getPublic());
-        spec = signature.getParameters().getParameterSpec(PSSParameterSpec.class);
+        spec = getPssParams(signature);
         assertEquals("SHA-1", spec.getDigestAlgorithm());
         assertEquals("SHA-1", ((MGF1ParameterSpec) spec.getMGFParameters()).getDigestAlgorithm());
 
@@ -433,7 +434,7 @@ public final class EvpSignatureSpecificTest {
         signature = Signature.getInstance("RSASSA-PSS", NATIVE_PROVIDER);
         signature.setParameter(null);
         signature.initSign(pair.getPrivate());
-        spec = signature.getParameters().getParameterSpec(PSSParameterSpec.class);
+        spec = getPssParams(signature);
         assertEquals("SHA-1", spec.getDigestAlgorithm());
         assertEquals("SHA-1", ((MGF1ParameterSpec) spec.getMGFParameters()).getDigestAlgorithm());
     }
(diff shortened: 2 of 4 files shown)


tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -468,10 +468,26 @@ void testRsaPSSTryUpdateParamDuringBuffer() throws Exception {
...
     }
 
+    // RSASSA-PSS not available on Java10, so skip the test if we can't get get a AlgorithmParameters
+    // object for it
+    private static PSSParameterSpec getPssParams(Signature signature) {
+        try {
+            return signature.getParameters().getParameterSpec(PSSParameterSpec.class);
+        } catch (Throwable e) {
+            // Unfortunately, we can't be too specific in our catch statement here
+            // because ACCP appears to sometimes sneakily throw a (checked)
```
SalusaSecondus, line 482: This sounds like a bug in ACCP that we should fix.

WillChilds-Klein: yeah, it seems a bit fishy. I'm guessing that somewhere along the call stack we're not properly catching a NoSuchAlgorithmException where we should be wrapping and re-throwing with [InvalidParameterSpecException](https://docs.oracle.com/javase/7/docs/api/java/security/spec/InvalidParameterSpecException.html) ([which `getParameterSpec` throws](https://docs.oracle.com/javase/7/docs/api/java/security/AlgorithmParameters.html#getParameterSpec(java.lang.Class))).

what's really throwing me for a loop is how a checked exception (i.e. NoSuchAlgorithmException)can be thrown by a method that doesn't declare it... NoSuchAlgorithmException and InvalidParameterSpecException both share an ancestor in GeneralSecurityException, but `getParameterSpec` explicitly `throws InvalidParameterSpecException`. i've never seen this before...

I've [cut a SIM to track](https://sim.amazon.com/issues/CryptoAlg-1174), but I don't think we should gate this PR on that issue.

src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
```diff
@@ -217,12 +219,17 @@ protected synchronized void engineSetParameter(final AlgorithmParameterSpec para
...
         if (paddingType_ == RSA_PKCS1_PSS_PADDING && pssParams_ != null) {
             try {
                 final AlgorithmParameters params = AlgorithmParameters.getInstance("RSASSA-PSS");
                 params.init(pssParams_);
                 return params;
+            } catch (final NoSuchAlgorithmException ex) {
+                // NOTE: this method can only throw unchecked exceptions, and java.security's
+                //       InvalidParameterException seems like the most appropriate for JDK platforms
+                //       that don't support RSASSA-PSS.
+                throw new InvalidParameterException(ex.getMessage());
```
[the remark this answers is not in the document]
WillChilds-Klein: yeah, i waffled on this a bit. my thinking was that this wasn't really an "operation" (i usually see this exception on IO operations or similar), and the "parameter" in question is implicitly the instance itself. 

but i readily admit that's a bit of a stretch. i'll update the PR to change exceptions and drop the "throws" clause from engineGetParameters.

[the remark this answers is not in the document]
SalusaSecondus: An "operation" in the context of this exception is any possible thing that a computer might want to do but cannot.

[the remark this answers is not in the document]
WillChilds-Klein: er, re-reading your comment and [the javadocs](https://docs.oracle.com/javase/8/docs/api/java/security/SignatureSpi.html#engineGetParameters--), it looks like the method should only declare `throws UnsupportedOperationException` if we're not overriding it:

> Throws:
> [UnsupportedOperationException](https://docs.oracle.com/javase/8/docs/api/java/lang/UnsupportedOperationException.html) - if this method is not overridden by a provider

I'm not 100% sure I'm reading this correctly, but it sounds like our override shouldn't declare `throws UnsupportedOperationException`, so we're good to go with your 2 suggested commits.

src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
```diff
@@ -223,6 +224,10 @@ protected synchronized AlgorithmParameters engineGetParameters() {
                 final AlgorithmParameters params = AlgorithmParameters.getInstance("RSASSA-PSS");
                 params.init(pssParams_);
                 return params;
+            } catch (final NoSuchAlgorithmException ex) {
+                // NOTE: this method can only throw unchecked exceptions, and UnsupportedOperationException
+                // seems like the most appropriate for JDK platforms that don't support RSASSA-PSS.
+                throw new UnsupportedOperationException(ex.getMessage());
```
SalusaSecondus, line 230: By doing this we lose the stack trace. Please be sure to include the caught exception as the cause in the thrown exception.
```suggestion
                throw new UnsupportedOperationException(ex);
```

WillChilds-Klein: Sorry, this was a hold-over from when I was using InvalidParameterException, which [doesn't have a constructor that takes a `Throwable`](https://docs.oracle.com/javase/7/docs/api/java/security/InvalidParameterException.html). I've adjusted this line to wrap the exception and add a brief message.

SalusaSecondus (approved):
Ship it!

WillChilds-Klein:
x86 CI failure due to a transient failure [that we've seen before](https://github.com/corretto/amazon-corretto-crypto-provider/pull/220#issuecomment-1157227833). i'll kick CI to retry.

Screen Shot 2022-06-28 at 1 13 28 PM

SalusaSecondus (approved):

PR #232, opened 2022-07-12
Support release pipeline

SalusaSecondus:
*Description of changes:*

These changes are the repo-related changes necessary for the proof-of-concept publishing pipeline to work.  The PoC has been tested up to (but not including) actual release.

Important notes/tasks:
* PoC pipeline is linked through my personal account using GitHubV2. Must be V2 which isn't yet enabled for the Corretto organization. (This is needed to do a full clone of the repo, necessary for submodules.)
* When making real, ensure that the role has proper permissions to access github connection: https://docs.aws.amazon.com/codepipeline/latest/userguide/troubleshooting.html#codebuild-role-connections
* All build-specs are based off the stock AL2 image rather than a custom Docker image (this should be fixed)
* Staging to maven has an unlisted dependency on fakePublish (which gets all the files prepped). So long as fakePublish is called first, publishToSonatype will work. I haven't figured out how to fix this yet.
* New parameter to build lets us override the cmake executable. This is to make it easier to install and use cmake3. Once we properly install it in the docker image, we may not need this override anymore
* Test improvements needed
  * Test code needs to run on more platforms than just AL2
  * Tests don't actually using the staging version of Maven except to call wget. This is lower fidelity than using Maven correct and would be good to fix. However, as ACCP is stand-alone with zero dependencies the risk of this is relatively low.
  * No dieharder tests are currently in the pipeline as that needs to be properly installed (and is slow) and thus is hard without custom docker images. (Dieharder is run prior to this pipeline, so this is low-risk)
  * I now pull detailed test reports out of the builds in CodeBuild. We might want to port that to our existing CI runs to make it easier to debug problems.
* Consider using CodeBuild caching to speed up build of awslc
* Any references to `STAGING_ID` where it is passed as an environment variable are old and I missed removing them
* Current S3 bucket used by pipeline will need to be replaced when moved to production (and link to it from approval fixed up)
* wget based actions will sometimes fail and need to be retried. This primarily happens in setup (and thus goes away in a docker world) but has been seen elsewhere. Might be worth building in automatic retry logic if this becomes a real issue
* Internal documentation on release process will need to be brought up to date
* Coordinate with ESDK team to modernize how we handle secrets

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -218,6 +218,7 @@ if (SIGNED_JAR)
         OUTPUT ${ACCP_JAR}
         COMMAND ${CMAKE_COMMAND} -E copy ${SIGNED_JAR} ${ACCP_JAR}
         COMMAND ${CMAKE_COMMAND} -E echo Copied ${SIGNED_JAR} to ${ACCP_JAR}
+        COMMAND ${CMAKE_COMMAND} -E copy ${OPENSSL_CRYPTO_LIBRARY} $<TARGET_FILE_DIR:amazonCorrettoCryptoProvider>
     )
 else()
 # CMake's UseJar doesn't let us control the paths that resource files are
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,6 +217,7 @@ if (SIGNED_JAR)
     add_custom_command(
         OUTPUT ${ACCP_JAR}
         COMMAND ${CMAKE_COMMAND} -E copy ${SIGNED_JAR} ${ACCP_JAR}
+        COMMAND ${CMAKE_COMMAND} -E echo Copied ${SIGNED_JAR} to ${ACCP_JAR}
     )
 else()
 # CMake's UseJar doesn't let us control the paths that resource files are
--- a/build.gradle
+++ b/build.gradle
@@ -5,14 +5,36 @@ import groovy.xml.*
 
 plugins {
     id 'maven-publish'
+    id "io.github.gradle-nexus.publish-plugin" version "1.1.0"
     id 'signing'
+    id "com.google.osdetector" version "1.7.0"
 }
 
 group = 'software.amazon.cryptools'
 version = '2.0.0'
 ext.isFips = Boolean.getBoolean('FIPS')
 
-def awslcSrcPath = "aws-lc/"
+def awslcSrcPath = "${projectDir}/aws-lc/"
+def cmakeBin = System.properties['cmakeBin'] != null ? System.properties['cmakeBin'] : "cmake";
+
+ext.isJceSigned = { pathToJar ->
+    def stdout = new ByteArrayOutputStream()
+    exec {
+        executable "jarsigner"
+        args "-verify", pathToJar
+        standardOutput = stdout;
+    }
+    return stdout.toString().contains("jar verified")
+}
+
+ext.assertJceSigned = { pathToJar ->
+    if (!isJceSigned(pathToJar)) {
+        ant.fail("${pathToJar} is not signed")
+    } else {
+        println "${pathToJar} is signed"
+    }
+    return pathToJar
+}
 
 configurations {
     jacocoAgent
@@ -69,12 +91,15 @@ task buildAwsLc {
         if (file(awslcSrcPath).list().size() == 0) {
             throw new GradleException("aws-lc dir empty! run 'git submodule update --init --recursive' to populate.")
         }
+        mkdir "${buildDir}/awslc"
+        mkdir sharedObjectOutDir
+        mkdir cMakeBuildDir
     }
 
     doLast {
         exec {
             workingDir awslcSrcPath
-            executable 'cmake'
+            executable cmakeBin
             args "-B${cMakeBuildDir}"
             args '-DBUILD_SHARED_LIBS=ON'
             args '-DCMAKE_BUILD_TYPE=RelWithDebInfo'
@@ -84,6 +109,7 @@ task buildAwsLc {
             if (isFips) {
                 args '-DFIPS=1'
             }
+            args '.'
         }
         exec {
             workingDir awslcSrcPath
@@ -108,18 +134,39 @@ task executeCmake(type: Exec) {
 
     workingDir "${buildDir}/cmake"
     def prebuiltJar = null
-    if (project.hasProperty('stagingUrl')) {
+    def projectName = isFips ? project.name + "-FIPS" : project.name
+    def prebuiltJarFileName = "${projectName}-${version}-${osdetector.classifier}.jar"
+    if (System.properties['stagingProperties'] != null) {
+        def stagingProperties = new Properties()
+        if (isFips) {
+            stagingProperties.load(
+                new java.io.FileInputStream("${System.env.CODEBUILD_SRC_DIR_Stage_FIPS}/fake_maven/staging.properties"));
+        } else {
+            stagingProperties.load(
+                new java.io.FileInputStream("${System.env.CODEBUILD_SRC_DIR_Stage}/fake_maven/staging.properties"));
+        }
+        println "Loaded staging id is " + stagingProperties['staging.id']
+        def repoBase = "https://aws.oss.sonatype.org/content/repositories/" + stagingProperties['staging.id']
         mkdir "${buildDir}/tmp"
         exec {
             workingDir "${buildDir}/tmp"
-            commandLine 'wget', "${stagingUrl}/software/amazon/cryptools/${project.name}/${stagingVersion}/${project.name}-${stagingVersion}-linux-x86_64.jar"
+            commandLine 'wget', "${repoBase}/software/amazon/cryptools/${projectName}/${version}/${prebuiltJarFileName}"
         }
-        prebuiltJar = "${buildDir}/tmp/${project.name}-${stagingVersion}-linux-x86_64.jar"
+        prebuiltJar = assertJceSigned("${buildDir}/tmp/${prebuiltJarFileName}")
     } else if (System.properties['prebuiltJar'] != null) {
         prebuiltJar = "${projectDir}/" + System.properties['prebuiltJar']
+    } else if (System.properties['fakeMaven'] != null) {
+        def fakeMavenBase
+        if (isFips) {
+            fakeMavenBase = System.env.CODEBUILD_SRC_DIR_Stage_FIPS
+        } else {
+            fakeMavenBase = System.env.CODEBUILD_SRC_DIR_Stage
+        }
+        fakeMavenBase = "${fakeMavenBase}/fake_maven"
+        prebuiltJar = "${fakeMavenBase}/${prebuiltJarFileName}"
     }
 
-    executable 'cmake'
+    executable cmakeBin
     args "-DTEST_CLASSPATH=${configurations.testDep.asPath}", "-DJACOCO_AGENT_JAR=${configurations.jacocoAgent.singleFile}"
     args "-DOPENSSL_ROOT_DIR=${buildDir}/awslc/bin", '-DCMAKE_BUILD_TYPE=Release', '-DPROVIDER_VERSION_STRING=' + version
     args "-DTEST_RUNNER_JAR=${configurations.testRunner.singleFile}"
@@ -131,6 +178,7 @@ task executeCmake(type: Exec) {
 
     if (prebuiltJar != null) {
        args '-DSIGNED_JAR=' + prebuiltJar
+       print "Using SIGNED_JAR=${prebuiltJar}"
     }
 
     if (System.properties['TEST_JAVA_HOME'] != null) {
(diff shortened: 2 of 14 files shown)


CMakeLists.txt
```diff
@@ -217,6 +217,8 @@ if (SIGNED_JAR)
     add_custom_command(
         OUTPUT ${ACCP_JAR}
         COMMAND ${CMAKE_COMMAND} -E copy ${SIGNED_JAR} ${ACCP_JAR}
+        COMMAND ${CMAKE_COMMAND} -E echo Copied ${SIGNED_JAR} to ${ACCP_JAR}
+        COMMAND ${CMAKE_COMMAND} -E copy ${OPENSSL_CRYPTO_LIBRARY} $<TARGET_FILE_DIR:amazonCorrettoCryptoProvider>
```
WillChilds-Klein, line 221: why do we need to copy aws-lc over? don't we bundle it in the ACCP JAR?

build.gradle
```diff
@@ -5,14 +5,36 @@ import groovy.xml.*
...
 
-def awslcSrcPath = "aws-lc/"
+def awslcSrcPath = "${projectDir}/aws-lc/"
+def cmakeBin = System.properties['cmakeBin'] != null ? System.properties['cmakeBin'] : "cmake";
+
+ext.isJceSigned = { pathToJar ->
+    def stdout = new ByteArrayOutputStream()
+    exec {
+        executable "jarsigner"
+        args "-verify", pathToJar
```
WillChilds-Klein, line 24: what public key is this verifying a signature against?

build.gradle
```diff
@@ -108,18 +134,39 @@ task executeCmake(type: Exec) {
...
+    def projectName = isFips ? project.name + "-FIPS" : project.name
+    def prebuiltJarFileName = "${projectName}-${version}-${osdetector.classifier}.jar"
+    if (System.properties['stagingProperties'] != null) {
+        def stagingProperties = new Properties()
+        if (isFips) {
+            stagingProperties.load(
+                new java.io.FileInputStream("${System.env.CODEBUILD_SRC_DIR_Stage_FIPS}/fake_maven/staging.properties"));
+        } else {
+            stagingProperties.load(
+                new java.io.FileInputStream("${System.env.CODEBUILD_SRC_DIR_Stage}/fake_maven/staging.properties"));
```
WillChilds-Klein, line 146: why are staging properties different between FIPS/non-FIPS?

build.gradle
```diff
@@ -108,18 +134,39 @@ task executeCmake(type: Exec) {
 
     workingDir "${buildDir}/cmake"
     def prebuiltJar = null
-    if (project.hasProperty('stagingUrl')) {
+    def projectName = isFips ? project.name + "-FIPS" : project.name
+    def prebuiltJarFileName = "${projectName}-${version}-${osdetector.classifier}.jar"
```
WillChilds-Klein, line 138: note for other reviewers -- from osdetector [plugin README](https://github.com/google/osdetector-gradle-plugin#usage):

> `osdetector.classifier`: classifier, which is `osdetector.os + '-' + osdetector.arch`, e.g., `linux-x86_64

build.gradle
```diff
@@ -377,19 +453,96 @@ task overkill {
...
+            rename 'AmazonCorrettoCryptoProvider.jar', "${newPrefix}-linux-aarch_64.jar"
+            into "${destinationDir}"
+        }
+        copy {
+            from x64Base
+            rename { String filename ->
+                if (filename.equals("AmazonCorrettoCryptoProvider.jar")) {
+                    return "${newPrefix}-linux-x86_64.jar"
+                } else {
+                    return filename.replace("AmazonCorrettoCryptoProvider-", "${newPrefix}-")
```
WillChilds-Klein, line 496: when would we hit this `else` case?

why is this treated differently from aarch64?

build.gradle
```diff
@@ -377,19 +453,96 @@ task overkill {
...
+            }
+            into "${destinationDir}"
+        }
+        if (project.hasProperty('jcecertAlias')) {
+            ant.signjar(
+                alias: jcecertAlias,
+                jar: "${destinationDir}/${newPrefix}-linux-aarch_64.jar",
+                storepass: jcecertPassword,
+                keystore: "${project.gradle.gradleUserHomeDir}/${jcecertJks}",
+                tsaurl: "http://timestamp.digicert.com"
```
WillChilds-Klein, line 507: https?

build.gradle
```diff
@@ -413,20 +566,50 @@ if (project.hasProperty('mavenUser') && project.hasProperty('jcecertAlias')) {
...
     }
 
     signing {
         sign publishing.publications.mavenJava
     }
+
+    closeSonatypeStagingRepository {
+        doLast {
+            def idAsString = stagingRepositoryId.get()
+            def fw = new java.io.FileWriter("${buildDir}/fake_maven/staging.properties")
```
WillChilds-Klein, line 588: do we need to write to both `CODEBUILD_SRC_DIR_Stage`'s build dir and `CODEBUILD_SRC_DIR_Stage_FIPS`'s build dir?

tests/ci/codebuild/release/al2_aarch64_test.yml
```diff
@@ -0,0 +1,62 @@
...
+phases:
+  # TODO: Everything in the install section should be moved into a prebuilt image
+  install:
+    runtime-versions:
+      java: corretto11
+    commands:
+      - yum -y update
+      - yum -y group install 'Development Tools'
+      # - curl -L -O https://github.com/Kitware/CMake/releases/download/v3.24.0-rc2/cmake-3.24.0-rc2.tar.gz
+      - curl -L -O https://downloads.sourceforge.net/ltp/lcov-1.14-1.noarch.rpm
```
WillChilds-Klein, line 18: do we do anything with this RPM after it's downloaded?

also, minor nit: delete commented lines

tests/ci/codebuild/release/al2_aarch64_test.yml
```diff
@@ -0,0 +1,62 @@
...
+  discard-paths: no
+  base-directory: 'build'
+  name: $(date --iso-8601=s)
+reports:
+  unit-tests:
+    files:
+      - 'build/reports/unit-tests/**'
+    #base-directory: location
+    discard-paths: yes
+    file-format: JunitXml
```
WillChilds-Klein, line 56: nice! where can we view these reports?

tests/ci/codebuild/release/linux_aarch64_build.yml
```diff
@@ -0,0 +1,54 @@
...
+phases:
+  # TODO: Everything in the install section should be moved into a prebuilt image
+  install:
+    runtime-versions:
+      java: corretto11
+    commands:
+      - yum -y update
+      - yum -y group install 'Development Tools'
+      - curl -L -O https://downloads.sourceforge.net/ltp/lcov-1.14-1.noarch.rpm
+      - yum -y install cmake3; alias cmake=cmake3
```
WillChilds-Klein, line 18: do you need to alias this if you're specifying the cmakeBin override property?

tests/ci/codebuild/release/release.yml
```diff
@@ -0,0 +1,38 @@
...
+      - export GOROOT=$(pwd)/go_root
+      - export GO111MODULE=on
+      - export PATH=$GOROOT/bin:$PATH
+      # TODO: Can we move the recursive pull logic upstream?
+      - git submodule update --init --recursive
+      - pwd
+      - ls
+      # TODO: Uncomment next line to actually release
+      # - ./gradlew -DcmakeBin=cmake3 -DFIPS=${FIPS} -DstagingProperties=true releaseSonatypeStagingRepository
+      - ./gradlew -DcmakeBin=cmake3 -DFIPS=${FIPS} -DstagingProperties=true tasks
```
WillChilds-Klein, line 38: nit: could you please add a comment explaining the purpose of this no-op `tasks` invocation?

tests/ci/codebuild/release/stage.yml
```diff
@@ -0,0 +1,42 @@
...
+    commands:
+  build:
+    commands:
+      # TODO: Sign jars here
+      # TODO: Uncomment the following lines once configured
+      - pip install awscli aws-encryption-sdk-cli
+      # Debugging only
+      - aws sts get-caller-identity
+      # Retrieve encrypted configuration from S3
+      - aws s3 cp s3://${CONFIG_PREFIX}/${JKS_FILENAME}.enc $HOME/${JKS_FILENAME}.enc
```
WillChilds-Klein, line 24: who sets `CONFIG_PREFIX` and the other environment variables?

CMakeLists.txt
SalusaSecondus: This is for the test when we load the .so files from an external source (not the JAR), so libcrypto must be in the same directory as libaccp.

build.gradle
SalusaSecondus: Because FIPS and non-FIPS have different repos and staging Ids. It seems impractical to combine them.

build.gradle
SalusaSecondus: We pull all generic (non-platform specific) JARs from the x64 build. Why? They have to come from somewhere, so why not? The else branch is hit for the generic JARs (such as source and javadoc).

aarch64 only pulls the actual ACCP jar.

build.gradle
SalusaSecondus: Not relevant here as security is ensured by the protocol. I don't even think that `https` is supported.

build.gradle
SalusaSecondus: Not explicitly. CodePipelines does that for us.

tests/ci/codebuild/release/al2_aarch64_test.yml
SalusaSecondus: This should be installed. I'll readd that.

tests/ci/codebuild/release/linux_aarch64_build.yml
SalusaSecondus: Forgot to remove it here. Shall do.

tests/ci/codebuild/release/release.yml
SalusaSecondus: Sure.

tests/ci/codebuild/release/stage.yml
SalusaSecondus: CodePipeine/CodeBuild

tests/ci/codebuild/release/al2_aarch64_test.yml
WillChilds-Klein: whoah! i didn't know codebuild understood junit XML. this is pretty neat.

Screen Shot 2022-07-13 at 2 39 39 PM

build.gradle
WillChilds-Klein: sure, makes sense.

tests/ci/codebuild/release/al2_aarch64_test.yml
SalusaSecondus: This is something we should probably port to our standard CI jobs. Once this change is merged in all you need to do is update the build-specs. If you want, I can do that now too.

tests/ci/codebuild/run_accp_overkill.yml
```diff
@@ -7,3 +7,14 @@ phases:
   build:
     commands:
       - ./tests/ci/run_accp_overkill.sh
+reports:
+  unit-tests:
+    files:
+      - 'build/reports/unit-tests/**'
```
WillChilds-Klein, line 13: overkill/dieharder tests generate JUnit reports?

tests/ci/codebuild/release/al2_aarch64_test.yml
```diff
@@ -0,0 +1,54 @@
...
+      - mv go ../go_root
+      - cd ..
+  build:
+    commands:
+      # TODO: Move this to the docker image also
+      - export GOROOT=$(pwd)/go_root
+      - export GO111MODULE=on
+      - export PATH=$GOROOT/bin:$PATH
+      # TODO: Can we move the recursive pull logic upstream?
+      - git submodule update --init --recursive
```
WillChilds-Klein, line 32: if we're pulling the JAR from sonatype staging (see `-DstagingProperties=true` below, L153 of build.gradle), why do we need to initialize the AWS-LC submodule? won't the compiled precompiled `.so` be bundled with the JAR?

isn't the goal of this buildspec to _not build anything_, instead pulling the JAR from staging and testing against that?

WillChilds-Klein: got it. makes sense that we don't bundle test bytecode in our JAR and thus need to build it here.

WillChilds-Klein (dismissed):

tests/ci/codebuild/run_accp_overkill.yml
SalusaSecondus: Looks like I'm wrong on this and will remove it. We should figure out why at some point.

WillChilds-Klein (approved):

SalusaSecondus:
Overkill is failing due to a sporadic entropy failure. I didn't touch anything there so I'm comfortable merging.

PR #235, opened 2022-07-14
Implement AES KWP

WillChilds-Klein:
*Issue #, if available:* CryptoAlg-1098

*Description of changes:*

This commit adds a new CipherSpi implementation, AesKeyWrapSpi. This
cipher can be instantiated as AES with mode KWP, and NoPadding. AWS-LC
only provides one-shot functions for key wrapping and unwrapping, so we
buffer any incremental data fed to the Cipher until finalization, at
which point we pass all buffered data over the JNI and call AWS-LC. AES
KWP is defined in [RFC-5649][1], which we rely on for output buffer
sizing and other implementation-adjacent details/input validation.

While JCE and BouncyCastle support callers specifying custom IVs, AWS-LC
does not so neither do we. We do not throw when a caller tries to
set IV's via `setParameters` or similar, we instead ignore the call and
return `null` on calls to `getParameters` and `getIv`.

As described in [this document][2], NIST provides Known Answer Tests for
AES KWP. We pull down [these test vectors][3] and include `gzip`'s of
them in the repo. We then add AesKwpKatTest to read these test cases in,
and execute them against ACCP. This commit follows prior art for KAT
tests in AesGcmKatTest. We also add many parameterized and input
validation tests, achieving 87% line coverage and 80% branch coverage.

Finally, we fix up CMakeLists.txt to include the AES GCM KAT tests in
its test builds/runs, which had previously been dropped from the test
set.

[1]: https://datatracker.ietf.org/doc/html/rfc5649
[2]: https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Algorithm-Validation-Program/documents/mac/KWVS.pdf
[3]: https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Algorithm-Validation-Program/documents/mac/kwtestvectors.zip

---


csrc/aes_kwp.cpp
```diff
@@ -0,0 +1,96 @@
...
+        key.get_bytes(env, keybuf.buf, 0, key.len());
+        if (AES_set_encrypt_key(keybuf.buf, key.len()*8, &aes_key) != 0) {
+            throw java_ex::from_openssl(EX_RUNTIME_CRYPTO, "AES key init failed");
+        }
+
+        // TODO [childw] smaller more precise size
+        // TODO [childw] try to eliminate extra copies.
+        SecureBuffer<uint8_t, 4096> inbuf;
+        input.get_bytes(env, inbuf.buf, 0, input.len());
+        SecureBuffer<uint8_t, 4096> outbuf;
```
WillChilds-Klein, line 44: is there an opportunity to use jni_borrow here to make AWS-LC write directly to java's heap?

CMakeLists.txt
```diff
@@ -554,7 +556,9 @@ add_jar(
     tests-code-jar
     SOURCES
         tst/com/amazon/corretto/crypto/provider/test/AccessibleByteArrayOutputStreamTest.java
+        tst/com/amazon/corretto/crypto/provider/test/AesGcmKatTest.java
```
WillChilds-Klein, line 559: TODO [childw]: do full comparison of available tests against this list in CMakeLists.txt, ideally we'll just list the relevant directories to build this list automatically.

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    @Override
+    protected Key engineUnwrap(final byte[] wrappedKey, final String wrappedKeyAlgorithm, final int wrappedKeyType)
+            throws InvalidKeyException, NoSuchAlgorithmException {
+        if (opmode != Cipher.UNWRAP_MODE || keyBytes == null) {
+            throw new IllegalStateException("Cipher must be init'd in WRAP_MODE");
+        }
+        try {
+            final byte[] unwrappedKey = new byte[wrappedKey.length];
```
SalusaSecondus, line 197: You can estimate slightly tighter than this because you know that there is always at least 8 bytes of overhead (IIRC), so let's allocate the smaller amount.

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
```diff
@@ -0,0 +1,227 @@
...
+            int unwrappedKeyLen = unwrapKey(keyBytes, wrappedKey, unwrappedKey);
+            final byte[] unwrappedKeyTruncated = new byte[unwrappedKeyLen];
+            System.arraycopy(unwrappedKey, 0, unwrappedKeyTruncated, 0, unwrappedKeyLen);
```
SalusaSecondus, line 200: As before, don't truncate if we happened to get the correct length. Also, do you want to zero the array that copied and not further used?

(Do you also want to zeroize the used array after the new key is constructed?)

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
```diff
@@ -0,0 +1,227 @@
...
+            return Utils.buildUnwrappedKey(provider, unwrappedKeyTruncated, wrappedKeyAlgorithm, wrappedKeyType);
+        // TODO [childw] make this exception more specific, perhaps BadPaddingException?
+        } catch (final Exception ex) {
+            throw new InvalidKeyException("Unwrapping failed", ex);
```
SalusaSecondus, line 204: `InvalidKeyException` [looks correct](https://github.com/corretto/corretto-17/blob/4922f0805033d2f4a872add164f05320ef1592d3/src/java.base/share/classes/com/sun/crypto/provider/KeyWrapCipher.java#L768-L769) to me.

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
```diff
@@ -0,0 +1,227 @@
...
+        }
+    }
+
+    @Override
+    protected byte[] engineUpdate(byte[] in, int inOffset, int inLen) {
```
SalusaSecondus, line 209: FYI, it looks like the [JCE supports ENCRYPT and DECRYPT](https://github.com/corretto/corretto-17/blob/4922f0805033d2f4a872add164f05320ef1592d3/src/java.base/share/classes/com/sun/crypto/provider/KeyWrapCipher.java) as well as update/doFinal. We need to figure out our compatibility story. You might need to support those also. It should be relatively simple to do so (using AccessibleByteArrayOutputStream).

csrc/aes_kwp.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+    try {
+        raii_env env(pEnv);
+
+        java_buffer key = java_buffer::from_array(env, keyArray);
+        java_buffer input = java_buffer::from_array(env, inputArray);
+        java_buffer output = java_buffer::from_array(env, outputArray);
+
+        AES_KEY aes_key;
+        SecureBuffer<uint8_t, 32> keybuf;
```
SalusaSecondus, line 34: `32` is a magic value. Maybe `#define MAX_AES_SIZE_BYTES 32` earlier?

csrc/aes_kwp.cpp
```diff
@@ -0,0 +1,96 @@
...
+        key.get_bytes(env, keybuf.buf, 0, key.len());
```
SalusaSecondus, line 35: You need a check that key.len() <= 32

csrc/aes_kwp.cpp
```diff
@@ -0,0 +1,96 @@
...
+        if (AES_set_encrypt_key(keybuf.buf, key.len()*8, &aes_key) != 0) {
```
SalusaSecondus, line 36: FYI, this is an expensive call and you might want to save the result if the key is commonly reused. You can see that we do something like that for AES-GCM.

csrc/aes_kwp.cpp
```diff
@@ -0,0 +1,96 @@
...
+            throw java_ex::from_openssl(EX_RUNTIME_CRYPTO, "AES key init failed");
```
SalusaSecondus, line 37: Generally we do the following instead (which is equivalent). It's up to you:
```suggestion
            throw_openssl(EX_RUNTIME_CRYPTO, "AES key init failed");
```

csrc/aes_kwp.cpp
SalusaSecondus, line 44: Rather than this just start by using `jni_borrow` and worry about optimizing it later if need be. These can be arbitrarily big so you cannot allocate a constant sized piece for it.

csrc/aes_kwp.cpp
```diff
@@ -0,0 +1,96 @@
...
+        output.put_bytes(env, outbuf.buf, 0, outlen);
+
+        return outlen;
+    } catch (java_ex &ex) {
+        ex.throw_to_java(pEnv);
+        return 0;
+    }
+}
+
+JNIEXPORT int JNICALL Java_com_amazon_corretto_crypto_provider_AesKeyWrapSpi_unwrapKey(
```
SalusaSecondus, line 58: All comments on wrap apply equally here.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.jupiter.api.parallel.ResourceAccessMode;
+import org.junit.jupiter.api.parallel.ResourceLock;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+@ExtendWith(TestResultLogger.class)
+@Execution(ExecutionMode.CONCURRENT)
+@ResourceLock(value = TestUtil.RESOURCE_GLOBAL, mode = ResourceAccessMode.READ)
+public final class AesKeyWrapTest {
```
SalusaSecondus, line 45: Your next critical tests are:
* Keys of lengths: {3, 4, 7, 8, 9, 15, 16, 17, 1023, 1024, 1025}
* Compatibility with results from the stock Java implementation (or BC)

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
```diff
@@ -0,0 +1,227 @@
...
+            }
+        }
+        engineInit(opmode, key, paramSpec, random);
+    }
+
+    @Override
+    protected void engineInit(int opmode, Key key, AlgorithmParameterSpec params, SecureRandom random)
+            throws InvalidKeyException, InvalidAlgorithmParameterException {
+        if (params != null && !(params instanceof IvParameterSpec)
+                && !Arrays.equals(ICV2, ((IvParameterSpec) params).getIV())) {
```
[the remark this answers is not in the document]
WillChilds-Klein: sure, can do.

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    @Override
+    protected byte[] engineWrap(final Key key) throws IllegalBlockSizeException, InvalidKeyException {
+        if (opmode != Cipher.WRAP_MODE || keyBytes == null) {
+            throw new IllegalStateException("Cipher must be init'd in WRAP_MODE");
+        }
+        try {
+            final byte[] encoded = Utils.encodeForWrapping(provider, key);
+            final byte[] wrappedKey = new byte[4096];
```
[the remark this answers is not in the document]
WillChilds-Klein: yep, you're right. it's laid out [in RFC](https://datatracker.ietf.org/doc/html/rfc5649#section-4.1) pretty clearly.

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
WillChilds-Klein: will do. unfortunately, i don't think we can guess the _precise_ size of the unwrapped key because of encrypted padding, [but we can get within 8 bytes](https://datatracker.ietf.org/doc/html/rfc5649#section-4.2). makes sense to minimize memory overhead.

csrc/aes_kwp.cpp
WillChilds-Klein: yep, good call.

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
WillChilds-Klein: yeah, i was hoping we could get away without buffering stuff across calls and managing that state/reset logic. if we need to support those Cipher modes as well, i guess we can't get around that. i'll start adding the buffering logic.

DIFFERENCES.md
```diff
@@ -70,6 +70,10 @@ This is equivalent to using a [GCMParameterSpec](https://docs.oracle.com/javase/
 By supporting the same ParameterSpec as other ciphers (such as `AES/CBC/PKCS5Padding`, which should not be used as it is no longer secure), ACCP makes it easier to migrate to the secure choice of `AES/GCM/NoPadding`.
 (This behavior is identical to how [BouncyCastle](https://bouncycastle.org/java.html) treats `IvParameterSpec` when used with AES-GCM.)
 
+## AES-KWP restricted support for IvParameterSpec
```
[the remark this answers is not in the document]
WillChilds-Klein: added.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -59,6 +59,9 @@ private void buildServiceMap() {
         addService("Cipher", "AES_128/GCM/NoPadding", "AesGcmSpi");
         addService("Cipher", "AES_256/GCM/NoPadding", "AesGcmSpi");
 
+        addService("Cipher", "AESWrapPad", "AesKeyWrapSpi");
+        addService("Cipher", "AES/KWP/NoPadding", "AesKeyWrapSpi");
```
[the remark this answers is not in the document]
WillChilds-Klein: nice. fixed.

csrc/aes_kwp.cpp
WillChilds-Klein: done.

CMakeLists.txt
WillChilds-Klein: ok, we're all set here. the only tests not included in this list are the benchmarks:

```
$ find tst -name *.java  | while read file; do grep $file CMakeLists.txt &>/dev/null || echo "NO ENTRY FOR $file"; done
NO ENTRY FOR tst/com/amazon/corretto/crypto/provider/test/SmallHashBench.java
NO ENTRY FOR tst/com/amazon/corretto/crypto/provider/test/SecureRandomBench.java
NO ENTRY FOR tst/com/amazon/corretto/crypto/provider/test/HMACBench.java
```

csrc/aes_kwp.cpp
WillChilds-Klein: that's good to know. i wouldn't have expected this to be an expensive call (it's just DER-decoding a 16-32 byte key?). to keep the initial implementation simple, i think it would be advisable to punt on this now, and revisit it if it becomes a hot spot.

if we were to retain the AES_KEY across wrap/unwrap invocations, we'd track its pointer as a NativeResource, right?

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
```diff
@@ -0,0 +1,242 @@
...
+            // TODO [childw] explanatory comment
+            if (unwrappedKeyLen != unwrappedKey.length) {
+                final byte[] tmp = new byte[unwrappedKeyLen];
+                System.arraycopy(unwrappedKey, 0, tmp, 0, Math.min(unwrappedKey.length, unwrappedKeyLen));
+                if (unwrappedKeyLen > unwrappedKey.length) {
+                    System.arraycopy(extra, 0, tmp, unwrappedKey.length, unwrappedKeyLen - unwrappedKey.length);
+                }
+                Arrays.fill(unwrappedKey, (byte) 0);
+                unwrappedKey = tmp;
+            }
```
SalusaSecondus, line 213: Simplify and remove use of `extra`

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    public static List<Arguments> sizes() {
+        final int[] keySizes = {16, 24, 32};
+        final int[] secretSizes = {
+            8,                      // https://datatracker.ietf.org/doc/html/rfc5649#section-4.1
+            16, 24, 32,             // AES keys
+            512, 1024, 2048, 4096,  // RSA keys
+            123, 900, 81, 99, 37,   // weird sizes to exercise padding logic
```
SalusaSecondus, line 56: Please add the other weird cases I provided. Having a case smaller than 8 and between 8 and 16 is important. In general my cases are intentionally right on the edges of many common errors.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+        byte[] secretBytes = new byte[secretSize];
+        sr.nextBytes(secretBytes);
+        final SecretKey secret = new SecretKeySpec(secretBytes, "AES");
+
+        Cipher c = Cipher.getInstance("AES/KWP/NoPadding", TestUtil.NATIVE_PROVIDER);
+        c.init(Cipher.WRAP_MODE, key, sr);
+        byte[] wrapped = c.wrap(secret);
+        assertFalse(Arrays.equals(secretBytes, wrapped));
+        c = Cipher.getInstance("AES/KWP/NoPadding", TestUtil.NATIVE_PROVIDER);
```
SalusaSecondus, line 83: Why did you get a new instance? Why not just reuse the existing one?

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
SalusaSecondus: Those compatibility tests really need to come now. They are the most important ones you'll add.

WillChilds-Klein: got it. working on 'em.

> Keys of lengths: {3, 4, 7, 8, 9, 15, 16, 17, 1023, 1024, 1025}

should we just be testing these as AES keys? or also as RSA, EC, etc.?

csrc/aes_kwp.cpp
SalusaSecondus: AES keys are not DER encoded. However, the AES keyschedule is very expensive. The [Stick Figure Guide to the Advanced Encryption Standard (AES)](http://www.moserware.com/2009/09/stick-figure-guide-to-advanced.html) is a great read. (They call it "key expansion".)

Yes, it would be tracked as a NativeResource.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
SalusaSecondus: They aren't valid keys of any lengths. So, use type "Generic" and have them be `SecretKey`s.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,94 @@
...
+        c.init(Cipher.UNWRAP_MODE, key, sr);
+        Key unwrapped = c.unwrap(wrapped, "AES", Cipher.SECRET_KEY);
```
SalusaSecondus, line 85: With the varying lengths, many of these are not "AES" keys, so just make this "Generic" instead.
```suggestion
        Key unwrapped = c.unwrap(wrapped, "Generic", Cipher.SECRET_KEY);
```

csrc/aes_kwp.cpp
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        jni_borrow inbuf(env, input, "input");
+        input.get_bytes(env, inbuf.data(), 0, input.len());
+        jni_borrow outbuf(env, output, "output");
+        size_t outlen;
+        if (!AES_wrap_key_padded(&aes_key, outbuf.data(), &outlen, outbuf.len(), inbuf.data(), input.len())) {
+            throw_openssl(EX_RUNTIME_CRYPTO, "Error wrapping key");
+        }
+        output.put_bytes(env, outbuf.data(), 0, outlen);
```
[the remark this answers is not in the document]
WillChilds-Klein: oh, awesome. i'll eliminate those copies here and in the unwrap method.

csrc/buffer.h
```diff
@@ -404,6 +404,9 @@ class jni_borrow {
 
 inline void java_buffer::get_bytes(raii_env &env, uint8_t *dest, size_t offset, size_t len) const {
     check_bounds(offset, len);
+    if (len == 0) {
+        return;
+    }
```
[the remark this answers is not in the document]
WillChilds-Klein: sure, would save a couple ticks by skipping the bounds checking.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
WillChilds-Klein: ah, that was vestigial from some debugging. i'll reuse.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
WillChilds-Klein: added the other cases, also added -1/+1 for all secret sizes.

csrc/aes_kwp.cpp
```diff
@@ -86,16 +82,11 @@ JNIEXPORT int JNICALL Java_com_amazon_corretto_crypto_provider_AesKeyWrapSpi_unw
         }
 
         jni_borrow inbuf(env, input, "input");
-        input.get_bytes(env, inbuf.data(), 0, input.len());
         jni_borrow outbuf(env, output, "output");
         size_t outlen;
         if (!AES_unwrap_key_padded(&aes_key, outbuf.data(), &outlen, input.len(), inbuf.data(), input.len())) {
```
SalusaSecondus, line 87: Fourth parameter still looks wrong. Shouldn't it be `outbuf.len()`?

Also, please initialize `outlen` to some value. Uninitialized variables are dangerous. (Yes, there are still some in the code-base, but we should fix them as we find them.)

Finally, it's better style to pull `len()` from the `jni_borrow` (as that's the object you're actually using) rather than from the `java_buffer`. There are edge-cases where borrows are not the same length as the underlying buffer. (At least I think there might be. Check the APIs to create a borrow.)

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -48,41 +48,42 @@
...
+                args.add(Arguments.of(aeskeySize, secretSize+1));
             }
         }
         return args;
     }
 
     @ParameterizedTest
     @MethodSource("sizes")
-    public void roundtrip(int keySize, int secretSize) throws Exception {
+    public void roundtrip(int aeskeySize, int secretSize) throws Exception {
```
SalusaSecondus, line 71: Wonderful!

Now, make sure you have some cases which actually wrap and unwrap real RSA and EC keys. You can probably do something like this:

```java
private void roundTripAsymmetric(SecretKey wrappingKey, KeyPairGenerator kg, KeyFactory kf) {
    KeyPair pair = kg.generate();
    Cipher c = Cipher.getInstance("AES/KWP/NoPadding", TestUtil.NATIVE_PROVIDER);
    c.init(Cipher.WRAP_MODE, wrappingKey);
    byte[] wrappedPrivateKey = c.wrap(pair.getPrivate());
    byte[] wrappedPublicKey = c.wrap(pair.getPublic());

    // Compatibility checks with the OpenJDK and BC???

    c.init(Cipher.UNWRAP_MODE, wrappingKey);
    Key unwrappedPrivate = c.unwrap(wrappedPublic, pair.getPrivate().getAlgorithm(), Cipher.PRIVATE_KEY);
    Key unwrappedPublic = c.unwrap(wrappedPublic, pair.getPublic().getAlgorithm(), Cipher.PUBLIC_KEY);

    // By passing it through the factory we ensure that it is an understandable type
    assertHexArrayEquals(kf.getKeySpec(pair.getPrivate(), PKCS8EncodedKeySpec.class).getEncoded(), kf.getKeySpec(unwrappedPrivate , X509EncodedKeySpec.class).getEncoded());
    assertHexArrayEquals(kf.getKeySpec(pair.getPublic(), X509EncodedKeySpec.class).getEncoded(), kf.getKeySpec(unwrappedPublic , X509EncodedKeySpec.class).getEncoded());

    // Compatibility checks with the OpenJDK and BC???
}
```

csrc/aes_kwp.cpp
WillChilds-Klein: > Fourth parameter still looks wrong. Shouldn't it be outbuf.len()?

yep, miss on my part. we should be bounding the max output buffer write size by the length of the output buffer. change made.

> Also, please initialize outlen to some value. Uninitialized variables are dangerous. (Yes, there are still some in the code-base, but we should fix them as we find them.)

makes sense. i considered initializing it to an invalid value (i.e. `-1`), but went with `0` as you suggest because that is also correct in the failure case (i.e. if the operation fails, 0 bytes will (should?) be written by AWS-LC).

> Finally, it's better style to ...

fair enough, that makes sense. if we derive `outbuf` from something other than the output array parameter down the line, then using the output array's length would risk an overrun.

> There are edge-cases where borrows are not the same length as the underlying buffer. (At least I think there might be. Check the APIs to create a borrow.)

hm, i'm not [seeing any today](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/csrc/buffer.h#L326) but that doesn't mean there won't be tomorrow. point taken that we should be using the length value of the closest-at-hand buffer wrapping object that we have in a particular scope or function.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
WillChilds-Klein: will do, thank you. i have the compatibility tests with BC working locally across "generic" keys, am going to also work on parameterizing across:

- JCE compatibility for [Java >=17](https://bugs.openjdk.org/browse/JDK-8248268)
- RSA secrets across valid key sizes
- EC secrets across valid key sizes

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -46,50 +47,111 @@
...
+        } else {
+            assertTrue(unwrappingProvider == null);
+        }
         c.init(Cipher.UNWRAP_MODE, key, sr);
         Key unwrapped = c.unwrap(wrapped, "Generic", Cipher.SECRET_KEY);
-        //System.out.println("KEY:    " + Arrays.toString(key.getEncoded()));
-        //System.out.println("SECRET: " + Arrays.toString(secret.getEncoded()));
-        //System.out.println("WRAPPD: " + Arrays.toString(wrapped));
-        //System.out.println("UNWRAP: " + Arrays.toString(unwrapped.getEncoded()));
         assertTrue(Arrays.equals(secret.getEncoded(), unwrapped.getEncoded()));
```
SalusaSecondus, line 125: NitPick: [assertArraysHexEquals](https://github.com/corretto/amazon-corretto-crypto-provider/blob/3ad0ce7f46d4c0f5ef13ae7868502c10319af0de/tst/com/amazon/corretto/crypto/provider/test/TestUtil.java#L67) is your friend.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -46,50 +47,111 @@
...
+    public void roundtripBC2native(int aesKeySize, int secretSize) throws Exception {
+        roundtrip(aesKeySize, secretSize, TestUtil.BC_PROVIDER, TestUtil.NATIVE_PROVIDER, false);
+    }
+
+    // TODO [childw] add new compatibility test cases for JCE providers that skip if < Java17
+    // TODO [childw] add parameter for SecretKey type, parameterize that out across EC, RSA, etc.
+    //               bonus points if we can auto-detect valid key sizes and filter the overarching
+    //               secret key parameter sizes into these parameterized tests for particular key sizes.
+    private void roundtrip(int aesKeySize, int secretSize, Provider wrappingProvider, Provider unwrappingProvider, boolean reuseCipher) throws Exception {
         final SecureRandom sr = new SecureRandom();
```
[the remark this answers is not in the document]
WillChilds-Klein: nice, thanks for the tip.

WillChilds-Klein:
currently seeing some strange failures for BouncyCastle-to-native round trips for EC keys. BC appears to be adding some bytes in the middle of the key. i'm digging into the [ASN.1-defining RFC for EC private keys](https://datatracker.ietf.org/doc/html/rfc5915) to figure out what's going on here. i've included a sample below, separating runs of contiguously matching or mismatching hex bytes by a space.

```
ACCP: 3081 87 020100301306072a8648ce3d020106082a8648ce3d03010704 6d306b 0201010420 5e7a4c9f3799d995a83e7fd084ce1735e7c15cb6286b1697bee9c4fc87ae3009a                          144034200048253a6bc4f5cdc5d5c6e54e56d1de67eb30de7067e664f6422f6a83eee4fbd449b1ccdd961e279123f26dc662b80fba0b08626a1585a2cd7fe423d0d5f8f1c80
BC  : 3081 93 020100301306072a8648ce3d020106082a8648ce3d03010704 793077 0201010420 5e7a4c9f3799d995a83e7fd084ce1735e7c15cb6286b1697bee9c4fc87ae3009a 00a06082a8648ce3d030107a 144034200048253a6bc4f5cdc5d5c6e54e56d1de67eb30de7067e664f6422f6a83eee4fbd449b1ccdd961e279123f26dc662b80fba0b08626a1585a2cd7fe423d0d5f8f1c80

```

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
```diff
@@ -166,6 +195,86 @@ private void implInit(int opmode, Key key, SecureRandom ignored)
...
+            System.arraycopy(out, 0, tmp, 0, tmp.length);
+            Arrays.fill(out, (byte) 0);
+            out = tmp;
+        }
+        return out;
+    }
+
+    private int implDoFinal(byte[] in, int inOfs, int inLen, byte[] out, int outOfs) {
+        implUpdate(in, inOfs, inLen);
+        buffer.trim();
```
SalusaSecondus, line 261: Why do this instead of passing down the length as a parameter? Trimming costs extra byte copies and you can pass in a maximum length when instantiating `java_buffer`.

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
```diff
@@ -44,15 +44,19 @@ final class AesKeyWrapSpi extends CipherSpi {
...
     private byte[] keyBytes;
     private int opmode = -1;    // must be set by init(..)
+    private final AccessibleByteArrayOutputStream buffer;
 
     AesKeyWrapSpi(final AmazonCorrettoCryptoProvider provider) {
         Loader.checkNativeLibraryAvailability();
         this.provider = provider;
+        // set initial capacity for the buffer to 0 because we need it to grow
+        // _precisely_ with the amount of data we're buffering.
+        buffer = new AccessibleByteArrayOutputStream(0, Integer.MAX_VALUE);
```
SalusaSecondus, line 54: This comment doesn't make sense, nor the requirement. It probably makes sense to initialize it to 16 or 32 bytes.

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
WillChilds-Klein: ah, nice. i didn't see that overload of `java_buffer::from_array`. thanks for the tip.

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
WillChilds-Klein: this was a change i made while debugging yesterday, i forgot to revisit it. i've reverted to using the default no-arg constructor which initializes the internal buffer to 32 bytes.

CMakeLists.txt
alexw91: Opened a tracking issue: https://github.com/corretto/amazon-corretto-crypto-provider/issues/239

src/com/amazon/corretto/crypto/provider/AesKeyWrapPaddingSpi.java
```diff
@@ -0,0 +1,304 @@
...
+                throw new InvalidKeyException("Key doesn't support encoding");
+            }
+            jceKey = (SecretKey) key;
+        }
+        this.opmode = opmode;
+        buffer.reset();
+    }
+
+    @Override
+    protected byte[] engineUpdate(byte[] in, int inOfs, int inLen) {
```
alexw91, line 178: Naming nit, I'd prefer the more readable `inOffset` rather than `inOfs` throughout this file.

WillChilds-Klein: sure, i'll adjust that for `inOfs` and `outOfs`.

WillChilds-Klein:
x86 CI check is failing due to issues addressed in [PR #243](https://github.com/corretto/amazon-corretto-crypto-provider/pull/243). once that PR is merged and the docker images are rebuilt in the CI account, it should work fine.

alexw91 (approved):
Looks good! Just had one small comment about an unused variable.

src/com/amazon/corretto/crypto/provider/AesKeyWrapPaddingSpi.java
```diff
@@ -0,0 +1,304 @@
...
+
+
+final class AesKeyWrapPaddingSpi extends CipherSpi {
+    private static final int BLOCK_SIZE = 128 / 8;
+
+    static {
+        Loader.load();
+    }
+    private final AmazonCorrettoCryptoProvider provider;
+    private NativeResource context = null;
```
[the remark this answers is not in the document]
WillChilds-Klein: yes, i think that's vestigial from an initial stab at using NativeResource. i decided to defer that for a simpler initial implementation. if benchmarking/profiling shows that key re-initialization is a significant bottleneck, i'll revisit using NativeResource.

for now, i've removed the unused variable.

alexw91 (approved):

PR #236, opened 2022-07-18
Add support for ACCP AL2 CI

samuel40791765:
*Issue #, if available:*
`CryptoAlg-1178`

*Description of changes:*
Adding support for AL2 ACCP testing within our CI. This change also removes the newly introduced `-DcmakeBin` property in gradle. I've made detecting `cmake3` within our gradle scripts "automatic". If `cmake3` is not detected within the system, we'll fall back to regular `cmake`. OS that have `cmake3` are known to cap their regular `cmake` version below cmake versions 3. 
AL2 CI changes include AL2 + corretto 8,11, 17 on both ARM and x86. Tested on https://github.com/samuel40791765/amazon-corretto-crypto-provider/pull/16 (Windows failure can be ignored)


tests/ci/docker_images/linux-arm/amazonlinux-2_accp_base/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+FROM arm64v8/amazonlinux:2
+
+SHELL ["/bin/bash", "-c"]
+
+RUN set -ex && \
+    yum -y update && yum install -y \
+    cmake \
+    cmake3 \
```
WillChilds-Klein, line 11: instead of doing the `cmake3`-autodetect workaround, would could we just install cmake at version 3? or perhaps use [update-alternatives](https://linux.die.net/man/8/update-alternatives) to manage which executable `cmake` command points to?

overall, this is just kind of a nitpick on my part -- i don't have any real issue with doing the autodetect in build.gradle, it just seems like a  slight conflation of concerns between "how to build the thing" (i.e. build.gradle) and "what to build the thing with" (i.e. our Dockerfiles).

tests/ci/docker_images/linux-arm/amazonlinux-2_gcc-7x_corretto/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN mkdir /accp
+COPY . /accp
+WORKDIR /accp
+
+# run the gradlew script just to install gradle in the image
+RUN ./gradlew --no-daemon generateEclipseClasspath
+
+ENV CC=gcc
+ENV CXX=g++
+ENV JAVA_HOME=/usr/lib/jvm/java-11-amazon-corretto.aarch64
```
WillChilds-Klein, line 35: why set the JAVA_HOME to 11 here? won't all the extending Dockerfiles take care of that?

tests/ci/docker_images/linux-x86/amazonlinux-2_gcc-7x_corretto/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN mkdir /accp
+COPY . /accp
+WORKDIR /accp
+
+# run the gradlew script just to install gradle in the image
+RUN ./gradlew --no-daemon generateEclipseClasspath
+
+ENV CC=gcc
+ENV CXX=g++
+ENV JAVA_HOME=/usr/lib/jvm/java-11-amazon-corretto.x86_64
```
WillChilds-Klein, line 35: same question about JAVA_HOME

WillChilds-Klein (approved):

tests/ci/docker_images/linux-arm/amazonlinux-2_gcc-7x_corretto/Dockerfile
samuel40791765: The ACCP build complains when `JAVA_HOME` isn't set with an error like this:
```
CMake Error at /usr/share/cmake3/Modules/FindJava.cmake:166 (message):
  Error executing java -version
Call Stack (most recent call first):
  CMakeLists.txt:12 (find_package)


-- Configuring incomplete, errors occurred!
See also "/home/ubuntu/workplace/amazon-corretto-crypto-provider/build/cmake/CMakeFiles/CMakeOutput.log".
```
It seems like CMake relies on the variable to look for the java binary.

tests/ci/docker_images/linux-arm/amazonlinux-2_accp_base/Dockerfile
samuel40791765: Unfortunately on AL2 & centOS, the version of `cmake` installed is always capped at `2.8.x`. The OS default only provides Cmake versions over 3 with the `cmake3` command. The `update-alternatives` method is true from a development environment standpoint. 
The main reason I thought of configuring `cmake3` auto-detection was thinking of customers using ACCP on these platforms. We used to directly call the `cmake` command within `build.gradle` and this would force all customers on these platforms to point to the correct `cmake/cmake3` executable themselves. Using this method would let them use ACCP out of the box without any extra configurations.

WillChilds-Klein: > Unfortunately on AL2 & centOS, the version of cmake installed is always capped at 2.8.x. The OS default only provides Cmake versions over 3 with the cmake3 command

ah, bummer. that's an unfortunate restriction, but I'm sure there are reasons.

agreed that better usability is worth trading for any aesthetic nits I may have :)

tests/ci/docker_images/linux-arm/amazonlinux-2_gcc-7x_corretto/Dockerfile
WillChilds-Klein: ok, that makes sense  that it would need to know where java lives in order to build ACCP. i suppose it doesn't matter whether we choose java 11 or 17 to build with; IIRC we can't build with java 8.

WillChilds-Klein:
LGTM, thanks sam!

PR #237, opened 2022-07-20
Add CI support for old version of gcc

samuel40791765:
*Issue #, if available:*
`CryptoAlg-1166`

*Description of changes:*
Adding support for centos7/gcc4.8 since it's an OS which has an older gcc version and still has support for >= JAVA10. I've tried using the ubuntu7.04 docker image from AWS-LC (https://github.com/awslabs/aws-lc/commit/1e64736db3f709a7e9dd0f016702d42cb29a0902) since it replicates the internal RHEL5 gcc the best, but there are multiple problems with getting a proper version of JAVA working on there. More details can be found under the thread of the SIM.


samuel40791765:
In draft for now since this PR relies on some the new `cmake3` behavior in https://github.com/corretto/amazon-corretto-crypto-provider/pull/236

tests/ci/docker_images/linux-x86/centos-7_gcc-4x/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+# https://github.com/golang/go/issues/38373
+RUN set -ex && \
+    yum -y update && \
+    yum --enablerepo=extras install epel-release -y && \
+    yum -y install \
+    gcc \
+    libgcc.i686 \
+    glibc-devel.x86_64 \
+    glibc-devel.i686 \
+    gcc-c++ \
```
WillChilds-Klein, line 22: how can we be assured that this installs the GCC version we intend to test against?

samuel40791765:
Replacing this PR with https://github.com/corretto/amazon-corretto-crypto-provider/pull/240

PR #238, opened 2022-07-29
Report coverage in CodeBuild

WillChilds-Klein:
*Issue #, if available:* n/a

*Description of changes:*

This commit configures coverage reports for CodeBuild, as documented
[here][1]. This is very straightforward for our Jacoco java coverage
reports, as CodeBuild natively supports that report format. For our gcov
C++ coverage reports for libaccp, however, we need to install and use
[`gcovr`][2] to convert gcov reports to a compatible format. Coverage is
only reported for our basic tests (either via the `release` target or,
in the case of Java8, explicitly specifying the `coverage` target
alongside `test`).

CodeCommit's coverage reports should be useful for PR reviewers, but
they do not provide granular, line-by-line coverage information or a
diff overlay view. If we want something more granular, we may need to
evaluate 3rd party tools [like codecov][3].

This commit also addresses [Issue #239][4], using CMake's [`file(GLOB`
function][5] to find relevant src and test files, automatically
generating the file of files to include java in builds. This eliminates
the need to manually enumerate our src/tst files in CMakeLists.txt. We
use the `CONFIGURE_DEPENDS` flag, instructing CMake to watch for new (or
removed) files in those directories, and regenerate the build system if
it detects a change. However, this feature is not available until 3.12,
so we need to condition its use on CMake version (per @sacchaing, RHEL5
only supports up to CMake 3.9).

[1]: https://docs.aws.amazon.com/codebuild/latest/userguide/code-coverage-report.html
[2]: https://gcovr.com/en/stable/
[3]: https://docs.codecov.com/docs/github-checks
[4]: https://github.com/corretto/amazon-corretto-crypto-provider/issues/239
[5]: https://cmake.org/cmake/help/latest/command/file.html#glob

---

# Testing

Checks are currently failing on this PR because the CI Docker images need to be rebuilt in order to incorporate a new tool dependency, `gcovr`. I've rebuilt the images in my dev account, and have run the checks successfully as part of [PR #1 on my fork](https://github.com/WillChilds-Klein/amazon-corretto-crypto-provider/pull/1). I'll hold off on rebuilding images in our CI account until these changes are approved.

Here are some isenlinks to the reports:

- [successful build batch in my dev account](https://tiny.amazon.com/1d41k7ob9/IsenLink)
- [sample report set for above batch (unit tests, java coverage, and cpp coverage reports)](https://tiny.amazon.com/xi3k3gx9/IsenLink)

Here are some sample screenshots of an individual java report and coverage trends across builds:

Screen Shot 2022-08-01 at 10 36 46 AM

Screen Shot 2022-08-01 at 10 36 20 AM

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,48 +130,16 @@ add_custom_command(
     COMMENT "Generating hash function SPI classes..."
     )
 
-set(ACCP_SRC
-        src/com/amazon/corretto/crypto/provider/AccessibleByteArrayOutputStream.java
-        src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
-        src/com/amazon/corretto/crypto/provider/ConstantTime.java
-        src/com/amazon/corretto/crypto/provider/DebugFlag.java
-        src/com/amazon/corretto/crypto/provider/EcGen.java
-        src/com/amazon/corretto/crypto/provider/EvpHmac.java
-        src/com/amazon/corretto/crypto/provider/EvpKeyAgreement.java
-        src/com/amazon/corretto/crypto/provider/EvpKey.java
-        src/com/amazon/corretto/crypto/provider/EvpEcKey.java
-        src/com/amazon/corretto/crypto/provider/EvpEcPublicKey.java
-        src/com/amazon/corretto/crypto/provider/EvpEcPrivateKey.java
-        src/com/amazon/corretto/crypto/provider/EvpRsaKey.java
-        src/com/amazon/corretto/crypto/provider/EvpRsaPublicKey.java
-        src/com/amazon/corretto/crypto/provider/EvpRsaPrivateKey.java
-        src/com/amazon/corretto/crypto/provider/EvpRsaPrivateCrtKey.java
-        src/com/amazon/corretto/crypto/provider/EvpKeyFactory.java
-        src/com/amazon/corretto/crypto/provider/EvpKeyType.java
-        src/com/amazon/corretto/crypto/provider/EvpSignature.java
-        src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
-        src/com/amazon/corretto/crypto/provider/EvpSignatureRaw.java
-        src/com/amazon/corretto/crypto/provider/EcUtils.java
-        src/com/amazon/corretto/crypto/provider/ExtraCheck.java
-        src/com/amazon/corretto/crypto/provider/InputBuffer.java
-        src/com/amazon/corretto/crypto/provider/Janitor.java
-        src/com/amazon/corretto/crypto/provider/LibCryptoRng.java
-        src/com/amazon/corretto/crypto/provider/Loader.java
-        src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
-        src/com/amazon/corretto/crypto/provider/NativeResource.java
-        src/com/amazon/corretto/crypto/provider/MiscInterfaces.java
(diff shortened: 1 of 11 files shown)


CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.12)
```
samuel40791765, line 4: I'm not sure if this alright to upgrade, the internal RHEL5 Cmake version is capped at 3.9

WillChilds-Klein: ah, that's a bummer. turns out the `CONFIGURE_DEPENDS` feature was introduced in 3.12, and it's a nice-to-have because we won't need to `clean` after adding files. from [the docs](https://cmake.org/cmake/help/latest/command/file.html#glob):

> New in version 3.12: If the CONFIGURE_DEPENDS flag is specified, CMake will add logic to the main build system check target to rerun the flagged GLOB commands at build time. If any of the outputs change, CMake will regenerate the build system.

i think we can keep the min version as-is and condition on version in the `file(GLOB ... )` call.

samuel40791765: No worries, It's a huge pain.. Main reason seems to be Cmake versions after 3.10 require C++11 which RHEL5 doesn't have. The if else statement at Line 11-29 is also there to get around some non-existent features in older versions of cmake.

WillChilds-Klein:
note to reviewers -- please let me know whether you think the Dockerfile changes look OK. if they look good, i'll rebuild the docker images in our CI account and re-run the CI so other reviewers can see all the little green checks, mess around with the reports, etc. before approving.

CMakeLists.txt
```diff
@@ -130,48 +130,16 @@ add_custom_command(
...
-        src/com/amazon/corretto/crypto/provider/SelfTestSuite.java
-        src/com/amazon/corretto/crypto/provider/ServiceProviderFactory.java
-        src/com/amazon/corretto/crypto/provider/Utils.java
-        ${GENERATED_JAVA_SRC}
-)
+# NOTE: CMake introduced the CONFIGURE_DEPENDS feature in 3.12, so condition
+#       its use on CMake version. If the feature is not available, you may need
+#       to `touch CMakeLists.txt` after adding/removing files for them to be
+#       detected by CMake.
+if (${CMAKE_VERSION} VERSION_LESS "3.12.0")
```
samuel40791765, line 137: Nit: If we already have an alternative for Cmake versions below `3.12.0`, is the second sentence necessary? same for the comment on Line 524

samuel40791765 (approved):

CMakeLists.txt
WillChilds-Klein: i think so, my understanding is that without `CONFIGURE_DEPENDS`, cmake won't automatically detect new files. one way to force a re-build of the build system is to `touch CMakeLists.txt`. from [the docs](https://cmake.org/cmake/help/latest/command/file.html#filesystem):

> New in version 3.12: If the CONFIGURE_DEPENDS flag is specified, CMake will add logic to the main build system check target to rerun the flagged GLOB commands at build time. If any of the outputs change, CMake will regenerate the build system.

samuel40791765: Ah that's quite interesting, I misunderstood touching the file as "manually configuring it". Then I guess the comment's pretty helpful.

Issue #41, opened 2019-07-11
Support OSX


SalusaSecondus:
Fully Support OSX

kiwi1969:
I am interested in what this ticket is asking...
Can someone please expand on what is actually missing in OSX (if anything)

Ie. Today I added the ACCP Linux x86 jar to CLASSPATH via .zlogin script, installed Corretto8 JDK, changed java.security file to have ACCP as first in list.
When I run a test to show active providers, it prints the AACP as first in list, showing it is indeed installed.
If I remove from classpath, I get the previous list without AACP
From reading, I could use other JDK such as Adopt OpenJDK, and AACP should still work.

I believe ACCP has no impact on OpenSSL, Boring SSL, Libre SSL, Conscript, OKHTTP, but will enhance certain other security methods. I believe OSX pretends to have OpenSSL but has own internals, but can be thought of as OpenSSL
(Please correct me if I am wrong with the no-impact on OpenSSL comment)

SalusaSecondus:
ACCP uses a native library to perform its cryptography which contains a statically linked copy of OpenSSL. The form of these libraries is different depending on operating system. (Linux uses SO files, Windows uses DLLs, and OSX uses dynlibs.) So, while you may have seen ACCP in your provider list, I would be surprised it if advertised any supported algorithms. Instead, it detected that it was on an unsupported system and silently disabled itself so that Java falls back to the standard implementations and continues to work. (This makes it safe to use ACCP even when you aren't sure if the end system will support it. They may not have the performance benefits, but at least they will still run.)

I recommend trying your experiment and calling this method to see if ACCP is actually fully functional:
```java
AmazonCorrettoCryptoProvider.INSTANCE.assertHealthy();
```

I'm not sure what your question is from the last part of your comment. What do you mean by "no impact" and what are you trying to achieve?

Taffer:
If ACCP is now using AWS-LC for crypto, is this Issue obsolete now?

samuel40791765:
Not necessarily, although AWS-LC is buildable on OSX, ACCP itself still doesn't fully have build support for OSX yet. The same issues when ACCP was building with OpenSSL still apply.

PR #246, opened 2022-09-08
Support MacOS

WillChilds-Klein:
*Issue #, if available:* #41

# Notes
This commit adds functional and CI support for MacOS. The main blocker
on supporting MacOS was a runtime linking error caused by libaccp trying
to load a `libcrypto` it found in some `/usr` system directory on MacOS,
instead of the temporary directory that ACCP creates and restricts
`rpath` to during linking. The reason for this is that the `rpath` value
of `$ORIGIN` is specific to linux's linker, and is meaningless on MacOS.
So, the MacOS linker consulted a non-existent `rpath`, didn't find
anything, and fell back to system library directories. The fix is to
detect OS type at build time and use the appropriate value for `rpath`
(on MacOS that's `@loader_path`, see the "Locating External Resources"
section of [this documentation page][1]).  We also explicity set
`CMAKE_MACOSX_RPATH` on MacOS, which is documented [here][2].

To prevent regressions expanding RPATH beyond one entry, we add a
test check that asserts that libaccp only has a single RPATH entry
that's set to the appropriate value on all supported platforms.

The other pieces of this commit are related to CI. In order for the
coverate steps to pass, we need to install `lcov` and `gcovr`
dependencies on the GitHub Actions MacOS CI images. Lastly, now that
this runs on Mac the Apple JCE providers in our pre-baked properties
files are actually used (they're ignored on Linux, where they're
unavailable). This resulted in the same provider being installed twice,
causing SecurityPropertiesTester failures. We resolve this issue by
removing the class-specified entry in favor of the name-specified entry.
Because the [AppleProvider][3] is [not exported][4] from its module in
JDK 11, we cannot use the fully qualified class specification (see
comment in `etc/amazon-corretto-crypto-provider.security`) on JDK 9+. By
using only the name specification, there may be some issues on JDK8 on
Mac when using this providers file, but the impact should be minor -- no
regressions (because we didn't support MacOS before). Users are
encouraged to upgrade to more moder JDKs or manually modify/override the
supplied properties file.

This commit resolves [Issue #41][5].

[1]: https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryDesignGuidelines.html#//apple_ref/doc/uid/TP40002013-SW21
[2]: https://cmake.org/cmake/help/latest/prop_tgt/MACOSX_RPATH.html
[3]: https://docs.oracle.com/javase/8/docs/technotes/guides/security/crypto/CryptoSpec.html#Provider
[4]: https://github.com/corretto/corretto-11/blob/develop/src/java.base/macosx/classes/module-info.java.extra
[5]: https://github.com/corretto/amazon-corretto-crypto-provider/issues/41

# Testing

## Manual Inspection of libaccp RPATH

On MacOS:

```
$ sw_vers
ProductName:    macOS
ProductVersion: 12.5.1
BuildVersion:   21G83

$ uname -a
Darwin 08f8bc77642e.ant.amazon.com 21.6.0 Darwin Kernel Version 21.6.0: Wed Aug 10 14:25:27 PDT 2022; root:xnu-8020.141.5~2/RELEASE_X86_64 x86_64

$ otool -l ./build/cmake/libamazonCorrettoCryptoProvider.dylib | grep -A2 ' cmd LC_RPATH'
          cmd LC_RPATH
      cmdsize 32
         path @loader_path (offset 12)
```

On Ubuntu:

```
root@817bfef6156a:/accp# cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal

root@817bfef6156a:/accp# chrpath -l ./build/cmake/libamazonCorrettoCryptoProvider.so
./build/cmake/libamazonCorrettoCryptoProvider.so: RUNPATH=$ORIGIN
```

On AL2:

```
$ cat /etc/os-release
NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2:-:internal"
HOME_URL="https://amazonlinux.com/"
VARIANT="internal"

$ chrpath -l ./build/cmake/libamazonCorrettoCryptoProvider.so
./build/cmake/libamazonCorrettoCryptoProvider.so: RPATH=$ORIGIN
```

## RPATH Regression Testing Mechanism

Modified local code to include an extra RPATH entry.

Local changes:

```
$ git diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 0252625..da83da3 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,8 +86,6 @@ string(REPLACE ":" ";" TEST_CLASSPATH_LIST "${TEST_CLASSPATH}")

 # Needed as we abuse some of the test compile macros to test shared lib links
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
-# CMake's default RPath handling includes the full build directory path, which isn't needed. Turn it off.
-set(CMAKE_SKIP_RPATH TRUE)
 set(GENERATE_HASHERS ${CMAKE_CURRENT_SOURCE_DIR}/build-tools/bin/generate-java-hash-spi)
 set(GENERATED_JAVA_DIR ${CMAKE_CURRENT_BINARY_DIR}/generated-java/com/amazon/corretto/crypto/provider)
 set(JNI_HEADER_DIR ${CMAKE_CURRENT_BINARY_DIR}/generated-include)
```

On MacOS:

```
$ ./gradlew clean && ./gradlew test
...
/Library/Developer/CommandLineTools/usr/bin/make  -f CMakeFiles/check-libaccp-rpath.dir/build.make CMakeFiles/check-libaccp-rpath.dir/build
otool -l /Users/childw/workplace/github/corretto/amazon-corretto-crypto-provider/build/cmake/libamazonCorrettoCryptoProvider.dylib | grep 'cmd LC_RPATH' | wc -l | tr -d '[:blank:]' | xargs test 1 -eq
...
BUILD FAILED in 54s
3 actionable tasks: 2 executed, 1 up-to-date
```

On Ubuntu:

```
$ ./gradlew clean && ./gradlew test
...
make[3]: Entering directory '/accp/build/cmake'
objdump -x /accp/build/cmake/libamazonCorrettoCryptoProvider.so | grep -E 'RPATH|RUNPATH' | grep -v ':'
make[3]: *** [CMakeFiles/check-libaccp-rpath.dir/build.make:60: CMakeFiles/check-libaccp-rpath] Error 1
...
BUILD FAILED in 1m 28s
3 actionable tasks: 2 executed, 1 up-to-date
```

On AL2:

```
$ ./gradlew clean && ./gradlew test
...
make[3]: Entering directory `/local/home/childw/workplace/github/corretto/amazon-corretto-crypto-provider/build/cmake'
objdump -x /local/home/childw/workplace/github/corretto/amazon-corretto-crypto-provider/build/cmake/libamazonCorrettoCryptoProvider.so | grep -E 'RPATH|RUNPATH' | grep -v ':'
...
BUILD FAILED in 1m 31s
3 actionable tasks: 3 executed
```

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,7 +269,15 @@ add_library(
 # We MUST set ACCP's RPATH to "$ORIGIN" so that the runtime dynamic loader will look in the same directory as ACCP for
 # any shared library dependencies BEFORE looking for a system-installed depencencies (such as libcrypto). If RPath is
 # not set, the loader may attempt to load an incompatible system-installed libcrypto and cause runtime crashes.
-set_target_properties(amazonCorrettoCryptoProvider PROPERTIES LINK_FLAGS "-Wl,-rpath,\$ORIGIN")
+if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    set(CMAKE_MACOSX_RPATH TRUE)
+    set(RPATH_ORIGIN "@loader_path")
+elseif(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+    set(RPATH_ORIGIN "\$ORIGIN")
+else()
+    message( FATAL_ERROR "Unsupported OS, need to determine OS-specific analog of linux rpath's $ORIGIN")
+endif()
+set_target_properties(amazonCorrettoCryptoProvider PROPERTIES LINK_FLAGS "-Wl,-rpath,${RPATH_ORIGIN}")
 
 add_custom_command(
     OUTPUT ${ACCP_JAR_SOURCE}
@@ -286,7 +294,7 @@ if(ENABLE_NATIVE_TEST_HOOKS)
   add_executable(test_keyutils EXCLUDE_FROM_ALL
         csrc/test_keyutils.cpp
     )
-    set_target_properties(test_keyutils PROPERTIES LINK_FLAGS "-Wl,-rpath,\$ORIGIN")
+    set_target_properties(test_keyutils PROPERTIES LINK_FLAGS "-Wl,-rpath,${RPATH_ORIGIN}")
     target_link_libraries(test_keyutils ${OPENSSL_CRYPTO_LIBRARY})
     target_link_libraries(test_keyutils amazonCorrettoCryptoProvider)
 endif()
@@ -723,7 +731,7 @@ add_custom_target(check-install-via-properties-with-debug
     DEPENDS accp-jar tests-jar)
 
 add_custom_target(check
-    DEPENDS check-recursive-init check-install-via-properties check-install-via-properties-with-debug check-junit check-junit-SecurityManager check-external-lib)
+    DEPENDS check-recursive-init check-install-via-properties check-install-via-properties-with-debug check-junit check-junit-SecurityManager check-external-lib check-libaccp-rpath)
 
 add_custom_target(checkstyle
     WORKING_DIRECTORY ${ORIG_SRCROOT}
@@ -837,5 +845,34 @@ add_custom_target(check-dieharder-threads
 
     DEPENDS run-dieharder-libcrypto-rng-threads run-dieharder-libcrypto-rng-threads-tail)
 endif() # End of Dieharder targets
+
+# Add a target to assert that the libaccp shared object's rpath meets 2 conditions:
+#   1. rpath contains only a single entry
+#   2. that entry is set to $ORIGIN on linux or @loader_path on mac
+if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    add_custom_target(check-libaccp-rpath
+        COMMAND otool -l $<TARGET_FILE:amazonCorrettoCryptoProvider>
+            | grep 'cmd LC_RPATH'
+            | wc -l
+            | tr -d '[:blank:]'
+            | xargs test 1 -eq
+        COMMAND otool -l $<TARGET_FILE:amazonCorrettoCryptoProvider>
+            | grep -A2 'cmd LC_RPATH'
+            | tail -1
+            | grep '@loader_path'
+        )
+elseif(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+    add_custom_target(check-libaccp-rpath
+        COMMAND objdump -x $<TARGET_FILE:amazonCorrettoCryptoProvider>
+            | grep -E 'RPATH|RUNPATH'   # RPATH on AL, RUNPATH on Ubuntu
+            | grep -v ':'
+        COMMAND objdump -x $<TARGET_FILE:amazonCorrettoCryptoProvider>
+            | grep -E 'RPATH|RUNPATH'   # RPATH on AL, RUNPATH on Ubuntu
+            | grep -E "\$ORIGIN"
+        )
+else()
+    message( FATAL_ERROR "Unsupported OS, need to validate OS-specific rpath $ORIGIN")
+endif()
+
 # Do this at the end, after we finish all our feature tests, or it'll be missing flags
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/csrc/config.h.in ${JNI_HEADER_DIR}/config.h)
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ KeyFactory algorithms:
 # Compatibility & Requirements
 ACCP has the following requirements:
 * JDK8 or newer (This includes both OracleJDK and [Amazon Corretto](https://aws.amazon.com/corretto/))
-* 64-bit Linux running on x86_64 (also known as x64 or AMD64)
+* 64-bit Linux or MacOs running on x86_64 (also known as x64 or AMD64)
 
 If ACCP is used/installed on a system it does not support, it will disable itself and the JVM will behave as if ACCP weren't installed at all.
 
@@ -100,7 +100,8 @@ in the latest version for best performance and bug-fixes.
 
 Whether you're using Maven, Gradle, or some other build system that also pulls
 packages from Maven Central, it's important to specify `linux-x86_64` as the
-classifier. You'll get an empty package otherwise.
+classifier. You'll get an empty package otherwise. Note that ACCP will not be
+available for MacOS on Maven Central until 2.0 is released.
 
 Regardless of how you acquire ACCP (Maven, manual build, etc.) you will still need to follow the guidance in the [Configuration section](#configuration) to enable ACCP in your application.
 
@@ -147,7 +148,7 @@ The OracleJDK requires that JCA providers be cryptographically signed by a trust
 The JARs we publish via Maven and our official [releases](https://github.com/corretto/amazon-corretto-crypto-provider/releases) are signed by our private key,
 but yours will not be.*
 
-Building this provider requires a 64 bit Linux build system with the following prerequisites installed:
+Building this provider requires a 64 bit Linux or MacOS build system with the following prerequisites installed:
 * OpenJDK 10 or newer
 * [cmake](https://cmake.org/) 3.8 or newer
 * C++ build chain
--- a/etc/amazon-corretto-crypto-provider.security
+++ b/etc/amazon-corretto-crypto-provider.security
@@ -23,8 +23,14 @@
 # JDK8 will determine that there is no class with the short provider name and skip it.
 # JDK9+ will determine that the fully-qualified class is not visible and will skip it.
 #
-# This results in a single list with proper behavior on both JDK8 and JDK9+ systems.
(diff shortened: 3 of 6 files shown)


CMakeLists.txt
```diff
@@ -86,8 +86,6 @@ string(REPLACE ":" ";" TEST_CLASSPATH_LIST "${TEST_CLASSPATH}")
 
 # Needed as we abuse some of the test compile macros to test shared lib links
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
-# CMake's default RPath handling includes the full build directory path, which isn't needed. Turn it off.
-set(CMAKE_SKIP_RPATH TRUE)
```
alexw91, line 90: If I remember correctly, before I added this line, the RPath of `libamazonCorrettoCryptoProvider.so` on my Ubuntu desktop contained both of the following when running `chrpath libamazonCorrettoCryptoProvider.so`:
1. `/home/aweibel/workspace/github/amazon-corretto-crypto-provider/build/awslc/bin/lib/libcrypto.so` (The path I where I was building ACCP on my desktop.)
2. `$ORIGIN`

Adding these 2 lines caused the 1st RPath value of the temporary build directory to be removed from the resulting  output, since only the 2nd RPath value is needed. Deleting these two lines **might** re-add the temporary build path's back into the libACCP on Linux platforms, and this should be checked before merging. Theoretically, if there's an extra RPath value containing a hard-coded directory from where our CI is configured to build and publish ACCP from, then any user on a host that is using ACCP could recreate that hard-coded path and put a different `libcrypto.so` file there that would loaded into any Java processes using ACCP on that host.

We may want to add a check during the build that ensures that only 1 RPath value is present in `libamazonCorrettoCryptoProvider.so`, and that it's value is `$ORIGIN`.

WillChilds-Klein: nice catch. looks like deleting this line does indeed introduce multiple rpath entries (albeit in the order we want)

```
root@817bfef6156a:/accp# cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
root@817bfef6156a:/accp# chrpath -l ./build/cmake/libamazonCorrettoCryptoProvider.so
./build/cmake/libamazonCorrettoCryptoProvider.so: RUNPATH=$ORIGIN:/accp/build/awslc/bin/lib
```

agreed that rpath should have one entry and point to `$ORIGIN`. i'll walk this back and start looking at ways to prune the rpath programmatically on MacOS and also look into a build-time check that asserts our rpath requirements.

WillChilds-Klein:
checks are failing here due to CI weirdness, and we need to address the issues that alex called out before merging. but we're not far from full functionality. I have a PR [here](https://github.com/WillChilds-Klein/amazon-corretto-crypto-provider/pull/4) with all checks passing on MacOS.

CMakeLists.txt
WillChilds-Klein: ok, digging into it further, it looks like the root cause for lack of MacOS support before this commit was that rpath value of `$ORIGIN` doesn't mean anything on MacOS. As shown above, my prior approach included the aws-lc build dir in rpath, which i think is where `libcrypto.so` was being found in the initial draft of this PR. the fix was to use the MacOS equivalent of `$ORIGIN`, which is [`@loader_path`](https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryDesignGuidelines.html#//apple_ref/doc/uid/TP40002013-SW21). I've confirmed that in this PR's current version, there's only one RPATH entry (see testing section in the PR overview) and its set appropriately for mac and linux.

i also looked into adding a build-time check for this fact, but my CMake-fu is rather weak when it comes to invoking compound commands. in bash with `-e` and `-pipefail` set, i think the checks would look something like this:

mac:

```
# check that there's only one rpath value (probably not the best way to say "expect only one matching line")
[[ $(echo 1) == $(otool -l ./path/to/libamazonCorrettoCryptoProvider.dylib | ' cmd LC_RPATH' | wc -l)

# check that it's set to @loader_path
otool -l ./path/to/libamazonCorrettoCryptoProvider.dylib | grep -A2 ' cmd LC_RPATH' | tail -1 | grep 'libamazonCorrettoCryptoProvider'
```

linux:

```
# check that there's only one rpath value and that it's set to @loader_path
chrpath -l ./path/to/libamazonCorrettoCryptoProvider.so | grep -P 'RUNPATH=\$ORIGIN$'
```

etc/amazon-corretto-crypto-provider.security
```diff
@@ -46,7 +46,6 @@ security.provider.17=JdkSASL
 security.provider.18=sun.security.mscapi.SunMSCAPI
 security.provider.19=SunMSCAPI
 security.provider.20=apple.security.AppleProvider
-security.provider.21=Apple
```
alexw91, line 49: Was this change intended?

WillChilds-Klein: yes, there's an explanation in the last paragraph of the Notes/commit message.

CMakeLists.txt
WillChilds-Klein: test check added in most recent revision.

etc/amazon-corretto-crypto-provider.security
WillChilds-Klein: updated explanation:

> Now that this runs on Mac the Apple JCE providers in our pre-baked properties
> files are actually used (they're ignored on Linux, where they're unavailable).
> This resulted in the same provider being installed twice, causing
> SecurityPropertiesTester failures. We resolve this issue by removing the
> class-specified entry in favor of the name-specified entry.  Because the
> [AppleProvider][3] is [not exported][4] from its module in JDK 11, we cannot
> use the fully qualified class specification (see comment in
> `etc/amazon-corretto-crypto-provider.security`) on JDK 9+. By using only the
> name specification, there may be some issues on JDK8 on Mac when using this
> providers file, but the impact should be minor -- no regressions (because we
> didn't support MacOS before). Users are encouraged to upgrade to more moder
> JDKs or manually modify/override the supplied properties file.

[3]: https://github.com/corretto/corretto-11/blob/develop/src/java.base/macosx/classes/apple/security/AppleProvider.java
[4]: https://github.com/corretto/corretto-11/blob/develop/src/java.base/macosx/classes/module-info.java.extra

SalusaSecondus:
Please update the main readme to list the new support for MacOS. Thanks.

README.md
```diff
@@ -83,7 +83,7 @@ KeyFactory algorithms:
 # Compatibility & Requirements
 ACCP has the following requirements:
 * JDK8 or newer (This includes both OracleJDK and [Amazon Corretto](https://aws.amazon.com/corretto/))
-* 64-bit Linux running on x86_64 (also known as x64 or AMD64)
+* 64-bit Linux or MacOs running on x86_64 (also known as x64 or AMD64)
```
SalusaSecondus, line 86: It works on Linux and MacOS running on aarch64 also, doesn't it? :-)

WillChilds-Klein: presumably. we know that it works on aarch64 linux, but we haven't yet tested it yet on aarch64 MacOS as far as I know (I use an x86 mac, and it's unclear what the underlying architecture is for the MacOS jobs).

in any case, I wanted to keep the README guidance on architecture and OS hewed to whatever is available in maven until we publish an external release of ACCP 2.0. until then, maven-vended ACCP will still only be supported on x86 linux.

alexw91 (approved):

PR #247, opened 2022-10-07
Fix EC tests for OpenJDK10

WillChilds-Klein:
# Notes

The SunEC provider in OpenJDK10 [has a bug][1] that causes EC keys with
0-valued leading or trailing bytes to be truncated below their standard size.
This bug is particularly significant for 521-bit EC keys. Because such keys
have only a single bit in the most-significant byte, there is a 50% chance of
the leading byte being zero.  Presumably it also affects other key orders, but
with much lower probability. For orders divisible by 8, the probability of a
leading or trailing byte being zero is only 2\*(1/256) = 1/128. This bug was
[fixed][2] and backported [to OpenJDK11][3] and [to OpenJDK8][4] (as well as
[corretto 8][5] and up). However, it was not backported to JDK10 because JDK10
is End-of-Life'd (EoL'd).

Under this bug, OpenJDK will represent some EC keys' private value `s` with
non-standard byte array lengths.  ACCP always uses the standard byte array
length (i.e. the maximum number of bytes a key of that order can occupy).
Because both [ACCP][6] and [OpenJDK][7] determine equality by comparing DER
encoding byte-by-byte, under this bug some JDK10 keys with an `s` value
numerically equivalent to those of ACCP will be considered unequal.  Given that
this bug is only present in JDK10 (and notably not in _any_ Corretto version,
nor more recent OpenJDK versions), and that OpenJDK10 is EoL'd and should be
migrated away from, this incompatibility is acceptable.

[1]: https://github.com/openjdk/jdk10u/blob/master/src/jdk.crypto.ec/share/classes/sun/security/ec/ECPrivateKeyImpl.java#L94
[2]: https://github.com/openjdk/jdk11u/commit/eb894e22233a26111c096a7833f23fd7b1f7630f#diff-79d013d90386a4e617702723d7579887196260686b4ec06645b64ad08b00ffd6L94
[3]: https://github.com/openjdk/jdk11u/blob/master/src/jdk.crypto.ec/share/classes/sun/security/ec/ECPrivateKeyImpl.java#L118-L125
[4]: https://github.com/openjdk/jdk8u/blob/master/jdk/src/share/classes/sun/security/ec/ECPrivateKeyImpl.java#L124
[5]: https://github.com/corretto/corretto-8/blob/develop/jdk/src/share/classes/sun/security/ec/ECPrivateKeyImpl.java#L123-L136
[6]: https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/src/com/amazon/corretto/crypto/provider/EvpKey.java#L123-L153
[7]: https://github.com/openjdk/jdk10/blob/master/src/java.base/share/classes/sun/security/pkcs/PKCS8Key.java#L411-L423

---


src/com/amazon/corretto/crypto/provider/EvpEcPrivateKey.java
```diff
@@ -23,6 +24,17 @@ class EvpEcPrivateKey extends EvpEcKey implements ECPrivateKey, CanDerivePublicK
...
+    @Override
+    public boolean equals(final Object obj) {
+        final boolean parentEquals = super.equals(obj);
+        if (parentEquals) {
+            return parentEquals;
+        } else if (Utils.getJavaVersion() == 10 && (obj instanceof ECPrivateKey)) {
+            return this.getS().equals(((ECPrivateKey) obj).getS());
+        }
+        return parentEquals;
+    }
```
WillChilds-Klein, line 36: i'm not thrilled about this override.

on one hand, it might be useful that we have a shared notion of equality with JDK10 here for _some_ users (although JDK10 is long-since EOL'd and shouldn't be used).

on the other, it feels like we're taking a half-measure towards compatibility. we're gaining compatibility with JDK10 on object equality, but there's nothing we can do about compatibility on `hashCode` without DER-decoding the bytes, and re-encoding them having stripped the leading `0`-valued byte, which feels _very much_ like the wrong thing to do -- both for performance and because it accedes to a model of "bug-for-bug" compatibility that gets ugly quickly.

WillChilds-Klein: after digging into JDK history a bit more, i backed this override out. we need to prioritize java as it is, not java as it was.

tst/com/amazon/corretto/crypto/provider/test/EcGenTest.java
```diff
@@ -325,9 +325,12 @@ public void threadStorm() throws Throwable {
 
     public static void assertECEquals(final String message, final ECPrivateKey expected,
                                        final ECPrivateKey actual) {
+        final boolean expectEncodingCompat = TestUtil.getJavaVersion() != 10;
```
amirhosv, line 328: Could the implementation of `getEncoded` in Java 10 be switched in future or other versions decide to do similar things? If yes, in the test code, could we trim the zeros regardless of the Java version and always compare? It would be nice to make the tests not rely on implementation details so that if other versions of Java decide to trim, these tests would not fail.

WillChilds-Klein: no, Java 10 is end-of-life and hasn't even taken security patches in years (hence the bug we're seeing) so i don't think its `getEncoded` implementation will change.

if other versions of java do decide to re-introduce the trimming bug, i'd rather our tests fail. if we're made aware of the issue by our tests failing, we can submit a fix upstream to OpenJDK before the affected version is EOL'd and cast into carbonite like JDK10.

amirhosv (approved):

PR #248, opened 2022-10-11
Expose a way to re-run self tests

amirhosv:
## Expose a way to re-run self tests
The result of self tests are cached and subsequent calls to `runSelfTests` would rely on the cached results. In this PR, we allow clients to re-run tests by setting a system property.

### Testing
* CI in personal account
* Unit tests and ensured the new code has coverage by looking into coverage report generated in `build/cmake-coverage/coverage/results/html`

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
+++ b/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
@@ -351,8 +351,7 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
      * The result of running tests are cached, and the subsequent calls would
      * avoid re-running tests. To modify this behaviour, one can set the system
      * property <em>com.amazon.corretto.crypto.provider.cacheselftestresults=false</em>
-     * so that every call to this method would result in re-running tests, or explicity
-     * invoking {@link #reRunSelfTests()}}.
+     * so that every call to this method would result in re-running tests.
      *
      * @see #getSelfTestStatus()
      */
@@ -360,21 +359,6 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
         if (!relyOnCachedSelfTestResults) {
             resetAllSelfTests();
         }
-        return selfTestSuite.runTests();
-    }
-
-    /**
-     * Runs all available self-tests and returns the result.
-     * Please see {@link #getSelfTestStatus()} for the algorithms tested and
-     * the possible return values. (though this method will never return
-     * {@link SelfTestStatus#NOT_RUN}).
-     * This method does not rely on cached test results, and each invocation would
-     * re-running tests.
-     *
-     * @see #getSelfTestStatus()
-     */
-    public SelfTestStatus reRunSelfTests() {
-        resetAllSelfTests();
         return selfTestSuite.runTests();
     }
 
--- a/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
+++ b/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
@@ -39,6 +39,9 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
     public static final String PROVIDER_NAME = "AmazonCorrettoCryptoProvider";
 
     private final EnumSet<ExtraCheck> extraChecks = EnumSet.noneOf(ExtraCheck.class);
+
+    private final boolean relyOnCachedSelfTestResults;
+
     private transient SelfTestSuite selfTestSuite = new SelfTestSuite();
 
     static {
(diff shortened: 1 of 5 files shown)


src/com/amazon/corretto/crypto/provider/Utils.java
```diff
@@ -524,5 +525,14 @@ static int getJavaVersion() {
         }
         return JAVA_VERSION;
     }
+
+    static boolean getCacheSelfTestResultsProperty() {
+        final String relyOnCacheSelfTestResultsStr = Loader.getProperty(CACHE_SELF_TEST_RESULTS, "true").toLowerCase();
+        if (!relyOnCacheSelfTestResultsStr.equals("true") && !relyOnCacheSelfTestResultsStr.equals("false")) {
+            System.err.format("Valid values for %s are false and true, with true as default", CACHE_SELF_TEST_RESULTS);
+            return true;
```
WillChilds-Klein, line 533: why not just rely on parseBoolean for error handling? it seems like we'd want to fail loudly if an invalid property value was specified so the user knows about the issue

tst/com/amazon/corretto/crypto/provider/test/ServiceSelfTestMetaTest.java
```diff
@@ -157,4 +157,13 @@ public void whenMultipleThreadsTest_doesNotDeadlock() throws Throwable {
...
         assertEquals(SelfTestStatus.PASSED, ((SelfTestResult)sneakyInvoke(selfTest, "runTest")).getStatus());
     }
+
+    @Test
+    public void givenACCPCacheSelfTestResultsPropertySetToFalse_whenRunTests_ExpectReset() throws Throwable  {
+        reset();
+        System.setProperty("com.amazon.corretto.crypto.provider.cacheselftestresults", "false");
+        accp = new AmazonCorrettoCryptoProvider();
+        assertTrue(SelfTestStatus.FAILED != accp.runSelfTests());
+        assertTrue(SelfTestStatus.FAILED != accp.reRunSelfTests());
```
WillChilds-Klein, line 167: looks like this test just asserts that the self tests pass, regardless of caching status.

not sure if it's worth the effort, but is there some way we can assert that the tests are actually being re-run? perhaps by doing a `sneakyInvoke` on the `accp` instance to somehow corrupt it in a way that would cause the re-run to fail?

WillChilds-Klein (commented):
woo! congrats on the first PR :)

we should update this section of the README as well:

https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/README.md#other-system-properties

src/com/amazon/corretto/crypto/provider/Utils.java
```diff
@@ -524,5 +525,14 @@ static int getJavaVersion() {
         }
         return JAVA_VERSION;
     }
+
+    static boolean getCacheSelfTestResultsProperty() {
+        final String relyOnCacheSelfTestResultsStr = Loader.getProperty(CACHE_SELF_TEST_RESULTS, "true").toLowerCase();
```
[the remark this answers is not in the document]
amirhosv: This value is cached in the main provider class.

src/com/amazon/corretto/crypto/provider/Utils.java
amirhosv: I just followed the other system variable reading: in case of failure, default to a safe value, which in this case is the existing behavior.

tst/com/amazon/corretto/crypto/provider/test/ServiceSelfTestMetaTest.java
amirhosv: Done

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -344,10 +348,33 @@ public SelfTestStatus getSelfTestStatus() {
...
+     * Runs all available self-tests and returns the result.
+     * Please see {@link #getSelfTestStatus()} for the algorithms tested and
+     * the possible return values. (though this method will never return
+     * {@link SelfTestStatus#NOT_RUN}).
+     * This method does not rely on cached test results, and each invocation would
+     * re-running tests.
+     *
+     * @see #getSelfTestStatus()
+     */
+    public SelfTestStatus reRunSelfTests() {
```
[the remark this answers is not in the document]
amirhosv: I'll remove it.

src/com/amazon/corretto/crypto/provider/Utils.java
WillChilds-Klein: hm, that's a good point. reading over the class again, i agree that [property parsing tends](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/src/com/amazon/corretto/crypto/provider/Utils.java#L482) towards "fall back to [safe default](https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/src/com/amazon/corretto/crypto/provider/Utils.java#L523)" in error cases.

WillChilds-Klein (approved):

PR #258, opened 2023-01-05
Sample code to use artifacts from Maven

amirhosv:
*Description of changes:*

Example project to use ACCP from Maven.

* The example uses Gradle with Kotlin as DSL
* The example itself is also implemented in Kotlin


examples/gradle-kt-dsl/lib/src/test/kotlin/com/amazon/corretto/crypto/examples/AesGcm.kt
```diff
@@ -0,0 +1,39 @@
...
+import java.security.Security
+import javax.crypto.Cipher
+import javax.crypto.spec.GCMParameterSpec
+import javax.crypto.spec.SecretKeySpec
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+fun generateData(len: Int): ByteArray {
+    return ByteArray(len, Int::toByte)
+}
```
SalusaSecondus, line 17: Shouldn't this return random data rather than just all zeros?

amirhosv: It's a good suggestion. This test is just exercising the AES-GCM mode and it's not much concerned with how IVs and keys are generated. I'll add another test that would exercise the LibCryptoRng backed by ACCP, once ACCP 2.0 is released.

examples/gradle-kt-dsl/lib/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
+val accpVersion = "1.6.1"
```
WillChilds-Klein, line 1: should we bump this to 2.0?

is there some way to define this centrall?

examples/gradle-kt-dsl/lib/src/test/kotlin/com/amazon/corretto/crypto/examples/AesGcm.kt
```diff
@@ -0,0 +1,39 @@
...
+
+class AesGcm {
+    @Test fun aesGcmEncryptDecrypt() {
+        Security.insertProviderAt(AmazonCorrettoCryptoProvider(), 1)
```
WillChilds-Klein, line 21: nit: use `install()` method to align with installation/usage recommendations in our README

examples/gradle-kt-dsl/lib/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
...
+val archEnv = System.getenv("ARCH")?: "x86_64"
+//val fips = if (System.getenv("FIPS") == null) "" else "-FIPS"
```
WillChilds-Klein, line 3: should we also pull down + install the FIPS artifact and assert its FIPSness?

examples/gradle-kt-dsl/lib/build.gradle.kts
amirhosv: This version is the one published to Maven, which is not necessarily in sync with the `version`  value in `build.gradle` for develop branch. The commit that we decide to release to Maven should bump up this version.

examples/gradle-kt-dsl/lib/src/test/kotlin/com/amazon/corretto/crypto/examples/AesGcm.kt
amirhosv: Old habits. I'll fix this.

examples/gradle-kt-dsl/lib/build.gradle.kts
amirhosv: This line is commented out because I'm not sure if we will be publishing the FIPS artifacts since AWS-LC does not have its FIPS certificate yet.

examples/gradle-kt-dsl/lib/build.gradle.kts
WillChilds-Klein: i see, that makes sense.

how will we "remember" to come back and bump this? perhaps an MCM validation step to make sure that this version matches the one defined in top-level build.gradle?

amirhosv: That's one of the pre-flight steps to check that the version values are updated.

examples/gradle-kt-dsl/lib/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
+val accpVersion = "1.6.1"
+val archEnv = System.getenv("ARCH")?: "x86_64"
```
WillChilds-Klein, line 2: can we default to system-defined architecture instead?

https://stackoverflow.com/a/29508308

examples/gradle-kt-dsl/run-tests.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+# Retry the build for 30 minutes, every 10 seconds
+NUMBER_OF_RETRIES=180
+DELAY_BETWEEN_RETRIES=10
```
WillChilds-Klein, line 5: what would cause the build to fail? and why 10s between build attempts?

examples/gradle-kt-dsl/run-tests.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+# Retry the build for 30 minutes, every 10 seconds
+NUMBER_OF_RETRIES=180
```
WillChilds-Klein, line 4: if the build time between retries is nonzero, wouldn't this be substantially more than 30m?

examples/gradle-kt-dsl/lib/src/test/kotlin/com/amazon/corretto/crypto/examples/AesGcm.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider
+import java.nio.charset.Charset
+import java.security.SecureRandom
+import javax.crypto.Cipher
+import javax.crypto.spec.GCMParameterSpec
+import javax.crypto.spec.SecretKeySpec
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+fun SecureRandom.randomSeq(len: Int): ByteArray {
```
WillChilds-Klein, line 12: huh, that's kind of neat that kotlin lets you define methods on existing classes in-line like this

examples/gradle-kt-dsl/lib/build.gradle.kts
amirhosv: The issue with "os.arch" is that it's not very clear what the possible values are. So instead I opted for the option of explicitly passing it to the script via an environment variable.

examples/gradle-kt-dsl/run-tests.sh
amirhosv: The build is going to be executed after artifacts are released to Maven. At this point, I don't know how long it would take for artifacts to show up on public repos, so here is a relatively generous retry strategy around the build. Once we have more data, we can tune these parameters.

examples/gradle-kt-dsl/lib/build.gradle.kts
WillChilds-Klein: ok, sounds good

WillChilds-Klein (approved):

PR #274, opened 2023-02-23
Implement AlgorithmParameters for EC

WillChilds-Klein:
*Issue #, if available:* i/CryptoAlg-1413

*Description of changes:*

---

## Implement AlgorithmParameters for EC

This commit registers support for EC AlgorithmParameters with the JCA,
in an effort to get ACCP closer to functioning as a stand-alone
providers. At startup (i.e. `static` initialization time), we query
AWS-LC for which curves are avaialable at startup and cache the relevant
information in order to avoid redundant JNI calls fetching the same,
invariant information.

For some tests to function with correctly with the new
AlgorithmParameters on Java17+, we need to add the sun.security.util
package (internal to JDK, not exported outside of the java.base module)
to the list of specified "opens", allowing for deep reflection.  We
attempt reflection in some cases across java versions because JCE will
sometimes pass `ECKeySizeParameterSpec` instances to
`AlgorithmParameters.init`, but `ECKeySizeParameterSpec` is not a public
class in JDK8 or exported within a module in JDK11+. So, as a "best
effort", we try to handle this using reflection to extract the key size,
which must be explicitly allowed for JDK-internal classes/modules in
Java17+. It's not an ideal, but is an unfortunate reality of the JDK's
module configuration and interface limitations.

I confirmed that the new EcParameters class has full coverage, save for
a few exception-throwing cases that are difficult to trigger due to
current inability to mock our JNI/AWS-LC layer.

---

## Make ACCP-vended EC AlgorithmParameters opt-in

This commit makes ACCP-provided EC paramter not only optional, but
opt-in. The reason for this is somewhat complex, but is fundamentally
due to later JDK versions' addition of the module system and default
tightening of access to JDK-internal classes via reflection.

In testing with JDK17, I started to notice this odd exception:

```
Exception in thread "main" java.security.NoSuchAlgorithmException: Error constructing KeyPairGenerator for EC using SunEC
        at jdk.crypto.ec/sun.security.ec.SunEC$ProviderService.newInstance(SunEC.java:181)
        at java.base/sun.security.jca.GetInstance.getInstance(GetInstance.java:236)
        at java.base/sun.security.jca.GetInstance.getInstance(GetInstance.java:206)
        at java.base/java.security.KeyPairGenerator.getInstance(KeyPairGenerator.java:300)
        at com.amazon.corretto.crypto.provider.test.SecurityPropertyTester.main(SecurityPropertyTester.java:43)
Caused by: java.security.InvalidParameterException: No EC parameters available for key size 256 bits
        at jdk.crypto.ec/sun.security.ec.ECKeyPairGenerator.initialize(ECKeyPairGenerator.java:81)
        at jdk.crypto.ec/sun.security.ec.ECKeyPairGenerator.<init>(ECKeyPairGenerator.java:71)
        at jdk.crypto.ec/sun.security.ec.SunEC$ProviderService.newInstance(SunEC.java:155)
        ... 4 more
```

The issue, as alluded to in the prior commit message, is that SunEC will
pass a `ECKeySizeParameterSpec` instance into `AlgorithmParameters.init`
and thus our EC parameters implementation needs to handle that class.
Unfortunately, `ECKeySizeParameterSpec` is a JDK-internal class and is
not exported by its module configuration. We work around this for JDK11
by using reflection to access the key size parameter, but JDK17
restricts the relevant reflection permissions by default. See [these
docs][1].

The upshot for us is that in order for ACCP to play nicely with JDK17,
users would need to include this flag to their JDK invocations:

```
--add-opens java.base/sun.security.util=ALL-UNNAMED
```

This permissions expansion to application code may be either onerous or
impermissible for customers, so we're making the EC parameters feature
opt-in for now. It should only be used by customers wishing to use ACCP
as a standalone provider that doesn't need to interface with
JCE/Java-internal cryptography.

[1]: https://docs.oracle.com/en/java/javase/18/migrate/migrating-jdk-8-later-jdk-releases.html#GUID-12F945EB-71D6-46AF-8C3D-D354FD0B1781

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -589,6 +589,7 @@ if (TEST_JAVA_MAJOR_VERSION VERSION_GREATER_EQUAL 17)
     --add-opens java.base/javax.crypto=ALL-UNNAMED
     --add-opens java.base/java.lang.invoke=ALL-UNNAMED
     --add-opens java.base/java.security=ALL-UNNAMED
+    --add-opens=java.base/sun.security.util=ALL-UNNAMED
     --add-exports jdk.crypto.ec/sun.security.ec=java.base
     )
 endif()
@@ -693,6 +694,7 @@ add_custom_target(check-install-via-properties
     COMMAND ${TEST_JAVA_EXECUTABLE}
         ${COVERAGE_ARGUMENTS}
         ${TEST_FIPS_PROPERTY}
+        ${TEST_ADD_OPENS}
         -cp $<TARGET_PROPERTY:accp-jar,JAR_FILE>:$<TARGET_PROPERTY:tests-jar,JAR_FILE>:${TEST_CLASSPATH}
         ${EXTERNAL_LIB_PROPERTY}
         -Dcom.amazon.corretto.crypto.provider.inTestSuite=hunter2
@@ -709,6 +711,7 @@ add_custom_target(check-external-lib
     COMMAND ${TEST_JAVA_EXECUTABLE}
         ${COVERAGE_ARGUMENTS}
         ${TEST_FIPS_PROPERTY}
+        ${TEST_ADD_OPENS}
         -cp $<TARGET_PROPERTY:accp-jar,JAR_FILE>:$<TARGET_PROPERTY:tests-jar,JAR_FILE>:${TEST_CLASSPATH}
         # Since this tests external loading we always provide this property
         -Djava.library.path=$<TARGET_FILE_DIR:amazonCorrettoCryptoProvider>
@@ -725,6 +728,7 @@ add_custom_target(check-install-via-properties-recursive
     COMMAND ${TEST_JAVA_EXECUTABLE}
         ${COVERAGE_ARGUMENTS}
         ${TEST_FIPS_PROPERTY}
+        ${TEST_ADD_OPENS}
         -cp $<TARGET_PROPERTY:accp-jar,JAR_FILE>:$<TARGET_PROPERTY:tests-jar,JAR_FILE>:${TEST_CLASSPATH}
         ${EXTERNAL_LIB_PROPERTY}
         -Dcom.amazon.corretto.crypto.provider.inTestSuite=hunter2
@@ -739,6 +743,7 @@ add_custom_target(check-install-via-properties-with-debug
     COMMAND ${TEST_JAVA_EXECUTABLE}
         ${COVERAGE_ARGUMENTS}
         ${TEST_FIPS_PROPERTY}
+        ${TEST_ADD_OPENS}
         -cp $<TARGET_PROPERTY:accp-jar,JAR_FILE>:$<TARGET_PROPERTY:tests-jar,JAR_FILE>:${TEST_CLASSPATH}
         ${EXTERNAL_LIB_PROPERTY}
         -Dcom.amazon.corretto.crypto.provider.inTestSuite=hunter2
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -589,7 +589,7 @@ if (TEST_JAVA_MAJOR_VERSION VERSION_GREATER_EQUAL 17)
     --add-opens java.base/javax.crypto=ALL-UNNAMED
     --add-opens java.base/java.lang.invoke=ALL-UNNAMED
     --add-opens java.base/java.security=ALL-UNNAMED
-    --add-opens=java.base/sun.security.util=ALL-UNNAMED
+    --add-opens java.base/sun.security.util=ALL-UNNAMED
     --add-exports jdk.crypto.ec/sun.security.ec=java.base
     )
 endif()
@@ -694,7 +694,6 @@ add_custom_target(check-install-via-properties
     COMMAND ${TEST_JAVA_EXECUTABLE}
         ${COVERAGE_ARGUMENTS}
         ${TEST_FIPS_PROPERTY}
-        ${TEST_ADD_OPENS}
         -cp $<TARGET_PROPERTY:accp-jar,JAR_FILE>:$<TARGET_PROPERTY:tests-jar,JAR_FILE>:${TEST_CLASSPATH}
         ${EXTERNAL_LIB_PROPERTY}
         -Dcom.amazon.corretto.crypto.provider.inTestSuite=hunter2
@@ -711,7 +710,6 @@ add_custom_target(check-external-lib
     COMMAND ${TEST_JAVA_EXECUTABLE}
         ${COVERAGE_ARGUMENTS}
         ${TEST_FIPS_PROPERTY}
-        ${TEST_ADD_OPENS}
         -cp $<TARGET_PROPERTY:accp-jar,JAR_FILE>:$<TARGET_PROPERTY:tests-jar,JAR_FILE>:${TEST_CLASSPATH}
         # Since this tests external loading we always provide this property
         -Djava.library.path=$<TARGET_FILE_DIR:amazonCorrettoCryptoProvider>
@@ -728,7 +726,6 @@ add_custom_target(check-install-via-properties-recursive
     COMMAND ${TEST_JAVA_EXECUTABLE}
         ${COVERAGE_ARGUMENTS}
         ${TEST_FIPS_PROPERTY}
-        ${TEST_ADD_OPENS}
         -cp $<TARGET_PROPERTY:accp-jar,JAR_FILE>:$<TARGET_PROPERTY:tests-jar,JAR_FILE>:${TEST_CLASSPATH}
         ${EXTERNAL_LIB_PROPERTY}
         -Dcom.amazon.corretto.crypto.provider.inTestSuite=hunter2
@@ -743,7 +740,6 @@ add_custom_target(check-install-via-properties-with-debug
     COMMAND ${TEST_JAVA_EXECUTABLE}
         ${COVERAGE_ARGUMENTS}
         ${TEST_FIPS_PROPERTY}
-        ${TEST_ADD_OPENS}
         -cp $<TARGET_PROPERTY:accp-jar,JAR_FILE>:$<TARGET_PROPERTY:tests-jar,JAR_FILE>:${TEST_CLASSPATH}
         ${EXTERNAL_LIB_PROPERTY}
         -Dcom.amazon.corretto.crypto.provider.inTestSuite=hunter2
@@ -790,6 +786,8 @@ add_custom_target(coverage
 add_custom_target(check-integration-extra-checks
     COMMAND ${TEST_JAVA_EXECUTABLE}
         -Dcom.amazon.corretto.crypto.provider.extrachecks=ALL
+        # perform standard integration tests with JCE EC parameters, and extra-checks integration tests with ACCP's
+        -Dcom.amazon.corretto.crypto.provider.registerEcParams=true
         ${TEST_RUNNER_ARGUMENTS}
         --select-package=com.amazon.corretto.crypto.provider.test.integration
 
--- a/csrc/ec_utils.cpp
+++ b/csrc/ec_utils.cpp
@@ -5,6 +5,7 @@
 #include <openssl/ec.h>
 #include <openssl/err.h>
 #include <openssl/objects.h>
+#include <vector>
 #include "generated-headers.h"
 #include "util.h"
 #include "env.h"
@@ -63,7 +64,9 @@ JNIEXPORT jint JNICALL Java_com_amazon_corretto_crypto_provider_EcUtils_curveNam
   jbyteArray cofactorArr,
   jbyteArray gxArr,
   jbyteArray gyArr,
-  jbyteArray orderArr)
+  jbyteArray orderArr,
+  jbyteArray oid,
+  jbyteArray encoded)
 {
     try {
         raii_env env(pEnv);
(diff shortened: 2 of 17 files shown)


src/com/amazon/corretto/crypto/provider/EcParameters.java
```diff
@@ -0,0 +1,115 @@
...
+            // reference this class directly using instanceof up until Java 8, but in Java 9 it was moved to the
+            // java.base module and only exported to other JDK-internal modules. To work around this, we use reflection
+            // to get the key size and get the corresponding curve name from our own database. Future versions of java
+            // may restrict reflective access to private modules, so this functionality may break.
+            //
+            // https://github.com/corretto/corretto-11/blob/14c02261590b4dc01284888a7a51d39ff581ac8d/src/java.base/share/classes/sun/security/util/ECParameters.java#L121
+            // https://github.com/corretto/corretto-11/blob/14c02261590b4dc01284888a7a51d39ff581ac8d/src/java.base/share/classes/module-info.java#L301-L314
+            try {
+                Method getKeySize = paramSpec.getClass().getMethod("getKeySize");
+                Integer keySize = Integer.class.cast(getKeySize.invoke(paramSpec));
```
geedo0, line 49: Standard reflection due diligence - We assume this interface is rock solid at this point since this is only necessary for older JDKs?

src/com/amazon/corretto/crypto/provider/EcParameters.java
```diff
@@ -0,0 +1,115 @@
...
+                name = EcUtils.getNameByKeySize(keySize);
+            } catch (ReflectiveOperationException e) {
+                // pass, perhaps due to reflective access control restrictions. rely null check below to throw.
+            }
+        }
+
+        if (name == null) {
+            throw new InvalidParameterSpecException("Only ECParameterSpec and ECGenParameterSpec supported");
```
geedo0, line 57: Error message is potentially misleading. We can also get a null name by the EcUtils.get*() queries failing. Parameterizing the error message would be ideal, but I can see how that's tricky due to scoping here.

src/com/amazon/corretto/crypto/provider/EcParameters.java
```diff
@@ -0,0 +1,115 @@
...
+        return ecInfo.encoded.clone();  // clone to avoid exposing static reference
+    }
+
+    protected byte[] engineGetEncoded(String encodingMethod) throws IOException {
+        return engineGetEncoded();
+    }
+
+    protected String engineToString() {
+        if (ecInfo == null) {
+            return "Not initialized";
```
geedo0, line 110: Throw a RuntimeException instead? This seems dodgy.

src/com/amazon/corretto/crypto/provider/EcUtils.java
```diff
@@ -108,12 +112,27 @@ private String normalizeName(final String name) {
...
         }
     };
 
+    private static final ConcurrentHashMap<ByteBuffer, String> EC_NAME_BY_ENCODED = new ConcurrentHashMap<>();
+    private static final ConcurrentHashMap<EllipticCurve, String> EC_NAME_BY_CURVE = new ConcurrentHashMap<>();
+    private static final ConcurrentHashMap<Integer, String> EC_NAME_BY_KEY_SIZE = new ConcurrentHashMap<>();
+    static {
+        for (String name : getCurveNames()) {
+            getSpecByName(name);                    // warm cache keyed by name, must come before warming by OID
+            getSpecByName(getOidFromName(name));    // warm cache keyed by OID
```
geedo0, line 129: Comments could be a bit clearer. Is this just a performance optimization for the static initializer and has no impact after the class is loaded?

tst/com/amazon/corretto/crypto/provider/test/EcParametersTest.java
```diff
@@ -0,0 +1,160 @@
...
+@ResourceLock(value = TestUtil.RESOURCE_GLOBAL, mode = ResourceAccessMode.READ)
+public class EcParametersTest {
+
+    private static String[][] legacyCurveParams() {
+        return TestUtil.LEGACY_CURVES;
+    }
+
+    @ParameterizedTest
+    @EnabledForJreRange(min=JRE.JAVA_8, max=JRE.JAVA_14)
+    @MethodSource("legacyCurveParams")
```
geedo0, line 45: Cool use of parameterization and that JRE range decorator.

src/com/amazon/corretto/crypto/provider/EcUtils.java
```diff
@@ -127,9 +146,22 @@ private static native int curveNameToInfo(String curveName,
...
     private static native long buildGroup(int nid);
     private static native void freeGroup(long ptr);
+    private static native String[] getCurveNames();
+    private static native String getCurveNameFromEncoded(byte[] encoded);
+
+    static String getOidFromName(String name) {
+        if (name == null) {
+            return null;
+        }
+        return getSpecByName(name) == null ? null : getSpecByName(name).oid;
```
amirhosv, line 163: Maybe first store `getSpecByName` to avoid invoking it again when it's not null.

src/com/amazon/corretto/crypto/provider/EcParameters.java
```diff
@@ -0,0 +1,115 @@
...
+            throw new IOException("Only named EcParameters supported");
+        }
+        ecInfo = EcUtils.getSpecByName(name);
+        if (ecInfo == null) {
+            throw new IOException("Unknown named curve: " + name);
+        }
+    }
+
+    protected void engineInit(byte[] params, String unused) throws IOException {
+        engineInit(params);
```
amirhosv, line 83: Is this the desired behavior? Alternative is to throw something like `UnsupportedOperationException`. Just checking.

CMakeLists.txt
```diff
@@ -589,6 +589,7 @@ if (TEST_JAVA_MAJOR_VERSION VERSION_GREATER_EQUAL 17)
     --add-opens java.base/javax.crypto=ALL-UNNAMED
     --add-opens java.base/java.lang.invoke=ALL-UNNAMED
     --add-opens java.base/java.security=ALL-UNNAMED
+    --add-opens=java.base/sun.security.util=ALL-UNNAMED
```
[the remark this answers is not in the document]
WillChilds-Klein: sure.

src/com/amazon/corretto/crypto/provider/EcParameters.java
```diff
@@ -0,0 +1,115 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.amazon.corretto.crypto.provider;
+
+import java.io.IOException;
+import java.lang.reflect.Method;
+import java.security.*;
+import java.security.spec.*;
```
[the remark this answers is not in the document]
WillChilds-Klein: hm, no. good catch.

src/com/amazon/corretto/crypto/provider/EcUtils.java
WillChilds-Klein: > Is this just a performance optimization for the static initializer and has no impact after the class is 

it's a performance enhancement for runtime post-initialization. we pay the (small) JNI cost a few times during static initialization, then every cache hit after that is just a HashMap get(). this is safe because the set of supported curves is invariant after loading AWS-LC.

i'll expand the comments a bit.

tst/com/amazon/corretto/crypto/provider/test/EcParametersTest.java
WillChilds-Klein: yeah! i cribbed this from [EcGenTest](https://github.com/corretto/amazon-corretto-crypto-provider/blob/3cddd78c92f72bce13eb78f96a18f1a2ec63f369/tst/com/amazon/corretto/crypto/provider/test/EcGenTest.java#L108). it's a nice JUnit feature.

src/com/amazon/corretto/crypto/provider/EcUtils.java
WillChilds-Klein: yeah might as well save a method invocation. will fix.

src/com/amazon/corretto/crypto/provider/EcParameters.java
WillChilds-Klein: this is intentional, and is intended to preserve behavioral compatibility with [SunJCE's implementation](https://github.com/corretto/corretto-11/blob/develop/src/java.base/share/classes/sun/security/util/ECParameters.java#L191).

src/com/amazon/corretto/crypto/provider/EcParameters.java
WillChilds-Klein: this is intentional, and is intended to preserve behavioral compatibility with [SunJCE's implementation](https://github.com/corretto/corretto-11/blob/develop/src/java.base/share/classes/sun/security/util/ECParameters.java#L191).

src/com/amazon/corretto/crypto/provider/EcParameters.java
WillChilds-Klein: Because ECKeySizeParameterSpec is a JDK-internal class, I consider the reflective check we do to be an implementation detail and thus intentionally don't want to expose it to callers. the only other two classes we handle are ECParameterSpec and ECGenParameterSpec (as articulated by this message). If any other class is specified, `name` will be null at the end of this conditional block and the error message will be thrown.

src/com/amazon/corretto/crypto/provider/EcParameters.java
WillChilds-Klein: very fair. in my testing, it's actually necessary for all JDKs tested under integration tests (8, 11, 17). the difference with 8 is that we _could_ import it there and test with `instanceof` instead of using reflection. in later versions, reflection is necessary due to modularization in JDK9, but not always possible. it's not possible unless `--add-opens=java.base/sun.security.util=ALL-UNNAMED` or similar is specified in JVM invocation.

this reflective call is designed to try the reflective check, and if it is not possible to fall back "cleanly" to the exception case on L57.

ECKeySizeParameterSpec is a JDK-internal class, so i wouldn't really call it an "interface", but the goal here is to fail in a "clean" way when the required permissions aren't available.

all this said, i think we need a test case for the negative case when reflective permissions aren't allowed. i'll work on adding that.

csrc/ec_utils.cpp
```diff
@@ -155,9 +158,101 @@ JNIEXPORT jint JNICALL Java_com_amazon_corretto_crypto_provider_EcUtils_curveNam
         }
         orderBN.toJavaArray(env, orderArr);
 
+        // Get the decoded (string) curve OID
+        jni_borrow oidBorrow = jni_borrow(env, java_buffer::from_array(env, oid), "curveNameToInfo");
+        if (!OBJ_obj2txt((char*) oidBorrow.data(), (int) oidBorrow.len(), OBJ_nid2obj(nid), /*always_return_oid*/1)) {
+            throw_openssl("Unable to get decoded curve OID");
+        }
+        oidBorrow.release();
```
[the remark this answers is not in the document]
WillChilds-Klein: yes, while `release` is [called by the dtor](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/csrc/buffer.h#L349) when a `jni_borrow` goes out of scope, we need to explicitly call it here before taking out the next borrow to avoid a segfault. if i don't manually release the first `jni_borrow`, i get this:

```
Attempted to use JNI environment while buffer locks were outstanding:
        0x7f85ddeb9df0: "curveNameToInfo"
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f85dd71e4f0, pid=27432, tid=27567
#
# JRE version: OpenJDK Runtime Environment Corretto-11.0.18.10.1 (11.0.18+10) (build 11.0.18+10-LTS)
# Java VM: OpenJDK 64-Bit Server VM Corretto-11.0.18.10.1 (11.0.18+10-LTS, mixed mode, tiered, compressed oops, g1 gc, linux-amd64)
# Problematic frame:
# C  [libamazonCorrettoCryptoProvider.so+0x1b4f0]  Java_com_amazon_corretto_crypto_provider_EcUtils_curveNameToInfo+0x4f0
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /local/home/childw/workplace/github/corretto/amazon-corretto-crypto-provider/build/cmake/hs_err_pid27432.log
End trace (1 locks)
```

why do we get a segfault if we fail to manually release the first `jni_borrow` before taking out the second? i'm not 100% sure, as I found the manual release workaround and kept going. but digging into it a little, i suspect that it has to do with how we manage our own locks on top of the JNI functions or with [our usage of GetPrimitiveArrayCritical](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/csrc/buffer.h#L333). to my mind, it still shouldn't be an issue as the arrays don't overlap (why would we need to lock?), and the C++ compiler should call the dtors in the correct order when they go out of scope (LIFO -- first `encodedBorrow`, then `oidBorrow`). if we do need to take out multiple `jni_borrow`s at a time in the same scope, we'll need to investigate this further.

i'm using `jni_borrow` here for performance reasons -- we want to write directly to the java heap if we can to avoid copies.

the `release` call on 175 is unnecessary; i included it for consistency.

amirhosv (approved):

PR #279, opened 2023-03-13
Service class for KeyGenerator

amirhosv:
*Issue:* 

* CryptoAlg-1412

*Description of changes:*

* Service class for KeyGenerator
* The implementation is very similar to [SunJCE](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/com/sun/crypto/provider/AESKeyGenerator.java).
* Support AES key generation
* Improve error handling in case creation of tmp dir fails

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
+++ b/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
@@ -3,12 +3,9 @@
 
 package com.amazon.corretto.crypto.provider;
 
-import static com.amazon.corretto.crypto.provider.Loader.PROVIDER_VERSION;
-import static com.amazon.corretto.crypto.provider.Loader.PROVIDER_VERSION_STR;
-import static java.lang.String.format;
-import static java.util.Arrays.asList;
-import static java.util.Collections.singletonMap;
-import static java.util.logging.Logger.getLogger;
+import com.amazon.corretto.crypto.provider.keygeneratorspi.AesSecretKeyProperties;
+import com.amazon.corretto.crypto.provider.keygeneratorspi.DefaultSecureRandomSupplier;
+import com.amazon.corretto.crypto.provider.keygeneratorspi.SecretKeyGenerator;
 
 import java.io.IOException;
 import java.lang.invoke.MethodHandle;
@@ -27,6 +24,13 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ForkJoinPool;
 import java.util.function.Supplier;
+
+import static com.amazon.corretto.crypto.provider.Loader.PROVIDER_VERSION;
+import static com.amazon.corretto.crypto.provider.Loader.PROVIDER_VERSION_STR;
+import static java.lang.String.format;
+import static java.util.Arrays.asList;
+import static java.util.Collections.singletonMap;
+import static java.util.logging.Logger.getLogger;
 
 
 public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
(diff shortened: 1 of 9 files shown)


src/com/amazon/corretto/crypto/provider/LibCryptoRng.java
```diff
@@ -10,7 +10,7 @@
  * A simple wrapper around the linked LibCrypto's RAND_bytes() API.
  */
 public class LibCryptoRng extends SecureRandom {
-    private static final String ALGORITHM_NAME = "LibCryptoRng";
+    public static final String ALGORITHM_NAME = "LibCryptoRng";
```
geedo0, line 13: Any reason we can't use the public `getAlgorithm()` mathod? Or do we need the static accessor here?

src/com/amazon/corretto/crypto/provider/keygeneratorspi/SecretKeyGenerator.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.crypto.spec.SecretKeySpec;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.SecureRandom;
+import java.security.spec.AlgorithmParameterSpec;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+public class SecretKeyGenerator extends KeyGeneratorSpi {
```
geedo0, line 16: What's Spi?

src/com/amazon/corretto/crypto/provider/keygeneratorspi/SecretKeyGenerator.java
```diff
@@ -0,0 +1,55 @@
...
+    private final Supplier<SecureRandom> defaultSecureRandom;
+    private final SecretKeyProperties secretKeyProperties;
+    private Optional<SecureRandom> secureRandom;
+    private int keySize;
+
+    public SecretKeyGenerator(final Supplier<SecureRandom> defaultSecureRandom, final SecretKeyProperties secretKeyProperties) {
+        this.defaultSecureRandom = Objects.requireNonNull(defaultSecureRandom, "defaultSecureRandom cannot be null.");
```
geedo0, line 23: There's a degenerate case here where the defaultSecureRandom supplier supplies null objects. Suggest adding a safety check either here, or in `engineGenerateKey` (probably the better place) that asserts that the defaultSecureRandom we get is non-null.

src/com/amazon/corretto/crypto/provider/keygeneratorspi/SecretKeyGenerator.java
```diff
@@ -0,0 +1,55 @@
...
+        secretKeyProperties.checkKeySizeIsValid(keySize);
+        this.keySize = keySize;
+        this.secureRandom = Optional.ofNullable(random);
+    }
+
+    @Override
+    protected SecretKey engineGenerateKey() {
+        final byte[] keyBytes = new byte[keySize / 8];
+        final SecureRandom srand = secureRandom.orElseGet(defaultSecureRandom);
+        srand.nextBytes(keyBytes);
```
geedo0, line 50: This can throw an NPE if defaultSecureRandom is null.

src/com/amazon/corretto/crypto/provider/LibCryptoRng.java
amirhosv: `getAlgorithm()` is not static, as you pointed out.

src/com/amazon/corretto/crypto/provider/keygeneratorspi/SecretKeyGenerator.java
amirhosv: It's interesting that `orElseGet` does not check the nullness of return value of the supplier.

src/com/amazon/corretto/crypto/provider/keygeneratorspi/SecretKeyGenerator.java
amirhosv: I think the `Objects.requireNonNull`s are redundant and I should remove them. This class is instantiated  by ACCP and client code has no control over it. We should statically ensure that the arguments we provide to this constructor are not null rather than adding run-time checks. The same argument goes for the return value of `defaultSecureRandom`, since we provide its implementation.

src/com/amazon/corretto/crypto/provider/keygeneratorspi/SecretKeyGenerator.java
amirhosv: Service Provider Interface: in JCA, classes used by client code, such as `KeyGenerator`, are called engine classes. [Engine classes delegate the operation to SPI classes that are implemented by a provider](https://docs.oracle.com/en/java/javase/11/security/howtoimplaprovider.html#GUID-B35E3749-0A12-42C3-BA0B-C444D7E140BB).

src/com/amazon/corretto/crypto/provider/keygeneratorspi/SecretKeyGenerator.java
amirhosv: We should start using something like [CheckerFramework](https://checkerframework.org/) to assert that values are not null and there would be no need to add nullness check.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -137,21 +153,30 @@ private class ACCPService extends Service {
 
         public ACCPService(
                 final String type, final String algorithm, final String className,
+                final boolean useReflection, // this flag determines if the instantiation of the service class is supposed to be done via reflection or explicitly
```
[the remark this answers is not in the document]
amirhosv: In general, it's better not to use reflection due to its performance overhead. Another reason to not use reflection is our `addService` method assumes the constructor of the SPI class does not take any argument; in case an SPI's constructor has arguments, it's best to explicitly instantiate it with the right arguments.

src/com/amazon/corretto/crypto/provider/LibCryptoRng.java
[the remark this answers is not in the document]
amirhosv: I'll fix this.

PR #283, opened 2023-03-24
Include AWS-LC's self tests in ACCP's self tests

amirhosv:
*Issue #, if available:*

*Description of changes:*
* Include AWS-LC's self tests in ACCP's self tests.
* Modify `assertHealthy` to re-run tests in case it should not rely on cached results
* Perform AWS-LC's KAT FIPS tests as part of `isFips` API too.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
+++ b/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
@@ -336,12 +336,15 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
             return;
         }
 
+        // The order of adding determines the order of failure reporting and test execution. We do not short circuit
+        // the execution of tests when a test fails.
+        selfTestSuite.addSelfTest(SelfTestSuite.AWS_LC_SELF_TESTS);
+        selfTestSuite.addSelfTest(LibCryptoRng.SPI.SELF_TEST);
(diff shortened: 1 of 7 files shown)


src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -342,6 +342,7 @@ private synchronized void initializeSelfTests() {
         selfTestSuite.addSelfTest(EvpHmac.SHA1.SELF_TEST);
         selfTestSuite.addSelfTest(EvpHmac.MD5.SELF_TEST);
         selfTestSuite.addSelfTest(LibCryptoRng.SPI.SELF_TEST);
+        selfTestSuite.addSelfTest(SelfTestSuite.AWS_LC_SELF_TESTS);
```
alexw91, line 345: This is all hypothetical, but shouldn't we run AWS-LC's self tests first before ACCP's own internal self tests? 

If an algorithm in AWS-LC will fail AWS-LC's self tests (and presumably our ACCP self tests as well), I'd rather that an ACCP user see the AWS-LC self-test failure first so that they know where to start debugging.

amirhosv: That makes sense. The self-tests are stored in a HashMap and the order they get inserted does not determine the order they get executed and reported. I can change the underlying data structure to ensure the order of execution matches the order of insertion.

alexw91 (approved):

PR #285, opened 2023-04-14
Provide isFipsStatusOk API to check if FIPS tests have passed

amirhosv:
*Description of changes:*

* `AWS_LC_fips_failure_callback` is a method invoked by AWS-LC in case a FIPS related check fails
* ACCP updates a global flag to false in case this callback is invoked
* The value of the global flag is returned as the result of `isFipsStatusOk`
* ACCP built in FIPS mode checks `isFipsStatusOk` whenever a crypto service is requested

*Testing*

*  Used break-tests from examples, and did the following:
  * `./gradlew clean -DALLOW_FIPS_TEST_BREAK=true -DFIPS=true build`
  * `cd examples/break-tests`
  *  ` BORINGSSL_FIPS_BREAK_TEST=CRNG ./gradlew run`
  * The output shows that AWS-LC invokes the provided callback in case a FIPS related check fails.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -264,6 +264,7 @@ add_library(
         csrc/util.cpp
         csrc/util_class.cpp
         csrc/fips_kat_self_test.cpp
+        csrc/fips_status.cpp
         ${JNI_HEADER_DIR}/generated-headers.h
 )
 
@@ -462,7 +463,7 @@ if(NOT ENABLE_NATIVE_TEST_HOOKS)
     CHECK_LINKER_FLAG_SUPPORT(USE_VERSION_SCRIPT "-Wl,--version-script -Wl,${CMAKE_CURRENT_SOURCE_DIR}/final-link.version")
 
 # This does the same thing as the version script, but works on Darwin platforms
-    CHECK_LINKER_FLAG_SUPPORT(USE_EXPORTED_SYMBOL "-Wl,-exported_symbol '-Wl,_Java_*' -Wl,-exported_symbol '-Wl,_JNI_*'")
+    CHECK_LINKER_FLAG_SUPPORT(USE_EXPORTED_SYMBOL "-Wl,-exported_symbol '-Wl,_Java_*' '-Wl,_AWS_LC_fips_failure_callback' -Wl,-exported_symbol '-Wl,_JNI_*'")
 endif()
 
 # Attempt to drop unused sections; the idea here is to exclude unreferenced
@@ -640,6 +641,7 @@ add_custom_target(check-junit
         --select-package=com.amazon.corretto.crypto.provider.test
         --exclude-package=com.amazon.corretto.crypto.provider.test.integration
         --exclude-classname=com.amazon.corretto.crypto.provider.test.SecurityManagerTest
+        --exclude-classname=com.amazon.corretto.crypto.provider.test.FipsStatusTest
 
     DEPENDS accp-jar tests-jar)
 
@@ -661,6 +663,15 @@ add_custom_target(check-junit-SecurityManager
 
     DEPENDS accp-jar tests-jar)
 
+add_custom_target(check-junit-FipsStatus
+    COMMAND ${TEST_JAVA_EXECUTABLE}
+        ${TEST_RUNNER_ARGUMENTS}
+        --select-class=com.amazon.corretto.crypto.provider.test.AesTest # Force loading ciphers
+        --select-class=com.amazon.corretto.crypto.provider.test.SHA1Test # Force loading digests
+        --select-class=com.amazon.corretto.crypto.provider.test.FipsStatusTest
+
+    DEPENDS accp-jar tests-jar)
+
 add_custom_target(check-junit-extra-checks
     COMMAND ${TEST_JAVA_EXECUTABLE}
         -Dcom.amazon.corretto.crypto.provider.extrachecks=ALL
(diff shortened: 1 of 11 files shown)


CMakeLists.txt
```diff
@@ -462,7 +463,7 @@ if(NOT ENABLE_NATIVE_TEST_HOOKS)
     CHECK_LINKER_FLAG_SUPPORT(USE_VERSION_SCRIPT "-Wl,--version-script -Wl,${CMAKE_CURRENT_SOURCE_DIR}/final-link.version")
 
 # This does the same thing as the version script, but works on Darwin platforms
-    CHECK_LINKER_FLAG_SUPPORT(USE_EXPORTED_SYMBOL "-Wl,-exported_symbol '-Wl,_Java_*' -Wl,-exported_symbol '-Wl,_JNI_*'")
+    CHECK_LINKER_FLAG_SUPPORT(USE_EXPORTED_SYMBOL "-Wl,-exported_symbol '-Wl,_Java_*' '-Wl,_AWS_LC_fips_*' -Wl,-exported_symbol '-Wl,_JNI_*'")
```
WillChilds-Klein, line 466: do we need to wildcard this, or can we specifically stipulate `AWS_LC_fips_failure_callback`?

examples/break-tests/app/src/main/java/com/amazon/accp/breaktests/App.java
```diff
@@ -28,11 +29,12 @@ private static void breakingTestsWithAwsLcHooks() throws Exception {
             System.out.println("BORINGSSL_FIPS_BREAK_TEST is set to " + boringsslFipsBreakTestEnvVar);
         } else {
             System.out.println("BORINGSSL_FIPS_BREAK_TEST is not defined");
+            return;
         }
 
         System.out.println("Checking if ACCP is installed properly. If not, we'll get an exception ...");
-        p.assertHealthy();
-        System.out.println("Checking if ACCP FIPS is good ... " + p.isFips());
+        //p.assertHealthy();
```
WillChilds-Klein, line 36: does this assertion still hold?

tst/com/amazon/corretto/crypto/provider/test/FipsStatusTest.java
```diff
@@ -0,0 +1,41 @@
...
+@ExtendWith(TestResultLogger.class)
+@Execution(ExecutionMode.CONCURRENT)
+@ResourceLock(value = TestUtil.RESOURCE_GLOBAL, mode = ResourceAccessMode.READ)
+public class FipsStatusTest {
+
+    private final AmazonCorrettoCryptoProvider provider = AmazonCorrettoCryptoProvider.INSTANCE;
+
+    @Test
+    public void givenAccpBuiltWithFips_whenAWS_LC_fips_failure_callback_expectException() throws Exception {
+        if (provider.isFips()) {
```
WillChilds-Klein, line 29: what's the expected behavior when `isFipsStatusOk()` is called on a non-FIPS build?


is there anything we should test in the non-FIPS case, if only to codify that expected behavior?

tst/com/amazon/corretto/crypto/provider/test/FipsStatusTest.java
```diff
@@ -0,0 +1,41 @@
...
+            assertTrue(provider.isFipsStatusOk());
+            assertNotNull(KeyGenerator.getInstance("AES", provider));
+            // call the failure callback
+            NativeTestHooks.callAwsLcFipsFailureCallback();
+            assertFalse(provider.isFipsStatusOk());
+            // we should not be able to get any service object
+            assertThrows(FipsStatusException.class, () -> KeyGenerator.getInstance("AES", provider));
+            // we need to flip the status back to OK so the rest of tests would work
+            NativeTestHooks.flipFipsStatus();
```
WillChilds-Klein, line 38: perhaps add one more `assertNotNull(KeyGenerator.getInstance("AES", provider));` on L39 to ensure we can get instances now?

csrc/atomic_bool.h
```diff
@@ -0,0 +1,90 @@
...
+            exit(EXIT_FAILURE);
+        }
+    }
+
+    ~AtomicBool() {
+        pthread_mutex_destroy(&value_mutex_);
+    }
+
+    bool load() {
+        UniquePthreadMutexLock lock(&value_mutex_);
```
WillChilds-Klein, line 67: do we need a read lock here? changing a bool is just a bit flip, so partial updates should be impossible.

csrc/atomic_bool.h
```diff
@@ -0,0 +1,90 @@
...
+public:
+
+    UniquePthreadMutexLock(pthread_mutex_t* mutex): mutex_(mutex) {
+        int status = pthread_mutex_lock(mutex_);
+        if (status != 0) {
+            fprintf(stderr, "pthread_mutex_lock failed with error %d", status);
+        }
+    }
+
+    ~UniquePthreadMutexLock() {
```
WillChilds-Klein, line 39: nice. using the dtor to unlock gives a really nice usage pattern

csrc/atomic_bool.h
amirhosv: I researched this and apparently C++ does not provide any atomic read guarantees.

tst/com/amazon/corretto/crypto/provider/test/FipsStatusTest.java
amirhosv: Once the flag is false, it remains false. I added a new target in our CMake for this test so we can run it independent of the others and not to worry about flipping the flag back for the sake of the other tests.

tst/com/amazon/corretto/crypto/provider/test/FipsStatusTest.java
amirhosv: For non-FIPs builds, it always returns true since AWS-LC doesn't do callback registration. We also don't check for instantiate service classes

examples/break-tests/app/src/main/java/com/amazon/accp/breaktests/App.java
amirhosv: Good call. I'll add it back.

CMakeLists.txt
amirhosv: We should be able. Let me try that.

WillChilds-Klein (approved):

PR #286, opened 2023-04-19
Only register LibCryptoRng by default in non-FIPS mode

WillChilds-Klein:
*Issue #, if available:* t/P82298726

*Description of changes:*

# Notes

See the updated README for this option's motivation.

Unlike `registerEcParams`, we only provide this option as a JVM property
and not as a runtime method call. The reason for this is that the JCA
treats SecureRandomly differently from other algorithms, maintining
provider registry state across both [SecureRandom][1] and [Provider][2]
classes. This significantly complicates registering SecureRandom
implementations post-initialization for a given provider (e.g. ACCP)
after instances of SecureRandom have already been generated from a
different provider (e.g. SUN). So, for now, we only expose this option
as a JVM property set at process start.

[1]: https://github.com/corretto/corretto-11/blob/1fee80d48d98e91a6a7ad1ca24220ccc7c148ef3/src/java.base/share/classes/java/security/SecureRandom.java#L274
[2]: https://github.com/corretto/corretto-11/blob/1fee80d48d98e91a6a7ad1ca24220ccc7c148ef3/src/java.base/share/classes/java/security/Provider.java#L1411

# Benchmarking

The new benchmark (`newThreadPerRequest`) clearly demonstrates that
the performance cost of per-thread jitter initialization.

```diff
$ git diff
diff --git a/build.gradle b/build.gradle
index 1ba8ea9..e88f633 100644
--- a/build.gradle
+++ b/build.gradle
@@ -19,6 +19,7 @@ ext.isFips = Boolean.getBoolean('FIPS')
 ext.isLegacyBuild = Boolean.getBoolean('LEGACY_BUILD')

 jmh {
+    includes = ['Drbg']
     fork = 1
     benchmarkMode = ['thrpt']
     threads = 1
@@ -385,7 +386,6 @@ task unit_tests(type: Copy) {
     from "${buildDir}/cmake/unit-tests/"
     into "${buildDir}/reports/unit-tests"
 }
-test.dependsOn unit_tests

 task checkstyle(type: Exec) {
     dependsOn executeCmake

$ ./gradlew -DFIPS=true clean jmh jmhReport
...
Benchmark                                   (provider)  (size)   Mode  Cnt       Score        Error  Units
Drbg.multiThreaded        AmazonCorrettoCryptoProvider    1024  thrpt    5  746237.847 _ 115563.930  ops/s
Drbg.multiThreaded                                  BC    1024  thrpt    5   35516.850 _    283.508  ops/s
Drbg.multiThreaded                                 SUN    1024  thrpt    5  100245.322 _   3642.675  ops/s
Drbg.newThreadPerRequest  AmazonCorrettoCryptoProvider    1024  thrpt    5      30.888 _      0.779  ops/s
Drbg.newThreadPerRequest                            BC    1024  thrpt    5    9296.259 _   1265.277  ops/s
Drbg.newThreadPerRequest                           SUN    1024  thrpt    5   11473.685 _   1121.174  ops/s
Drbg.singleThreaded       AmazonCorrettoCryptoProvider    1024  thrpt    5  761595.931 _   4866.979  ops/s
Drbg.singleThreaded                                 BC    1024  thrpt    5   35317.928 _    996.175  ops/s
Drbg.singleThreaded                                SUN    1024  thrpt    5   99673.483 _   1124.420  ops/s
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -601,6 +601,9 @@ set(COVERAGE_ARGUMENTS
 
 if(FIPS)
     set(TEST_FIPS_PROPERTY "-DFIPS=true")
+    # ACCP's default behavior in FIPS mode is to not register a SecureRandom implementation.
+    # However, we explicitly register it here to ensure its coverage under test.
+    set(REGISTER_RNG_PROPERTY "-Dcom.amazon.corretto.crypto.provider.registerSecureRandom=true")
 else()
     set(TEST_FIPS_PROPERTY "-DFIPS=false")
 endif()
@@ -614,6 +617,7 @@ set(TEST_RUNNER_ARGUMENTS
     ${TEST_ADD_OPENS}
     ${TEST_FIPS_PROPERTY}
     ${EXTERNAL_LIB_PROPERTY}
+    ${REGISTER_RNG_PROPERTY}
     -Dcom.amazon.corretto.crypto.provider.inTestSuite=hunter2
     -Dtest.data.dir=${TEST_DATA_DIR}
     -Djunit.jupiter.execution.parallel.enabled=true
@@ -811,6 +815,7 @@ if (CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
 add_custom_target(run-dieharder-libcrypto-rng
     COMMAND ${TEST_JAVA_EXECUTABLE} -cp $<TARGET_PROPERTY:accp-jar,JAR_FILE>:$<TARGET_PROPERTY:tests-jar,JAR_FILE>:${TEST_CLASSPATH}
         ${EXTERNAL_LIB_PROPERTY}
+        ${REGISTER_RNG_PROPERTY}
         com.amazon.corretto.crypto.provider.test.SecureRandomGenerator 'LibCryptoRng' 8192 1 |
         ${DIEHARDER_EXECUTABLE} -a -g 200 -Y 1 -k 2 |
         tee dieharder-results-libcrypto-rng.txt
@@ -820,6 +825,7 @@ add_custom_target(run-dieharder-libcrypto-rng
 add_custom_target(run-dieharder-libcrypto-rng-tail
     COMMAND ${TEST_JAVA_EXECUTABLE} -cp $<TARGET_PROPERTY:accp-jar,JAR_FILE>:$<TARGET_PROPERTY:tests-jar,JAR_FILE>:${TEST_CLASSPATH}
         ${EXTERNAL_LIB_PROPERTY}
+        ${REGISTER_RNG_PROPERTY}
         com.amazon.corretto.crypto.provider.test.SecureRandomGenerator 'LibCryptoRng' 31 1 |
         ${DIEHARDER_EXECUTABLE} -d 15 -g 200 -Y 1 -k 2 |
         tee dieharder-results-libcrypto-rng-tail.txt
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -268,6 +268,14 @@ Thus, these should all be set on the JVM command line using `-D`.
...
   Using JCE's impelmentation is generally recommended unless using ACCP as a standalone provider
   Callers can choose to register ACCP's implementation at runtime with a call to `AmazonCorrettoCryptoProvider.registerEcParams()`
+* `com.amazon.corretto.crypto.provider.registerSecureRandom`
+  Takes in `true` or `false` (defaults to `false` in FIPS mode, `true` in non-FIPS).
+  If `true`, then ACCP will register a SecureRandom implementation (`LibCryptoRng`) backed by AWS-LC
+  Else, ACCP will not register a SecureRandom implementation, meaning that the JCA will source SecureRandom instances from another registered provider.
+  LibCryptoRng is very fast during steady state operation in all cases. In FIPS mode, however, AWS-LC-FIPS's CPU jitter-based entropy source incurs a ~10ms initialization cost for every new thread.
+  This means that there is a slight "pause" before randomness production in highly threaded environments.
+  Because, in extreme cases, this could prevent an availability risk, we do not register LibCryptoRng by default in cases where this initialization cost is incurred (i.e. FIPS mode).
```
brian-jarvis-aws, line 277: >this could prevent an availability risk

I think "prevent" should be "present" here.

WillChilds-Klein: hm, i agree. pretty much the opposite of what we're trying to say... thanks. will fix.

CMakeLists.txt
```diff
@@ -601,6 +601,7 @@ set(COVERAGE_ARGUMENTS
 
 if(FIPS)
     set(TEST_FIPS_PROPERTY "-DFIPS=true")
+    set(REGISTER_RNG_PROPERTY "-Dcom.amazon.corretto.crypto.provider.registerSecureRandom=true")
```
geedo0, line 604: This warrants a comment to explain the rationale as it does run counter to the default behavior of the library. Something along the lines of "we want to test our own deterministically selected RNG in FIPS mode".

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -310,6 +314,7 @@ public AmazonCorrettoCryptoProvider() {
         super("AmazonCorrettoCryptoProvider", PROVIDER_VERSION, "");
         this.relyOnCachedSelfTestResults = Utils.getBooleanProperty(PROPERTY_CACHE_SELF_TEST_RESULTS, true);
         this.shouldRegisterEcParams = Utils.getBooleanProperty(PROPERTY_REGISTER_EC_PARAMS, false);
+        this.shouldRegisterSecureRandom = Utils.getBooleanProperty(PROPERTY_REGISTER_SECURE_RANDOM, !isFips());
```
geedo0, line 317: An explanatory comment around the FIPS performance penalty would be helpful here. It's not immediately apparent why the defaulting is the way it is here.

geedo0 (dismissed):

CMakeLists.txt
WillChilds-Klein: good call. will add.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
WillChilds-Klein: i hesitated to add one out of fear of it growing stale through time (AWS-LC's DRBG init latency may improve), but I agree that it's odd enough that it warrants a comment. will add.

README.md
```diff
@@ -268,6 +268,14 @@ Thus, these should all be set on the JVM command line using `-D`.
   Else, the JCA will get the implementation from another registered provider (usually stock JCE).
   Using JCE's impelmentation is generally recommended unless using ACCP as a standalone provider
   Callers can choose to register ACCP's implementation at runtime with a call to `AmazonCorrettoCryptoProvider.registerEcParams()`
+* `com.amazon.corretto.crypto.provider.registerSecureRandom`
+  Takes in `true` or `false` (defaults to `false` in FIPS mode, `true` in non-FIPS).
```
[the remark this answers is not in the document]
WillChilds-Klein: also an excellent call-out. we need to be very explicit about this. will add.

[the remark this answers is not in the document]
brian-jarvis-aws: I agree, it does seem counterintuitive. Especially since we're being opinionated about the default state (as opposed to having registerSecureRandom always default to false). But perhaps it's worth clarifying that this property only affects the plumbing up to the Java layer. In FIPS mode, AWS-LC will continue to use its own FIPS-validated entropy source internally regardless of this property value.

brian-jarvis-aws (approved):

geedo0 (approved):

amirhosv (approved):
LGTM

PR #287, opened 2023-04-20
Setup Automated Code Formatting with Spotless

geedo0:
*Description of changes:*
[Spotless](https://github.com/diffplug/spotless) is a Gradle plugin that enforces style checking and automates remediation of any violations. This replaces our existing Checkstyle integration which has 
not been kept up and did not do any build-time enforcement of the rule-set.

I've configured it to check our Java source against the [Google Java Format](https://github.com/google/google-java-format) and apply our license header consistently. In the future we can integrate with 
the clang-format extension to handle the C++ source.

After this gets merged, you'll notice a few changes to our Gradle build. First and foremost, style violations will fail the build either locally or in our CI builds if it gets up to the PR. The build 
failure will prompt the user to run `./gradlew spotlessApply`. Running that target will apply any style changes needed to pass the check phase of the build and we can all move along after it gets 
committed.

Note to reviewers:
All changes under src/tst are automatically applied by the code formatter. The actual meat of this review are the changes to the build scripts.


amirhosv:
It would be great to add the information in the description of this PR to `DEVELOPMENT-GUIDE.md`.

DEVELOPMENT-GUIDE.md
```diff
@@ -45,6 +45,25 @@ In decreasing order of importance:
...
+
+```
+# Run this to automatically run the formatter
+./gradlew spotlessApply
+
+# Example of an embarassing build failure which could've been avoided
+> Task :spotlessJavaCheck FAILED
+...
+> The following files had format violations:
+  Run './gradlew :spotlessApply' to fix these violations.
```
WillChilds-Klein, line 65: do you need the `:` prefix on the task name?

build.gradle
```diff
@@ -535,9 +543,9 @@ task coverage {
 
 task release {
     if (isLegacyBuild) {
-        dependsOn build, test, javadoc, src_jar, checkstyle
```
WillChilds-Klein, line 538: should we invoke spotless here instead?

DEVELOPMENT-GUIDE.md
geedo0: No, but this is exactly what Gradle prints out for us on a failure. Maybe they both work?

build.gradle
geedo0: No, Spotless registers itself as a dependency on the `check` step which is a dependency of `build`. The reason `checkstyle` was declared here was probably because we implemented the task by hand and didn't completely follow Gradle convention.

WillChilds-Klein: hm, it would be nice to only invoke spotless on `release` so that we can iterate quickly (if un-stylishly) with `./gradlew cmake_clean build`. but i guess it's not too bad to just apply the fixes inline with `./gradlew cmake_clean spotlessApply build` and then refresh your editor pane.

> Spotless registers itself as a dependency on the check step which is a dependency of build

what does `check` do?

geedo0: I think having spotless automate formatting is painless enough to allow us to make it a required part of the build. It's not as annoying as checkstyle or findbugs which would just break your build and force you to either look up how to bypass them or figure out how to fix your code.

Here's a blurb from the Gradle docs about `check`.

>   check — lifecycle task -  Plugins and build authors should attach their verification tasks, such as ones that run tests, to this lifecycle task using check.dependsOn(task).

https://docs.gradle.org/current/userguide/base_plugin.html#sec:base_tasks

WillChilds-Klein: ok, cool. TIL about `-x`, very useful.

WillChilds-Klein (approved):

amirhosv (dismissed):
There are some conflicts with main that needs to be resolved.

amirhosv (approved):

WillChilds-Klein (approved):

PR #290, opened 2023-04-27
Check if generated RSA keys pass FIPS checks when built in FIPS mode

amirhosv:
*Description of changes:*

Check if generated RSA keys pass FIPS checks when built in FIPS mode

### DIFF
--- a/csrc/rsa_gen.cpp
+++ b/csrc/rsa_gen.cpp
@@ -5,6 +5,7 @@
 #include <cstring> // for memset
 #include <openssl/rsa.h>
 #include <openssl/bn.h>
+#include <openssl/crypto.h>
 #include "generated-headers.h"
 #include "keyutils.h"
 #include "util.h"
@@ -29,14 +30,22 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_RsaGen_generate
         BigNumObj bne;
         jarr2bn(env, pubExp, bne);
 
-        if (1 != RSA_generate_key_ex(r, bits, bne, NULL))
-        {
+        if (1 != RSA_generate_key_ex(r, bits, bne, NULL)) {
             throw_openssl("Unable to generate key");
         }
 
-        if (checkConsistency && RSA_check_key(r) != 1)
-        {
-            throw_openssl("Key failed consistency check");
+        if (FIPS_mode() == 1) {
+            // We should be using RSA_generate_key_fips, but since RSA_generate_key_fips doesn't allow
+            // bit lengths greater than 4096 and public exponents other than F4, we use RSA_generate_key_ex
+            // and explicitly check FIPS related conditions. This is done to ensure that ACCP built in FIPS mode
+            // is compatible with BC-FIPS.
+            if (RSA_check_fips(r) != 1) {
+                throw_openssl("RSA_check_fips failed");
+            }
+        } else {
+            if (checkConsistency && RSA_check_key(r) != 1) {
+                throw_openssl("Key failed consistency check");
+            }
         }
 
         EVP_PKEY_auto result = EVP_PKEY_auto::from(EVP_PKEY_new());
--- a/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
+++ b/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
@@ -491,12 +491,9 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
     selfTestSuite.assertAllTestsPassed();
   }
 
-  /**
-   * Returns {@code true} if and only if the underlying libcrypto library is a FIPS build and passes
-   * FIPS KAT self tests
-   */
+  /** Returns {@code true} if and only if the underlying libcrypto library is a FIPS build */
   public boolean isFips() {
-    return Loader.FIPS_BUILD && SelfTestSuite.awsLcSelfTestsPassed();
+    return Loader.FIPS_BUILD;
   }
 
   /**
--- a/src/com/amazon/corretto/crypto/provider/RsaGen.java
+++ b/src/com/amazon/corretto/crypto/provider/RsaGen.java
@@ -12,6 +12,7 @@ import java.security.spec.AlgorithmParameterSpec;
 import java.security.spec.RSAKeyGenParameterSpec;
 
 class RsaGen extends KeyPairGeneratorSpi {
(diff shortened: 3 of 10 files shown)


src/com/amazon/corretto/crypto/provider/RsaGen.java
```diff
@@ -66,11 +67,17 @@ public void initialize(int keysize, SecureRandom rnd) throws InvalidParameterExc
     }
   }
 
-  private static RSAKeyGenParameterSpec validateParameter(RSAKeyGenParameterSpec spec)
+  private static RSAKeyGenParameterSpec validateParameter(final RSAKeyGenParameterSpec spec)
       throws InvalidAlgorithmParameterException {
-    if (spec.getKeysize() < 512) {
+    if (Loader.FIPS_BUILD && spec.getPublicExponent().compareTo(RSAKeyGenParameterSpec.F4) <= -1) {
```
geedo0, line 72: Earlier, you say this:
> RSA_generate_key_fips doesn't allow  ... public exponents other than F4

That suggests that we should be checking for equality, not a minimum exponent size. Which is correct?

geedo0 (changes requested):

csrc/rsa_gen.cpp
```diff
@@ -29,11 +29,18 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_RsaGen_generate
         BigNumObj bne;
         jarr2bn(env, pubExp, bne);
 
-        if (1 != RSA_generate_key_ex(r, bits, bne, NULL))
-        {
+        if (1 != RSA_generate_key_ex(r, bits, bne, NULL)) {
             throw_openssl("Unable to generate key");
         }
 
+#ifdef ACCP_FIPS_BUILD
```
WillChilds-Klein, line 36: instead of adding a new build flag, why not use this function to determine FIPSness as we do elsewhere?

https://github.com/aws/aws-lc/blob/ae87faf735c0241a115542b1c1022d125564bf55/include/openssl/crypto.h#L92

src/com/amazon/corretto/crypto/provider/RsaGen.java
amirhosv: AWS-LC doesn't follow BC-FIPS behavior in this case. Here I'm following BC-FIPS behavior.

WillChilds-Klein: i think L72 is the public exp equality check and L77 is the key size check. from [BigInteger docs](https://docs.oracle.com/javase/7/docs/api/java/math/BigInteger.html#compareTo(java.math.BigInteger)):

> Compares this BigInteger with the specified BigInteger. This method is provided in preference to individual methods for each of the six boolean comparison operators (<, ==, >, >=, !=, <=). The suggested idiom for performing these comparisons is: (x.compareTo(y) <op> 0), where <op> is one of the six comparison operators.

by that convention, perhaps this should be

```
spec.getPublicExponent().compareTo(RSAKeyGenParameterSpec.F4) != 0
```

or just use `.equals` (we risk NPE if proper input validation isn't done on `spec`)

```
!RSAKeyGenParameterSpec.F4.equals(spec.getPublicExponent())
```

?

WillChilds-Klein: hm, so FIPS allows non-standard public exponents as long as they're >= 65537?

amirhosv: I wouldn't say FIPS, but BC-FIPS yes.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -546,6 +546,7 @@ private static void assertPssParamsEqual(PSSParameterSpec s1, PSSParameterSpec s
 
   @Test
   void testRsaPSSInitSmallKeyAfterSetParameterLargeSaltThrows() throws Exception {
+    if (TestUtil.isFips()) return;
```
WillChilds-Klein, line 549: are we able to adapt this test case to work under FIPS? perhaps by setting `minimallySecureKeyLen = TestUtil.isFips()? 2048 : 4096` and `smallKeySize = minimallySecureKeyLen / 2`?

if not, instead of returning can we skip it with `Assumptions.assumeTrue(!TestUtil.isFips())` and add a comment explaining why we skip this test in FIPS?

csrc/rsa_gen.cpp
amirhosv: In the Java part of ACCP, you're right that we check FIPSness during run-time and I guess that's because in Java pre-processing the code is not common; however, in C/C++, we can do this at compile time and have a little bit of performance improvement compared to doing it at run-time.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
amirhosv: `Assumptions.assumeTrue` is a much cleaner way of doing it. I'll update the PR.

src/com/amazon/corretto/crypto/provider/RsaGen.java
amirhosv: @geedo0 I added comments to clarify the behavior.

csrc/rsa_gen.cpp
amirhosv: I agree that the performance gain is minuscule, but another benefit is that we would not have that code path when we build for FIPS; having said that, I switched to using `FIPS_mode` API in the latest patch.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -491,12 +491,9 @@ public void assertHealthy() throws RuntimeCryptoException {
     selfTestSuite.assertAllTestsPassed();
   }
 
-  /**
-   * Returns {@code true} if and only if the underlying libcrypto library is a FIPS build and passes
-   * FIPS KAT self tests
-   */
+  /** Returns {@code true} if and only if the underlying libcrypto library is a FIPS build */
   public boolean isFips() {
```
amirhosv, line 495: I'm reverting this to its old behavior as in 2.0.0; if a customer need to run the KAT self tests, they can invoke `runSelfTests` and it would run all self tests including KAT ones.

WillChilds-Klein (approved):

geedo0 (approved):

PR #292, opened 2023-04-28
Setup C++ Code Formatting with clang-format

geedo0:
*Issue #, if available:*

*Description of changes:*
This configures the clang-format Spotless extension for our C++ source. There was a .clang-format configuration file that existed in this repository. It's based on Webkit style with a few modifications which is fine. I removed a few of the style overrides around alignment and whitespace which I agree with Apple's stance on. The rest I left in place. It was not included as part of the build so our style has deviated over the years and re-applying the style rules resulted in minor breakage, but it was easy enough to fix-up manually (some unfortunately placed comments confused the formatter).

The clang-format extension differs from the Java formatter in that it does require a clang-format binary from the system to shell out to. To further complicate this, Spotless wants an exact version number which varies widely between package managers. I'm assuming that the ruleset is stable between these and that different versions of clang-format do not produce different results, but that remains to be proven by the first complete CI run. As a flexibility measure, I made C++ style checking less strict than Java and only enforcing it if the underlying build machine has the binary available. The binary will only become available when we get arund to spinning up new images.

Longer term, once all the kinks are worked out we can make this a required check in the build across all platforms.


geedo0:
Simply switching to C'03 didn't work out well, this will need a bit more investigation.

build.gradle
```diff
@@ -64,6 +64,40 @@ spotless {
...
+    native package managers. Work around this by shelling out and extracting
+    the version string.
+ */
+def getClangFormatVersion() {
+    def version_command = 'clang-format --version'
+    def shell_output = new ByteArrayOutputStream()
+    exec {
+        commandLine "bash", "-c", version_command
+        standardOutput = shell_output
+        ignoreExitValue true
```
WillChilds-Klein, line 92: presumably, this prevents gradle from blowing up when the `clang-format` executable can't be found on PATH?

WillChilds-Klein (approved):

amirhosv (approved):

WillChilds-Klein (approved):

build.gradle
geedo0: Yeah, non-zero exit code from the command within an exec call throws an exception in Gradle.

PR #295, opened 2023-05-05
SecretKeyGenerator's DefaultSecureRandomSupplier always uses LibCryptoRng

WillChilds-Klein:
# Notes

We recently made [a change][1] to ACCP making the JCA registry of a SecureRandom implementation opt-in in FIPS mode. This had the unintended effect of breaking SecretKey generation in FIPS mode, as our SecretKeyGenerator internally sourced its LibCryptoRng instance from the JCA. Because we didn't register LibCryptoRng with the JCA by default in FIPS mode, exceptions got thrown.

The fix here is to just instantiate LibCryptoRng directy instead of obtaining it through the JCA. This has the added benefit of making sure we're _always_ using AWS-LC's DRBG for SecretKey generation by default, eliminating the possiblity of throw/fallback flows.

[1]: https://github.com/corretto/amazon-corretto-crypto-provider/pull/286

# Testing

```
$ git diff
diff --git a/src/com/amazon/corretto/crypto/provider/SecretKeyGenerator.java b/src/com/amazon/corretto/crypto/provider/SecretKeyGenerator.java
index 01ab338..52775ae 100644
--- a/src/com/amazon/corretto/crypto/provider/SecretKeyGenerator.java
+++ b/src/com/amazon/corretto/crypto/provider/SecretKeyGenerator.java
@@ -73,7 +73,13 @@ class SecretKeyGenerator extends KeyGeneratorSpi {

     @Override
     public SecureRandom get() {
-      return new LibCryptoRng();
+      try {
+        return SecureRandom.getInstance(
+            LibCryptoRng.ALGORITHM_NAME, AmazonCorrettoCryptoProvider.INSTANCE);
+      } catch (final NoSuchAlgorithmException e) {
+            LibCryptoRng.ALGORITHM_NAME + " is not a provided by AmazonCorrettoCryptoProvider", e);
+      }
     }
   }


$ ./gradlew -DFIPS=true cmake_clean test
...
Exception in thread "main" java.lang.AssertionError: LibCryptoRng is not a provided by AmazonCorrettoCryptoProvider
        at com.amazon.corretto.crypto.provider.SecretKeyGenerator$DefaultSecureRandomSupplier.get(SecretKeyGenerator.java:80)
        at com.amazon.corretto.crypto.provider.SecretKeyGenerator$DefaultSecureRandomSupplier.get(SecretKeyGenerator.java:66)
        at java.base/java.util.Optional.orElseGet(Optional.java:369)
        at com.amazon.corretto.crypto.provider.SecretKeyGenerator.engineGenerateKey(SecretKeyGenerator.java:59)
        at java.base/javax.crypto.KeyGenerator.generateKey(KeyGenerator.java:563)
        at com.amazon.corretto.crypto.provider.test.SecurityPropertyTester.main(SecurityPropertyTester.java:62)
Caused by: java.security.NoSuchAlgorithmException: no such algorithm: LibCryptoRng for provider AmazonCorrettoCryptoProvider
        at java.base/sun.security.jca.GetInstance.getService(GetInstance.java:101)
        at java.base/sun.security.jca.GetInstance.getInstance(GetInstance.java:218)
        at java.base/java.security.SecureRandom.getInstance(SecureRandom.java:469)
        at com.amazon.corretto.crypto.provider.SecretKeyGenerator$DefaultSecureRandomSupplier.get(SecretKeyGenerator.java:77)
        ... 5 more
```

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/SecretKeyGenerator.java
+++ b/src/com/amazon/corretto/crypto/provider/SecretKeyGenerator.java
@@ -73,13 +73,7 @@ class SecretKeyGenerator extends KeyGeneratorSpi {
 
     @Override
     public SecureRandom get() {
-      try {
-        return SecureRandom.getInstance(
-            LibCryptoRng.ALGORITHM_NAME, AmazonCorrettoCryptoProvider.INSTANCE);
-      } catch (final NoSuchAlgorithmException e) {
-        throw new AssertionError(
-            LibCryptoRng.ALGORITHM_NAME + " is not a provided by AmazonCorrettoCryptoProvider", e);
-      }
+      return new LibCryptoRng();
     }
   }
 
--- a/tst/com/amazon/corretto/crypto/provider/test/SecurityPropertyTester.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/SecurityPropertyTester.java
@@ -13,6 +13,7 @@ import java.security.SecureRandom;
 import java.security.Security;
 import java.util.HashSet;
 import java.util.Set;
+import javax.crypto.KeyGenerator;
 import javax.net.ssl.SSLContext;
 
 /**
@@ -53,6 +54,12 @@ public final class SecurityPropertyTester {
       assertEquals(NATIVE_PROVIDER.getName(), new SecureRandom().getProvider().getName());
       assertEquals(NATIVE_PROVIDER.getName(), strongRng.getProvider().getName());
     }
+
+    // Ensure that we can successfully generate an AES key, regardless of FIPS
+    // mode or whether ACCP registers a SecureRandom implementation.
+    KeyGenerator aesKeyGen = KeyGenerator.getInstance("AES");
+    assertEquals(NATIVE_PROVIDER.getName(), aesKeyGen.getProvider().getName());
+    aesKeyGen.generateKey();
 
     // Also ensure that nothing shows up twice
     Set<String> names = new HashSet<>();


amirhosv (approved):
It's good that LibCryptoRng is a SecureRandom and not a SecureRandomSpi.

src/com/amazon/corretto/crypto/provider/SecretKeyGenerator.java
```diff
@@ -73,13 +73,7 @@ private DefaultSecureRandomSupplier() {
...
     @Override
     public SecureRandom get() {
-      try {
-        return SecureRandom.getInstance(
-            LibCryptoRng.ALGORITHM_NAME, AmazonCorrettoCryptoProvider.INSTANCE);
-      } catch (final NoSuchAlgorithmException e) {
-        throw new AssertionError(
-            LibCryptoRng.ALGORITHM_NAME + " is not a provided by AmazonCorrettoCryptoProvider", e);
-      }
+      return new LibCryptoRng();
```
alexw91, line 76: Just want to make sure, if we are NOT in FIPS mode then we still want to use `LibCryptoRng()`? Are there any cases where we'd want SecretKeyGenerator to fall back to non-ACCP vended SecureRandom?

geedo0 (approved):

src/com/amazon/corretto/crypto/provider/SecretKeyGenerator.java
WillChilds-Klein: > if we are NOT in FIPS mode then we still want to use LibCryptoRng()?

I would argue yes. The reason is that SecretKey material should be derived from as high-quality of a source of randomness as possible. This also fits the pattern we establish in other asymmetric key (pair) generators of always using AWS-LC's DRBG (though in those cases it happens internally within AWS-LC).

> Are there any cases where we'd want SecretKeyGenerator to fall back to non-ACCP vended SecureRandom?

Perhaps in a scenario similar to the one that motivated PR #286 (latency-sensitive key generation with high rate of thread creation). However, @amirhosv points out that he wrote SecretKeyGenerator in such a way that the SecureRandom implementation used is configurable. This change just affects the default implementation used.

alexw91 (approved):

src/com/amazon/corretto/crypto/provider/SecretKeyGenerator.java
amirhosv: It's part of JCE/JCA. [KeyGenerator::init](https://docs.oracle.com/javase/8/docs/api/javax/crypto/KeyGenerator.html#init-java.security.spec.AlgorithmParameterSpec-java.security.SecureRandom-) can accept `SecureRandom`. In case no SecureRandom is provided when [KeyGenerator::generateKey](https://docs.oracle.com/javase/8/docs/api/javax/crypto/KeyGenerator.html#generateKey--) is invoked, we fall back on `LibCryptoRng`.

Issue #271, opened 2023-02-15
Heavy allocation compared to the openjdk provider


carterkozak:
ACCP appears to allocate an order of magnitude more heavily than the default provider, hindering throughput and impacting application latency. I've recorded my benchmark results in this commit message: https://github.com/carterkozak/java-crypto-allocation-performance/commit/1313c9762e21f2cd1201295b4265c51e8ccb98d2

This branch updates an existing benchmark I've used elsewhere to include ACCP, the results (as they impact both throughput and a allocation/gc) apply to both the SSLSocket-based benchmark and DirectByteBuffer+SSLEngine variant: https://github.com/carterkozak/java-crypto-allocation-performance/tree/ckozak/accp

I'd be happy to provide more information if that's helpful.

travispeloton:
Hey @carterkozak , I'm not a maintainer here, but curious if this finding affects only `2.x` and if you have tested on `1.6.1`?

carterkozak:
My results were consistent across both releases, however I didn’t test 1.6.1 particularly vigorously. Feel free to run the linked benchmarks, it’s possible that my results don’t hold true on all systems (for example, some of the openjdk ciphers perform and allocate differently on cpus with vaes instructions)

analytically:
https://github.com/corretto/amazon-corretto-crypto-provider/commit/598a82bd483b6c6f86efb8d808a060956e059b2d

Doesn't seem replicated in their own performance tests.

carterkozak:
@analytically There's a great deal of complexity outside of the cipher itself. Much of the allocation is coming from a conditional where accp cannot tell if two buffers share a region of memory, thus must allocate additional buffers. OpenJDK has had similar issues which also weren't captured by cipher microbenchmarks for the same reason.

WillChilds-Klein:
thank you for the report @carterkozak. we're looking into this.

geedo0:
Here are the benchmark numbers after all these fixes have been applied. Our JVM memory usage is now below SunJCE's and we are spending way less time in GC overall. Once these PRs get merged, we can close this issue.

```
"Benchmark","Mode","Threads","Samples","Score","Score Error (99.9%)","Unit","Param: cipher","Param: length","Param: provider"
[...]

PR #296, opened 2023-05-11
Directly buffer ciphertext bytes during doUpdate() for AES-GCM decryption

geedo0:
*Issue #, if available:*
#271

*Description of changes:*
Our implementation of engineUpdate for AES-GCM does not return intermediate data to the output buffer. Instead, it simply caches the data to an internal buffer until doFinal() is called. This is to ensure that we validate the tag before returning the plain text and not leave the caller holding the bag.

That said, ideally, doUpdate should simply perform a memcopy into the internal buffer and return. If necessary, it should perform a dynamic re-allocation to accommodate additional data (amortized O(n). This change implements that ideal by completely bypassing the complicated workflow we were doing previously.

Things we avoid:
- Creating a complete copy of the input buffer which is done to avoid overwriting the input. However, in our decrypt implementation we do not write any output so this was pointless. O(n) memory savings.
- Tight loop which chunked our loading of the internal decrypt buffer in 64k increments.
  + O(n) Allocations of small `ShimArray` objects on each loop iteration.
  + O(n^2) memory I/O and O(n) additional memory due to the array resizing from the incremental loading of data

The end result of this is that we will have much lower garbage collector pressure overall for larger payload sizes (>10MB). Based on the benchmarks provided in issue #271, this is a 33% improvement. It's still not where we want to be but it gets us closer.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
@@ -890,65 +890,69 @@ final class AesGcmSpi extends CipherSpi {
   @Override
   protected int engineUpdate(ByteBuffer input, final ByteBuffer output)
       throws ShortBufferException {
-    ByteBuffer bufferForClear = null;
+    switch (opMode) {
+      case NATIVE_MODE_DECRYPT:
+        // Our implementation of engineUpdate for decrypt doesn't actually return any data, we
+        // simply buffer the ciphertext and leave the output buffer alone, so we don't bother
+        // passing it through.
+        ShimArray cipherText = new ShimArray(input, input.remaining());
+        engineUpdate(cipherText.array, cipherText.offset, cipherText.length, null, 0);
+        return 0;
+      case NATIVE_MODE_ENCRYPT:
+        ByteBuffer bufferForClear = null;
 
-    // The default JCE implementation of this bytebuffer-to-byte[] shim seems to break when
-    // engineGetOutputSize
-    // returns more bytes then is actually used in each round (it only calls engineGetOutputSize
-    // once, on the entire
-    // input size, and does not properly size the output buffer for each round). By coincidence this
-    // works as long
-    // as the cipher actually knows how much space it's going to use for its bounds checking and the
-    // actual buffer
-    // sizes for input and output match the cipher block size - but in our case we don't know what
-    // EVP's going to do
-    // and have to be conservative, requiring a larger output than input buffer. So we have to
-    // implement this loop
-    // ourselves.
+        // The default JCE implementation of this bytebuffer-to-byte[] shim seems to break when
+        // engineGetOutputSize returns more bytes then is actually used in each round (it only calls
+        // engineGetOutputSize once, on the entire input size, and does not properly size the output
+        // buffer for each round). By coincidence this works as long as the cipher actually knows
+        // how much space it's going to use for its bounds checking and the actual buffer sizes for
+        // input and output match the cipher block size - but in our case we don't know what EVP's
+        // going to do and have to be conservative, requiring a larger output than input buffer. So
+        // we have to implement this loop ourselves.
 
-    int initialPosition = output.position();
+        int initialPosition = output.position();
 
-    if (output.remaining() < engineGetOutputSize(input.remaining())) {
-      throw new ShortBufferException();
+        if (output.remaining() < engineGetOutputSize(input.remaining())) {
+          throw new ShortBufferException();
+        }
+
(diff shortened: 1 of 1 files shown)


src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -890,65 +890,69 @@ private void writeback() {
...
-    }
+    switch (opMode) {
+      case NATIVE_MODE_DECRYPT:
+        // Our implementation of engineUpdate for decrypt doesn't actually return any data, we
+        // simply buffer the ciphertext and leave the output buffer alone, so we don't bother
+        // passing it through.
+        ShimArray cipherText = new ShimArray(input, input.remaining());
+        engineUpdate(cipherText.array, cipherText.offset, cipherText.length, null, 0);
+        return 0;
+      case NATIVE_MODE_ENCRYPT:
```
geedo0, line 901: This is the only real change, the rest of it is just code motion.

WillChilds-Klein:
> The end result of this is that we will have much lower garbage collector pressure overall for larger payload sizes (>10MB). Based on the benchmarks provided in issue https://github.com/corretto/amazon-corretto-crypto-provider/issues/271, this is a 33% improvement.

Do we have benchmark data proving this out?

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -890,65 +890,69 @@ private void writeback() {
...
-
-    int initialPosition = output.position();
-
-    if (output.remaining() < engineGetOutputSize(input.remaining())) {
-      throw new ShortBufferException();
-    }
+    switch (opMode) {
+      case NATIVE_MODE_DECRYPT:
+        // Our implementation of engineUpdate for decrypt doesn't actually return any data, we
+        // simply buffer the ciphertext and leave the output buffer alone, so we don't bother
```
WillChilds-Klein, line 896: for my own edification, quick check of what we're doing here -- my understanding of AES GCM is that it computes the GMAC tag over the ciphertext. reading through the code a bit, it looks like AWS-LC [incrementally computes the tag](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/modes/gcm.c#L774) from incremental updates of the cipher object. that computed tag is [then compared against the actual tag](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/modes/gcm.c#L828) on decrypt finalization. because AWS-LC computes the tag incrementally, we don't need to buffer the input ciphertext.

1. does my understanding align with yours?
2. if so, why did we do the input ciphertext buffering in the first place? all just to allow callers to use the same buffer for input/output?

geedo0: > does my understanding align with yours?

From an algorithmic perspective, the AES-GCM tag is designed to be efficiently computed block by block. Nothing prevents an implementation from doing it this way and AWS-LC naturally exploits this.

> if so, why did we do the input ciphertext buffering in the first place? all just to allow callers to use the same buffer for input/output?

This is just guessing at the legacy, but I think the original author wanted to be conservative from a security perspective. Doing it this way allows us to zero out the output buffer when the tag validation fails. Though that doesn't stop an attacker from peering at the output buffer before it gets zero'd out so it's not a perfect mitigation.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -890,65 +890,69 @@ private void writeback() {
...
+        // passing it through.
+        ShimArray cipherText = new ShimArray(input, input.remaining());
```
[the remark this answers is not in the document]
geedo0: ShimArray is a pretty lightweight object. The key thing to note is that [duplicate](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#duplicate--) is a shallow copy so there aren't any savings here.

That said, there is room for improvement [here](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java#L930-L931) where we allocate a ton of these ShimArray's in a tight loop. Though the savings aren't going to be dramatic.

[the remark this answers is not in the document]
WillChilds-Klein: ah, didn't realize `duplicate` was a shallow copy. in that case, yeah agreed that the `new ShimArray` isn't significant.

WillChilds-Klein (approved):

amirhosv (approved):

PR #297, opened 2023-05-15
Fix bug in output buffer size check

geedo0:
*Issue #, if available:*
N/A

*Description of changes:*
This is calling the wrong function and getting the wrong size. Note how
the exception's error message used a different function than the one
used in the error checking logic. For encryptions, this causes our
safety check to be off by the tag length. Currently unused for
decryptions, but this would return 0 when it should be the size of the
plaintext.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
@@ -448,7 +448,7 @@ final class AesGcmSpi extends CipherSpi {
         }
       case NATIVE_MODE_ENCRYPT:
         {
-          checkOutputBuffer(length, output, outputOffset);
+          checkOutputBuffer(length, output, outputOffset, false);
 
           lazyInit();
 
@@ -542,7 +542,7 @@ final class AesGcmSpi extends CipherSpi {
         input = EMPTY_ARRAY;
       }
 
-      checkOutputBuffer(length, output, outputOffset);
+      checkOutputBuffer(length, output, outputOffset, true);
       checkArrayLimits(input, offset, length);
 
       final boolean overlaps =
@@ -803,7 +803,7 @@ final class AesGcmSpi extends CipherSpi {
     if (input == null) {
       input = EMPTY_ARRAY;
     }
-    checkOutputBuffer(length, output, outputOffset);
+    checkOutputBuffer(length, output, outputOffset, true);
     checkArrayLimits(input, offset, length);
     if (opMode == NATIVE_MODE_DECRYPT) {
       final byte[] plaintext = engineDecryptFinal(input, offset, length);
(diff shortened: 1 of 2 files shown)


src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -967,17 +967,20 @@ protected int engineDoFinal(final ByteBuffer input, final ByteBuffer output)
...
   }
 
-  private void checkOutputBuffer(final int inputLength, final byte[] output, final int outputOffset)
+  private void checkOutputBuffer(
+      final int inputLength, final byte[] output, final int outputOffset, boolean doFinal)
       throws ShortBufferException {
     if (inputLength < 0 || outputOffset < 0 || outputOffset > output.length) {
       throw new ArrayIndexOutOfBoundsException();
     }
-    if (output.length - outputOffset < getUpdateOutputSize(inputLength)) {
```
WillChilds-Klein, line 975: nice catch! i'm surprised we didn't have a unit test covering this. should we add a case [here](https://github.com/corretto/amazon-corretto-crypto-provider/blob/c2af9193453750db04c31171a51692b926e82693/tst/com/amazon/corretto/crypto/provider/test/AesTest.java#L704) to prevent regressions?

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -967,17 +967,20 @@ protected int engineDoFinal(final ByteBuffer input, final ByteBuffer output)
     return output.position() - initialPosition;
   }
 
-  private void checkOutputBuffer(final int inputLength, final byte[] output, final int outputOffset)
+  private void checkOutputBuffer(
+      final int inputLength, final byte[] output, final int outputOffset, boolean doFinal)
```
amirhosv, line 971: `final` for doFinal

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -967,17 +967,20 @@ protected int engineDoFinal(final ByteBuffer input, final ByteBuffer output)
...
       throws ShortBufferException {
     if (inputLength < 0 || outputOffset < 0 || outputOffset > output.length) {
```
amirhosv, line 973: Does `outputOffset == output.length` make sense?

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -967,17 +967,20 @@ protected int engineDoFinal(final ByteBuffer input, final ByteBuffer output)
...
       throw new ArrayIndexOutOfBoundsException();
     }
-    if (output.length - outputOffset < getUpdateOutputSize(inputLength)) {
+    int freeBufferSpace = output.length - outputOffset;
```
amirhosv, line 976: I guess these variables can be final too.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
geedo0: Good catch! That's yet another bug in this routine. That case equates to writing to the first byte beyond the array limits.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -971,17 +971,26 @@ protected int engineDoFinal(final ByteBuffer input, final ByteBuffer output)
...
+    final int freeBufferSpace = output.length - outputOffset;
+    final int requiredBufferSpace =
+        doFinal ? engineGetOutputSize(inputLength) : getUpdateOutputSize(inputLength);
+    if (inputLength < 0 || outputOffset < 0) {
       throw new ArrayIndexOutOfBoundsException();
     }
-    if (output.length - outputOffset < getUpdateOutputSize(inputLength)) {
+    // Allow outputOffset to index into a 0-length empty array. We must trust that the rest of the
+    // code doesn't actually do this.
+    if (outputOffset > output.length || (outputOffset == output.length && output.length != 0)) {
```
WillChilds-Klein, line 985: nit: might be good to have branch coverage on the output.length == 0 case.

WillChilds-Klein (approved):

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
geedo0: FWIW, this is exercised as the happy case for a lot of our doUpdate/decrypt tests and incidentally a part of the doFinal/decrypt paths.

PR #298, opened 2023-05-16
Invert function overloading order for AES-GCM doFinal decryption

geedo0:
*Issue #, if available:*
#271

*Description of changes:*
There are 3 overloaded methods for calling into a Cipher module which implement different ways of returning the output of the operation. From a memory management perspective these can be differentiated by whether or not the caller is responsible for allocating memory for the output.

Previously in our implementation, the top layer of the call stack is a method which allocates a buffer for the output and returns the result. For other interfaces in which the caller is responsible for allocating the output buffer we would allocate yet another buffer and copy it over to the buffer the caller gave us. This results in O(n) additional memory allocations and I/O.

This change inverts the call stack so that the top layer returns the result into an output buffer passed by the caller. That way, we only allocate a new output buffer when one is not provided. This gives us another 33% improvement in memory allocation from the benchmarks provided in issue #271.

n.b. Github is kinda inflexible about breaking up PRs so we scooped up the [pre-requisite PR](https://github.com/corretto/amazon-corretto-crypto-provider/pull/297) in this one. Once that's merged, I'll update this PR to omit the commit.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
@@ -248,8 +248,10 @@ final class AesGcmSpi extends CipherSpi {
     }
   }
 
-  /* Returns the maximum amount of data that could be returned from an update (not doFinal) operation.
-   * Not exposed via the Cipher API, but used internally to allocate buffers to return to the caller.
+  /**
+   * Returns the maximum amount of data that could be returned from an update (not doFinal)
+   * operation. Not exposed via the Cipher API, but used internally to allocate buffers to return to
+   * the caller.
    */
   private int getUpdateOutputSize(final int inputLen) {
     switch (opMode) {
@@ -612,10 +614,8 @@ final class AesGcmSpi extends CipherSpi {
         }
       } else {
         // We need to make sure to add resultLength here; engineUpdate in encrypt mode produces
-        // incremental
-        // output (unlike in decrypt mode) and so we need to carry forward whatever amount of data
-        // it produced
-        // in our return value.
+        // incremental output (unlike in decrypt mode) and so we need to carry forward whatever
+        // amount of data it produced in our return value.
 
         keyUsageCount++;
 
@@ -654,8 +654,13 @@ final class AesGcmSpi extends CipherSpi {
     }
   }
 
-  private byte[] engineDecryptFinal(byte[] input, final int offset, final int length)
-      throws AEADBadTagException {
+  private int engineDecryptFinal(
+      byte[] input,
+      final int inputOffset,
+      final int inputLen,
+      byte[] output,
+      final int outputOffset)
+      throws AEADBadTagException, ShortBufferException {
     try {
       if (opMode != NATIVE_MODE_DECRYPT) {
         throw new IllegalStateException("Cipher not initialized for decryption");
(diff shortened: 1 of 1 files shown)


src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -675,31 +680,21 @@ private byte[] engineDecryptFinal(byte[] input, final int offset, final int leng
...
+      workingInputArray = decryptInputBuf.getDataBuffer();
+      workingInputLength = decryptInputBuf.size();
+      workingInputOffset = 0;
 
       if (workingInputLength < tagLength) {
         throw new AEADBadTagException("Input too short - need tag");
       }
 
-      final int expectedOutputLength = workingInputLength - tagLength;
-      final byte[] result = new byte[expectedOutputLength];
```
geedo0, line 701: This is where the memory gets saved. Previously, when the caller provides the buffer this intermediate buffer gets allocated no matter what.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -675,31 +680,21 @@ private byte[] engineDecryptFinal(byte[] input, final int offset, final int leng
...
       final byte[] workingInputArray;
       final int workingInputOffset;
       final int workingInputLength;
-      if (decryptInputBuf.size() > 0) {
-        // We've already processed data so need to merge this in
-        engineUpdate(input, offset, length); // Decrypt mode never generates output for updates
-        workingInputArray = decryptInputBuf.getDataBuffer();
-        workingInputLength = decryptInputBuf.size();
-        workingInputOffset = 0;
-      } else {
```
geedo0, line 690: This is a small step backwards in favor of making things more uniform. Previously, we were able to get away with using the input buffer directly when the client calls doFinal using the byte[] interfaces and without calling doUpdate prior. With this change, internally buffering the input on decryption operations is now compulsory. If we have telemetry to suggest that this particular call pattern is common enough, we can try to add some logic to accommodate it. FWIW, Sun JSSE's TLS implementation calls into our library via the ByteBuffer interface where this has no impact.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -773,42 +762,37 @@ private byte[] engineDecryptFinal(byte[] input, final int offset, final int leng
...
-        throw new AssertionError(e);
-      }
-
-      if (actualLength == buf.length) {
-        return buf;
-      } else {
-        // This branch should never happen but is technically allowed by the underlying APIs.
-        // So, we cover it just in case.
-        return Arrays.copyOf(buf, actualLength);
+    final byte[] buf = new byte[engineGetOutputSize(length)];
```
geedo0, line 765: Allocation of the output buffer now happens here only in the "ACCP allocates the output buffer" interface. We avoid allocating the output buffer in the other two "caller allocates" interfaces.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -971,17 +955,22 @@ protected int engineDoFinal(final ByteBuffer input, final ByteBuffer output)
     return output.position() - initialPosition;
   }
 
-  private void checkOutputBuffer(final int inputLength, final byte[] output, final int outputOffset)
```
geedo0, line 974: These changes are all covered in the other PR. I couldn't figure out if Github supports commit slicing for PR.

https://github.com/corretto/amazon-corretto-crypto-provider/pull/297

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -756,15 +748,12 @@ private byte[] engineDecryptFinal(byte[] input, final int offset, final int leng
...
-        return result;
-      } else if (outLen < result.length) {
-        // While technically allowed by the APIs, this should never happen.
-        return Arrays.copyOf(result, outLen);
-      } else {
-        throw new RuntimeCryptoException("Output longer than expected.");
-      }
+      return outLen;
+    } catch (AEADBadTagException e) {
+      int maxFillSize = output.length - outputOffset;
```
WillChilds-Klein, line 753: minor nit: `final`

WillChilds-Klein (approved):

amirhosv (approved):
What's the advantage of overriding the methods that process ByteBuffer? Those methods are not abstract and JDK provides [an implementation for them] (https://github.com/openjdk/jdk/blob/jdk8-b120/jdk/src/share/classes/javax/crypto/CipherSpi.java#L744)

PR #300, opened 2023-05-22
Optimize handling of overlapping input/output buffers for AES-GCM

geedo0:
*Issue #, if available:*
#271

*Description of changes:*
Currently, ACCP does not tolerate any overlap between input/output buffers. If it detects any overlap, it will buffer a copy of the entire input prior to proceeding (consuming O(n) more memory). This is to prevent any risk of the output of the operation from overwriting the input prior to processing that input. However, it is a common case for our clients (e.g. Sun JSSE's TLS implementation) to preserve memory by reusing the input/output buffers since they have little use for the input once it has been successfully encrypted/decrypted.

AES, as a block cipher, operates on blocks. It reads one block of input into intermediate memory for processing and then writes one block of output. Implementations of AES can tolerate writing the output to the exact same block of memory where it read the input. Where we run into trouble is whenever the output block overwrites data from any input block beyond the block that was just processed.

This changeset relaxes the boundary checking in the input/output arrays so that it allows the case where we overlap exactly or any case where the output lags behind the input. We still safely handle cases where the output overwrites the input by detecting and cloning the input buffer as necessary. We also check for cases where the output leads the input, but beyond the limit of the input buffer and allow this configuration without cloning memory. The end result is that we unlock another 30% improvement in memory consumption under the benchmarks in issue #271.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
@@ -434,40 +434,39 @@ final class AesGcmSpi extends CipherSpi {
   }
 
   @Override
-  protected int engineUpdate(byte[] bytes, int offset, int length, byte[] output, int outputOffset)
+  protected int engineUpdate(
+      byte[] input, int inputOffset, int inputLen, byte[] output, int outputOffset)
       throws ShortBufferException {
-    checkArrayLimits(bytes, offset, length);
+    checkArrayLimits(input, inputOffset, inputLen);
 
     hasConsumedData = true;
 
     switch (opMode) {
       case NATIVE_MODE_DECRYPT:
         {
-          decryptInputBuf.write(bytes, offset, length);
+          decryptInputBuf.write(input, inputOffset, inputLen);
           return 0;
         }
       case NATIVE_MODE_ENCRYPT:
         {
-          checkOutputBuffer(length, output, outputOffset, false);
+          checkOutputBuffer(inputLen, output, outputOffset, false);
 
           lazyInit();
 
           // If we have an overlap, we'll need to clone the input buffer before we potentially start
-          // overwriting
-          // it.
+          // overwriting it.
           final byte[] finalBytes;
           final int finalOffset;
-          if (Utils.arraysOverlap(
-              bytes, offset, output, outputOffset, engineGetOutputSize(length))) {
-            finalBytes = Arrays.copyOfRange(bytes, offset, offset + length);
+          if (Utils.outputClobbersInput(input, inputOffset, inputLen, output, outputOffset)) {
+            finalBytes = Arrays.copyOfRange(input, inputOffset, inputOffset + inputLen);
             finalOffset = 0;
           } else {
-            finalBytes = bytes;
-            finalOffset = offset;
+            finalBytes = input;
+            finalOffset = inputOffset;
           }
 
           return context.use(
-              ptr -> encryptUpdate(ptr, finalBytes, finalOffset, length, output, outputOffset));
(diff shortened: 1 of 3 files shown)


src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -916,10 +905,8 @@ protected int engineUpdate(ByteBuffer input, final ByteBuffer output)
           throw new ShortBufferException();
         }
 
-        if (Utils.buffersMaybeOverlap(input, output)) {
-          // We'll just copy the whole input buffer if it might overlap with output. It's possible
-          // to do something more efficient for a couple of special cases here, but we'll do the
-          // simple and safe thing for now.
+        if (Utils.outputClobbersInput(input, output)) {
+          // We'll just copy the whole input buffer if it might overlap with output.
           ByteBuffer newInput = ByteBuffer.allocate(input.remaining());
```
geedo0, line 910: This is where we get our memory savings. That ByteBuffer allocation is now avoided for the common case of overlapping buffers.

src/com/amazon/corretto/crypto/provider/Utils.java
```diff
@@ -97,75 +97,77 @@ static int getMdLen(final long mdPtr) {
...
-      // different arrays, so no overlap
-      return false;
-    }
-
-    // Same array, check for overlap within array
-    long a_offset = (long) a.arrayOffset() + (long) a.position();
-    long b_offset = (long) b.arrayOffset() + (long) b.position();
-
-    if (a_offset > b_offset) {
-      return b_offset + (long) b.limit() > a_offset;
```
geedo0, line 146: This was the problematic logic which would trigger an allocation for all cases of the output being beyond the start of the input.

src/com/amazon/corretto/crypto/provider/Utils.java
```diff
@@ -97,75 +97,77 @@ static int getMdLen(final long mdPtr) {
   /**
    * Returns false if the two bytebuffers given definitely don't overlap; true if they do overlap,
    * or if we're unable to determine whether they overlap. Unfortunately it's not possible to
```
WillChilds-Klein, line 99: should we re-word this to be explicit that we allow for non-clobbering overlap?

tst/com/amazon/corretto/crypto/provider/test/UtilsTest.java
```diff
@@ -34,12 +34,11 @@ public class UtilsTest {
   }
 
   boolean maybeOverlaps(ByteBuffer a, ByteBuffer b) throws Throwable {
```
WillChilds-Klein, line 36: nit: update name to reflect underlying method name change?

tst/com/amazon/corretto/crypto/provider/test/UtilsTest.java
```diff
@@ -61,8 +59,8 @@ private void assertNoOverlap(ByteBuffer a, ByteBuffer b) throws Throwable {
     assertNoOverlap0(a.slice(), b.slice());
   }
 
-  boolean arraysOverlap(byte[] a1, int o1, byte[] a2, int o2, int length) throws Throwable {
-    return (Boolean) sneakyInvoke(UTILS_CLASS, "arraysOverlap", a1, o1, a2, o2, length);
+  boolean arraysOverlap(byte[] a1, int o1, int length, byte[] a2, int o2) throws Throwable {
```
WillChilds-Klein, line 62: nit: update name to reflect underlying method name change?

WillChilds-Klein (approved):
very nice!

amirhosv (commented):
Great work with updating the names. It's much more readable now.

amirhosv (approved):

PR #302, opened 2023-05-25
Avoid overriding CipherSpi::engineDoFinal(ByteBuffer, ByteBuffer)

amirhosv:
*Description of changes:*

CipherSpi, the parent class of AesGcmSpi provides an implementation for engineDoFinal(ByteBuffer, ByteBuffer) that has a smaller memory footprint.

For benchmarking, I used [ckozak/accp branch of java-crypto-allocation-performance](https://github.com/carterkozak/java-crypto-allocation-performance/tree/ckozak/accp). Here is section of the report:

Baseline:
```
Benchmark                                                                               (cipher)   (length)  (provider)  Mode  Cnt          Score           Error   Units
DirectBufferTransportLayerSecurityBenchmark.download                      TLS_AES_256_GCM_SHA384  104857600       ACCP2  avgt    4        161.150 ±        10.875   ms/op
DirectBufferTransportLayerSecurityBenchmark.download:·gc.alloc.rate       TLS_AES_256_GCM_SHA384  104857600       ACCP2  avgt    4       1048.520 ±      4310.095  MB/sec
DirectBufferTransportLayerSecurityBenchmark.download:·gc.alloc.rate.norm  TLS_AES_256_GCM_SHA384  104857600       ACCP2  avgt    4  177981662.606 ± 732218228.614    B/op
DirectBufferTransportLayerSecurityBenchmark.download:·gc.count            TLS_AES_256_GCM_SHA384  104857600       ACCP2  avgt    4         90.000                  counts
DirectBufferTransportLayerSecurityBenchmark.download:·gc.time             TLS_AES_256_GCM_SHA384  104857600       ACCP2  avgt    4         80.000                      ms
```

With this PR:
```
Benchmark                                                                               (cipher)   (length)  (provider)  Mode  Cnt          Score           Error   Units
DirectBufferTransportLayerSecurityBenchmark.download                      TLS_AES_256_GCM_SHA384  104857600       ACCP2  avgt    4        135.932 ±        14.041   ms/op
DirectBufferTransportLayerSecurityBenchmark.download:·gc.alloc.rate       TLS_AES_256_GCM_SHA384  104857600       ACCP2  avgt    4        772.883 ±      3156.272  MB/sec
DirectBufferTransportLayerSecurityBenchmark.download:·gc.alloc.rate.norm  TLS_AES_256_GCM_SHA384  104857600       ACCP2  avgt    4  110326939.149 ± 450469639.448    B/op
DirectBufferTransportLayerSecurityBenchmark.download:·gc.count            TLS_AES_256_GCM_SHA384  104857600       ACCP2  avgt    4         66.000                  counts
DirectBufferTransportLayerSecurityBenchmark.download:·gc.time             TLS_AES_256_GCM_SHA384  104857600       ACCP2  avgt    4         55.000                      ms
```

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
@@ -926,22 +926,6 @@ final class AesGcmSpi extends CipherSpi {
     }
   }
 
-  @Override
-  protected int engineDoFinal(final ByteBuffer input, final ByteBuffer output)
-      throws ShortBufferException, IllegalBlockSizeException, BadPaddingException {
-    int initialPosition = output.position();
-
-    engineUpdate(input, output);
-
-    ShimArray shim = new ShimArray(output, engineGetOutputSize(0));
-    int finalBytes = engineDoFinal(EMPTY_ARRAY, 0, 0, shim.array, shim.offset);
-
-    shim.writeback();
-    output.position(output.position() + finalBytes);
-
-    return output.position() - initialPosition;
-  }
-
   private void checkOutputBuffer(
       final int inputLength, final byte[] output, final int outputOffset, final boolean doFinal)
       throws ShortBufferException {


src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -926,22 +926,6 @@ protected int engineUpdate(ByteBuffer input, final ByteBuffer output)
...
-  protected int engineDoFinal(final ByteBuffer input, final ByteBuffer output)
-      throws ShortBufferException, IllegalBlockSizeException, BadPaddingException {
-    int initialPosition = output.position();
-
-    engineUpdate(input, output);
-
-    ShimArray shim = new ShimArray(output, engineGetOutputSize(0));
-    int finalBytes = engineDoFinal(EMPTY_ARRAY, 0, 0, shim.array, shim.offset);
-
-    shim.writeback();
```
geedo0, line 939: Is avoiding this ShimArray stuff where we got our performance gain?

geedo0 (approved):

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
WillChilds-Klein: we [do an O(n) copy on ShimArray construction](https://github.com/corretto/amazon-corretto-crypto-provider/blob/96fdeef44e41387f136b829f79068dc9f52b9510/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java#L844) if the input buffer is not backed by an "accessible byte[]". in that same case, we do another full O(n) copy on `writeback()`. if the buffer is backed by an array, then we [don't do the initial copy](https://github.com/corretto/amazon-corretto-crypto-provider/blob/96fdeef44e41387f136b829f79068dc9f52b9510/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java#L847-L850) and we [no-op on writeback](https://github.com/corretto/amazon-corretto-crypto-provider/blob/96fdeef44e41387f136b829f79068dc9f52b9510/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java#L840-L858).

JDK 11 also appears to do a [single O(n) copy if either the input or output buffers are not backed by an array](https://github.com/corretto/corretto-11/blob/develop/src/java.base/share/classes/javax/crypto/CipherSpi.java#L787-L829) (`get()` in a loop on L787, `put()` in a loop on L829) and [two O(n) copies](https://github.com/corretto/corretto-11/blob/develop/src/java.base/share/classes/javax/crypto/CipherSpi.java#L820-L829) if neither input nor output is backed by an accessible array. if both input and output buffers are backed by an array, then there are [no extra copies](https://github.com/corretto/corretto-11/blob/develop/src/java.base/share/classes/javax/crypto/CipherSpi.java#L769-L781).

so, it looks like ShimArray w/ the `writeback` call does one more O(n) allocation + copy than JDK11's implementation (2 allocations + copies vs. 1) in the case where the output buffer is not backed by an accessible array but the input buffer is.

WillChilds-Klein (approved):
nice find!

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
amirhosv: It's the combination of not chunking up the input, which results in fewer JNI calls when encrypting, and avoiding the creation of ShimArray objects.

WillChilds-Klein: i read through the code a bit, and it looks like when the input buffer is backed by an accessible array but the output buffer is not, OpenJDK does one fewer [allocation + copy](https://github.com/corretto/corretto-11/blob/develop/src/java.base/share/classes/javax/crypto/CipherSpi.java#L812) than [we do with the `new ShimArray()` and `writeback()`](https://github.com/corretto/amazon-corretto-crypto-provider/blob/96fdeef44e41387f136b829f79068dc9f52b9510/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java#L844-L858) calls (OpenJDK does 1, we do 2).

PR #304, opened 2023-06-05
Serializable Keys

SalusaSecondus:
*Description of changes:*

Java specifies that implementations of [Key are Serializable](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/security/Key.html). Currently ACCP does not allow its keys to be serialized due to the difficulties in handling pointers into native memory and fixing them up while still letting them be `final`. This change safely serializes and deserializes instances of `EvpKey` by using a secondary helper class. A code comment explaining how this works is duplicated below.

> Java Keys are Serializable but we cannot use the trivial serialization of an EvpKey.
> - They contain a pointer to native memory (and associated native memory).
>   This obviously cannot remain valid after being serialized/deserialized.
> - The pointer to native memory is final to ensure there is no risk of it being invalid.
>   This means that we cannot just fix it up when deserializing.
> 
> The trivial solution would be to make the pointer non-final, but that introduces a greater risk of serious bugs.
> 
> Instead, we use writeReplace() and readResolve() to store our information in a dedicated format for serialization.
> - writeReplace() on the object to be serialized (an EvpKey) returns the *real* object to be serialized.
>    In our case, we return an instance of SerializedKey which contains the minimal information to properly save and retrieve a key.
> - readResolve() is called on the object which is *actually* serialized (and thus being deserialized) and returns the object that users care about.
>   In our case, SeralizedKey.readResolve() returns an appropriate instance of EvpKey.
> 
> So, the entire (hidden from the user) flow goes like this.
> 1.  User tries to serialize an EvpKey
> 2.  Java detects that EvpKey.writeReplace() exists and calls it
> 3.  EvpKey.writeReplace() returns an instance of SerializedKey
> 4.  Java actually serializes SerializedKey
> 5.  (Some time passes)
> 6.  User tries to deserialize the bytes
> 7.  Java detects that the bytes contain a serialized SerializedKey and so deserializes it.
> 8.  Java detects that SerializedKey.readResolve() exists and calls it
> 9.  SerializedKey.readResolve() creates and returns an instance of EvpKey
> 10. User gets an instance of EvpKey and never realizes that the above complexity exists

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/EvpKey.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpKey.java
@@ -278,7 +278,7 @@ abstract class EvpKey implements Key, Destroyable {
     private final boolean isPublicKey;
     private final byte[] encoded;
 
-    public SerializedKey(EvpKeyType type, boolean isPublicKey, byte[] encoded) {
+    public SerializedKey(final EvpKeyType type, final boolean isPublicKey, final byte[] encoded) {
       this.type = type;
       this.isPublicKey = isPublicKey;
       this.encoded = encoded;
@@ -287,13 +287,11 @@ abstract class EvpKey implements Key, Destroyable {
     private Object readResolve() throws ObjectStreamException {
       try {
         final KeyFactory kf = AmazonCorrettoCryptoProvider.INSTANCE.getKeyFactory(type);
-        final EvpKey result;
         if (isPublicKey) {
-          result = (EvpKey) kf.generatePublic(new X509EncodedKeySpec(encoded));
+          return kf.generatePublic(new X509EncodedKeySpec(encoded));
         } else {
-          result = (EvpKey) kf.generatePrivate(new PKCS8EncodedKeySpec(encoded));
+          return kf.generatePrivate(new PKCS8EncodedKeySpec(encoded));
         }
-        return result;
       } catch (final InvalidKeySpecException ex) {
         throw new InvalidObjectException(ex.getMessage());
       }
--- a/src/com/amazon/corretto/crypto/provider/EvpKey.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpKey.java
@@ -3,17 +3,23 @@
 package com.amazon.corretto.crypto.provider;
 
 import java.io.IOException;
+import java.io.InvalidObjectException;
 import java.io.NotSerializableException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.ObjectStreamException;
+import java.io.Serializable;
 import java.math.BigInteger;
 import java.security.AlgorithmParameters;
 import java.security.GeneralSecurityException;
 import java.security.Key;
+import java.security.KeyFactory;
 import java.security.MessageDigest;
 import java.security.PublicKey;
 import java.security.spec.AlgorithmParameterSpec;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.security.spec.X509EncodedKeySpec;
 import java.util.Arrays;
 import java.util.Base64;
 import javax.security.auth.Destroyable;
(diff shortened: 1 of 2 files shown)


WillChilds-Klein (commented):
Thanks for the PR @SalusaSecondus !

amirhosv:
I'm not sure if it's a good idea to support or advocate this. Java's serialization mechanism is not the most suitable approach for serialization and it's best for customers to use more standard crypto formats like ASN1 encodings for such requirements.
Also, not all providers need to implement a meaningful implementation for serialization; for example, SunPKCS11 provider that allows applications use keys backed by HSM does not (and  cannot provide) meaningful implementation.

The problem with supporting this is it's maintenance can be become a hassle in future and it's best to push customers to avoid using this and rely on more standard formats.

src/com/amazon/corretto/crypto/provider/EvpKey.java
```diff
@@ -216,24 +222,81 @@ protected static class InternalKey extends NativeResource {
...
+   * 3.  EvpKey.writeReplace() returns an instance of SerializedKey
+   * 4.  Java actually serializes SerializedKey
+   * 5.  (Some time passes)
+   * 6.  User tries to deserialize the bytes
+   * 7.  Java detects that the bytes contain a serialized SerializedKey and so deserializes it.
+   * 8.  Java detects that SerializedKey.readResolve() exists and calls it
+   * 9.  SerializedKey.readResolve() creates and returns an instance of EvpKey
+   * 10. User gets an instance of EvpKey and never realizes that the above complexity exists
+   */
+  Object writeReplace() throws ObjectStreamException {
```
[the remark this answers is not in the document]
SalusaSecondus: I think that we cannot make it private. From the [Serializable documentation](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/io/Serializable.html) (emphasis mine):

> This writeReplace method is invoked by serialization if the method exists and it would be **accessible from a method defined within the class of the object being serialized**.

Thus, if we make it `private`, then it isn't accessible when we need to serialize any of the concrete implementations of `EvpKey`.

SalusaSecondus:
@amirhosv I completely agree that people shouldn't serialize keys in this way. However, it is a way in which ACCP differs from the providers bundled with the JDK. The JDK's tests assume that keys are serializable (and so fail for meaningless reasons). Because this solution is built off of existing standardized encodings, the ongoing maintenance cost should be minimal.

So, fixing this makes it so that ACCP is even more of a transparent drop in replacement than it was before.

amirhosv:
> @amirhosv I completely agree that people shouldn't serialize keys in this way. However, it is a way in which ACCP differs from the providers bundled with the JDK. The JDK's tests assume that keys are serializable (and so fail for meaningless reasons). Because this solution is built off of existing standardized encodings, the ongoing maintenance cost should be minimal.
> 
> So, fixing this makes it so that ACCP is even more of a transparent drop in replacement than it was before.

Could you share the tests that are failing in JDK?

brian-jarvis-aws:
Hey @SalusaSecondus, just wanted to give a little more context here. We're trying to get a better understanding of what the use case is for this. We aren't necessarily opposed to bringing it in; as you noted, it shouldn't add a lot of ongoing maintenance. However, a tenet of ACCP is that it is opinionated and I see from the correspondence that both you and @amirhosv are of the opinion that people should not serialize keys in this way. As such, I would expect that the justification for this should outweigh the enablement of what both of you describe as an anti-pattern (my paraphrasing). So, we're just wondering if you could share more about the motivation and use case so we can better understand what the problem is with the current solution.

WillChilds-Klein:
It sounds like merging this PR would unblock execution of JDK's JCE tests against ACCP, which would provide useful data on where we stand in terms of compatibility with the default JCE provider(s). Assuming that the maintenance cost of EvpKey's serialization format through time is low (which I agree that it is very likely to be given PKCS8/x509 formats' stability), I'm OK with merging this change.

WillChilds-Klein (approved):

src/com/amazon/corretto/crypto/provider/EvpKey.java
```diff
@@ -216,24 +222,81 @@ protected interface CanDerivePublicKey<T extends EvpKey & PublicKey> {
...
+      this.type = type;
+      this.isPublicKey = isPublicKey;
+      this.encoded = encoded;
+    }
+
+    private Object readResolve() throws ObjectStreamException {
+      try {
+        final KeyFactory kf = AmazonCorrettoCryptoProvider.INSTANCE.getKeyFactory(type);
+        final EvpKey result;
+        if (isPublicKey) {
```
amirhosv, line 291: Is the type cast needed?

src/com/amazon/corretto/crypto/provider/EvpKey.java
```diff
@@ -216,24 +222,81 @@ protected interface CanDerivePublicKey<T extends EvpKey & PublicKey> {
...
+   * Minimal information needed to serialize/deserialize any instance of EvpKey. Contains the key
+   * type, whether it is a public key, and the appropriate encoding of it.
+   */
+  private static class SerializedKey implements Serializable {
+    private static final long serialVersionUID = 1;
+    private final EvpKeyType type;
+    private final boolean isPublicKey;
+    private final byte[] encoded;
+
+    public SerializedKey(EvpKeyType type, boolean isPublicKey, byte[] encoded) {
```
amirhosv, line 281: Arguments can be marked final.

src/com/amazon/corretto/crypto/provider/EvpKey.java
SalusaSecondus: Needed? No. Present to make the return type slightly clearer (in that it *must* assignable to an `EvpKey`.) I'll change it to `Key` though which is the next best type.

amirhosv: or just `return isPublicKey ? kf.generatePublic(new X509EncodedKeySpec(encoded)) : kf.generatePrivate(new PKCS8EncodedKeySpec(encoded));`

amirhosv:
The mac CI task failure has nothing to do with this PR. [This PR](https://github.com/corretto/amazon-corretto-crypto-provider/pull/306) will resolve those. Once [PR 306](https://github.com/corretto/amazon-corretto-crypto-provider/pull/306) is merged, please consider rebasing.

amirhosv (approved):

brian-jarvis-aws (approved):

PR #306, opened 2023-06-15
AES-XTS support

amirhosv:
*Description of changes:*

* Cipher service for AES-XTS
* Only supports AES 256 bit keys. For XTS, two AES keys are needed that are passed to cipher as a single SecretKey object.
* Tweak must be provided as an IvParameterSpec
* Multi-step encryption/decryption is not supported since LC does not support that.

### DIFF
--- a/csrc/aes_xts.cpp
+++ b/csrc/aes_xts.cpp
@@ -1,0 +1,157 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+#include "buffer.h"
+#include "env.h"
+#include <openssl/evp.h>
+#include <jni.h>
+
+// Tweak and key are passed in the same buffer. The first 16 bytes
+// of this buffer is the tweak and the rest is the key.
+static int const AES_XTS_KEY_INDEX_START = 16;
+
+namespace AmazonCorrettoCryptoProvider {
+
+class AesXtsCipher {
+public:
+    AesXtsCipher(bool for_encryption, unsigned char const* key, unsigned char const* tweak)
+    {
+        ctx_ = EVP_CIPHER_CTX_new();
+        if (ctx_ == nullptr) {
+            throw_openssl(EX_RUNTIME_CRYPTO, "EVP_CIPHER_CTX_new failed.");
+        }
+
+        if (for_encryption) {
+            if (EVP_EncryptInit_ex(ctx_, EVP_aes_256_xts(), nullptr, key, tweak) != 1) {
+                throw_openssl(EX_RUNTIME_CRYPTO, "EVP_EncryptInit_ex failed.");
+            }
+        } else {
+            if (EVP_DecryptInit_ex(ctx_, EVP_aes_256_xts(), nullptr, key, tweak) != 1) {
+                throw_openssl(EX_RUNTIME_CRYPTO, "EVP_DecryptInit_ex failed.");
+            }
+        }
+
+        // this method always returns 1
+        if (EVP_CIPHER_CTX_set_padding(ctx_, 0) != 1) {
+            throw_openssl(EX_RUNTIME_CRYPTO, "EVP_CIPHER_CTX_set_padding");
+        }
+    }
+
+    ~AesXtsCipher() { EVP_CIPHER_CTX_free(ctx_); }
+
+    void encrypt(unsigned char* input, int input_len, unsigned char* output)
+    {
+        int out_len = 0;
+        if (EVP_EncryptUpdate(ctx_, output, &out_len, input, input_len) != 1) {
+            throw_openssl(EX_RUNTIME_CRYPTO, "EVP_EncryptUpdate failed.");
+        }
+
+        if (EVP_EncryptFinal_ex(ctx_, output + out_len, &out_len) != 1) {
+            throw_openssl(EX_RUNTIME_CRYPTO, "EVP_EncryptFinal_ex failed.");
+        }
+    }
+
+    void decrypt(unsigned char* input, int input_len, unsigned char* output)
+    {
+        int out_len = 0;
+        if (EVP_DecryptUpdate(ctx_, output, &out_len, input, input_len) != 1) {
+            throw_openssl(EX_RUNTIME_CRYPTO, "EVP_DecryptUpdate failed.");
+        }
+
+        if (EVP_DecryptFinal_ex(ctx_, output + out_len, &out_len) != 1) {
+            throw_openssl(EX_RUNTIME_CRYPTO, "EVP_DecryptFinal_ex failed.");
+        }
+    }
+
+private:
+    EVP_CIPHER_CTX* ctx_;
+};
+
+}
+
+using namespace AmazonCorrettoCryptoProvider;
+
+extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXtsSpi_enc(JNIEnv* env,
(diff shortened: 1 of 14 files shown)


csrc/aes_xts.cpp
```diff
@@ -0,0 +1,181 @@
...
+    {
+        int out_len = 0;
+        if (EVP_DecryptUpdate(ctx_, output, &out_len, input, input_len) != 1) {
+            throw_openssl(EX_RUNTIME_CRYPTO, "EVP_DecryptUpdate failed.");
+        }
+
+        if (EVP_DecryptFinal_ex(ctx_, output + out_len, &out_len) != 1) {
+            throw_openssl(EX_RUNTIME_CRYPTO, "EVP_DecryptFinal_ex failed.");
+        }
+    }
```
brian-jarvis-aws, line 56: Should `encrypt`'s input and `decrypt`'s output pointers be const? The EVP functions being called use that convention in their own calling signatures, and you did in the constructor as well.

amirhosv: The reason it's not const is that the same buffer for input can be used as output.

brian-jarvis-aws: Ack. Thanks for the clarification.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -77,6 +77,8 @@ private void buildServiceMap() {
 
     addService("KeyGenerator", "AES", "keygeneratorspi.SecretKeyGenerator", false);
 
+    addService("Cipher", "AES/XTS/NoPadding", "AesXtsSpi", false);
```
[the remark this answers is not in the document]
amirhosv: The 256 limit is due to AWS-LC only supporting 256.  I don't know of any JCA providers that supports XTS. If in future we need to support other key sizes, we can still use the same name `AES/XTS/NoPadding` and let the cipher determine the key size based on the key that is passed during initialization of the cipher.

src/com/amazon/corretto/crypto/provider/AesXtsSpi.java
```diff
@@ -0,0 +1,293 @@
...
+    cipherState = cipherState.nextStateAfterDoFinal();
+    return actualOutputLen;
+  }
+
+  private void encrypt(
+      byte[] input, int inputOffset, int inputLen, byte[] output, int outputOffset) {
+    if (input != output) {
+      enc(packedTweakKey, input, inputOffset, inputLen, output, outputOffset);
+    } else {
+      encSameBuffer(packedTweakKey, input, inputOffset, inputLen, outputOffset);
```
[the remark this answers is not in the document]
amirhosv: It handles it correctly but in this way, we can reduce the number of `GetPrimitiveArrayCritical/ReleasePrimitiveArrayCritical` invocations by one since they are the same buffer.

csrc/buffer.cpp
```diff
@@ -35,4 +35,18 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
     return array;
 }
 
+JByteArrayCritical::JByteArrayCritical(JNIEnv* env, jbyteArray jarray)
```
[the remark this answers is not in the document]
amirhosv: That makes sense. I'll update the comment.

src/com/amazon/corretto/crypto/provider/AesXtsSpi.java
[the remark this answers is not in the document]
WillChilds-Klein: got it. are the `{Get,Release}PrimitiveArrayCritical` calls expensive? my understanding is that these just return a view on memory and [manage some lock state/object pinning](https://github.com/corretto/corretto-11/blob/f4cc07566cf589fac42893ed8d926d4fae23b087/src/hotspot/share/prims/jni.cpp#L3190), but don't do any allocation.

tst/com/amazon/corretto/crypto/provider/test/AesXtsTest.java
```diff
@@ -0,0 +1,605 @@
...
+    }
+  }
+
+  @Test
+  public void xtsTestsFromAwsLCPositive() {
+    final Stream<AesXtsTestCase> tests =
+        Stream.of(
+            new AesXtsTestCase()
+                .addKey("2718281828459045235360287471352662497757247093699959574966967627314159")
+                .addKey("2653589793238462643383279502884197169399375105820974944592")
```
WillChilds-Klein, line 129: why use multiple builder calls for the key + tweak which are fixed size?

WillChilds-Klein (dismissed):
nice. very clean implementation :)

tst/com/amazon/corretto/crypto/provider/test/AesXtsTest.java
amirhosv: It helps with breaking long lines. It's purely cosmetic :)

csrc/aes_xts.cpp
```diff
@@ -0,0 +1,145 @@
...
+    EVP_CIPHER_CTX* ctx_;
+};
+
+}
+
+using namespace AmazonCorrettoCryptoProvider;
+
+extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXtsSpi_enc(JNIEnv* env,
+    jclass,
+    jbyteArray jtweakKey,
```
alexw91, line 73: Do we need to pack both keys into 1 jbyteArray here? Why not have two separate jbyteArrays? Is this just to avoid an extra `JByteArrayCritical` call?

Also, naming nit, if this byte array contains two different keys, you should follow the `packedTweakKey` naming convention that was used Java-side here as well.

tests/ci/run_accp_basic_tests_mac.sh
```diff
@@ -0,0 +1,38 @@
...
+	exit $?
+fi
+
+# Assign the JDK version we're testing as the system's default JDK and
+# assign JAVA_HOME variable to the path. Otherwise, Ubuntu will
+# default to the newest version of Java on the system.
+export JAVA_HOME=$TEST_JAVA_HOME
+export PATH=$JAVA_HOME/bin:$PATH
+
+./gradlew -DFIPS=$testing_fips -DLCOV_IGNORE=inconsistent release
```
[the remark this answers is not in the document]
amirhosv: done

csrc/aes_xts.cpp
```diff
@@ -0,0 +1,145 @@
...
+                throw_openssl(EX_RUNTIME_CRYPTO, "EVP_EncryptInit_ex failed.");
+            }
+        } else {
+            if (EVP_DecryptInit_ex(ctx_, EVP_aes_256_xts(), nullptr, key, tweak) != 1) {
+                throw_openssl(EX_RUNTIME_CRYPTO, "EVP_DecryptInit_ex failed.");
+            }
+        }
+
+        // this method always returns 1
+        EVP_CIPHER_CTX_set_padding(ctx_, 0);
```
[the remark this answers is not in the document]
amirhosv: done

csrc/aes_xts.cpp
amirhosv: It's just to avoid creating one extra `JByteArrayCritical`.
Fixed the naming.

brian-jarvis-aws (approved):

WillChilds-Klein (approved):

alexw91 (approved):

PR #310, opened 2023-07-12
HKDF

amirhosv:
## Support for HKDF as defined in RFC5869
* This implementation delegates the operation to [AWS-LC](https://github.com/aws/aws-lc/blob/main/include/openssl/hkdf.h).
* HKDF is exposed as a [SecretKeyFactory service](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/javax/crypto/SecretKeyFactory.html).
* The client code needs to invoke `SecretKeyFactory.getInstance("HkdfWith<digest>")` to get an instance of HKDF.
  * The possible values for `<digest>` are `SHA-1, SHA-256, SHA-384, SHA-512`.
* The output of HKDF is computed by invoking [SecretKeyFactory::generateSecret](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/javax/crypto/SecretKeyFactory.html#generateSecret(java.security.spec.KeySpec)).
* The input to [SecretKeyFactory::generateSecret](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/javax/crypto/SecretKeyFactory.html#generateSecret(java.security.spec.KeySpec)) must be an instance of `HkdfRfc5869Spec` class.
  * `HkdfSpec` provides three factory constructors, `hkdfSpec, hkdfExtractSpec, hkdfExpandSpec` for convenience.
  * Each factory constructor requires the necessary inputs to the the corresponding algorithms, HKDF, HKDF_Extract, and HKDF_Expand.
 * For complete sample usage, please refer to  `Hkdf::sampleUsage1` and  `Hkdf::sampleUsage2`.


src/com/amazon/corretto/crypto/provider/HkdfRfc5869SecretKeyFactorySpi.java
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  static final Map<String, HkdfRfc5869SecretKeyFactorySpi> INSTANCES = getInstances();
+
+  private static final String HKDF_RFC5869 = "HkdfRfc5869";
+  private static final String WITH = "With";
+  static final String HKDF_RFC5869_WITH_SHA1 = HKDF_RFC5869 + WITH + "SHA-1";
+  static final String HKDF_RFC5869_WITH_SHA256 = HKDF_RFC5869 + WITH + "SHA-256";
+  static final String HKDF_RFC5869_WITH_SHA384 = HKDF_RFC5869 + WITH + "SHA-384";
+  static final String HKDF_RFC5869_WITH_SHA512 = HKDF_RFC5869 + WITH + "SHA-512";
```
geedo0, line 125: Should we be using hyphenated forms for the hash algorithm? The standard signature algorithms are unhyphenated e.g. ` SHA256withRSA`.

src/com/amazon/corretto/crypto/provider/HkdfRfc5869Spec.java
```diff
@@ -0,0 +1,152 @@
...
+  final int mode;
+  final byte[] secretOrPrk;
+  final byte[] salt;
+  final byte[] info;
+  final int desiredLength;
+  final String algorithmName;
+  public static final String DEFAULT_ALGORITHM_NAME = "HkdfRfc5869";
+  public static final int HKDF_MODE = 1;
+  public static final int HKDF_EXTRACT_MODE = 2;
+  public static final int HKDF_EXPAND_MODE = 3;
```
geedo0, line 17: Why not an Enum? Is this a JCA convention?

src/com/amazon/corretto/crypto/provider/HkdfRfc5869SecretKeyFactorySpi.java
```diff
@@ -0,0 +1,143 @@
...
+            resultBytes.length,
+            digestCode,
+            spec.secretOrPrk,
+            spec.secretOrPrk.length,
+            spec.salt,
+            spec.salt.length);
+        break;
+      default:
+        // spec.mode is HKDF_EXPAND_MODE. We already checked for it in the constructor of
+        // HkdfRfc5869Spec
```
geedo0, line 62: Why not just be explicit and throw on the default case? Comment may be true today, but who knows what junk I'll check-in tomorrow.

csrc/hkdf.cpp
```diff
@@ -0,0 +1,114 @@
...
+static EVP_MD const* digest_code_to_EVP_MD(int digestCode)
+{
+    switch (digestCode) {
+    case 1:
+        return EVP_sha1();
+    case 2:
+        return EVP_sha256();
+    case 3:
+        return EVP_sha384();
+    default:
```
geedo0, line 21: Should we be explicit and throw rather than default to 512?

src/com/amazon/corretto/crypto/provider/HkdfRfc5869SecretKeyFactorySpi.java
```diff
@@ -0,0 +1,143 @@
...
+  static final String HKDF_RFC5869_WITH_SHA1 = HKDF_RFC5869 + WITH + "SHA-1";
+  static final String HKDF_RFC5869_WITH_SHA256 = HKDF_RFC5869 + WITH + "SHA-256";
+  static final String HKDF_RFC5869_WITH_SHA384 = HKDF_RFC5869 + WITH + "SHA-384";
+  static final String HKDF_RFC5869_WITH_SHA512 = HKDF_RFC5869 + WITH + "SHA-512";
+
+  private static Map<String, HkdfRfc5869SecretKeyFactorySpi> getInstances() {
+    final Map<String, HkdfRfc5869SecretKeyFactorySpi> result = new HashMap<>();
+    result.put(
+        HKDF_RFC5869_WITH_SHA1.toUpperCase(),
+        new HkdfRfc5869SecretKeyFactorySpi(1, getDigestLen(1)));
```
geedo0, line 131: These digest codes seem somewhat arbitrary, it'd be more readable to define some kind of aliasing for them.

src/com/amazon/corretto/crypto/provider/HkdfRfc5869Spec.java
```diff
@@ -0,0 +1,152 @@
...
+        this.desiredLength = validateDesiredLength(desiredLength);
+        break;
+      case HKDF_EXTRACT_MODE:
+        this.secretOrPrk = Utils.requireNonNull(secret, "secret cannot be null for HKDF_EXTRACT");
+        this.salt = Utils.requireNonNull(salt, "salt cannot be null for HKDF_EXTRACT");
+        this.info = null;
+        this.desiredLength = 0;
+        break;
+      case HKDF_EXPAND_MODE:
+        this.secretOrPrk = Utils.requireNonNull(prk, "prk cannot be null for HKDF_EXPAND");
```
geedo0, line 42: Should we check that the PRK length is at least the size of the digest? From RFC5869

> PRK      a pseudorandom key of at least HashLen octets

src/com/amazon/corretto/crypto/provider/HkdfRfc5869Spec.java
```diff
@@ -0,0 +1,152 @@
...
+        this.desiredLength = validateDesiredLength(desiredLength);
+        break;
+      default:
+        throw new IllegalArgumentException("mode is not a valid value");
+    }
+    this.mode = mode;
+    this.algorithmName = algorithmName != null ? algorithmName : DEFAULT_ALGORITHM_NAME;
+  }
+
+  private static int validateDesiredLength(final int desiredLength) {
```
geedo0, line 54: RFC5869 defines an upper bound as well.

>       L        length of output keying material in octets (<= 255*HashLen)

src/com/amazon/corretto/crypto/provider/HkdfRfc5869Spec.java
```diff
@@ -0,0 +1,152 @@
...
+      final byte[] secret,
+      final byte[] salt,
+      final byte[] info,
+      final byte[] prk,
+      final int desiredLength,
+      final String algorithmName) {
+
+    switch (mode) {
+      case HKDF_MODE:
+        this.secretOrPrk = Utils.requireNonNull(secret, "secret cannot be null for HKDF");
```
geedo0, line 30: We require the secret to be non-null, should it also be non-empty? Curious what LC does in this case.

src/com/amazon/corretto/crypto/provider/HkdfRfc5869Spec.java
```diff
@@ -0,0 +1,152 @@
...
+        this.secretOrPrk = Utils.requireNonNull(prk, "prk cannot be null for HKDF_EXPAND");
+        this.salt = null;
+        this.info = Utils.requireNonNull(info, "info cannot be null for HKDF_EXPAND");
+        this.desiredLength = validateDesiredLength(desiredLength);
+        break;
+      default:
+        throw new IllegalArgumentException("mode is not a valid value");
+    }
+    this.mode = mode;
+    this.algorithmName = algorithmName != null ? algorithmName : DEFAULT_ALGORITHM_NAME;
```
geedo0, line 51: Should we check that the algorithm name is one of the standard names we're defining? Also, the default algorithm name is not one of our standard names.

tst/com/amazon/corretto/crypto/provider/test/HkdfRfc5869Test.java
```diff
@@ -0,0 +1,129 @@
...
+      (byte) 0xf2,
+      (byte) 0xf3,
+      (byte) 0xf4,
+      (byte) 0xf5,
+      (byte) 0xf6,
+      (byte) 0xf7,
+      (byte) 0xf8,
+      (byte) 0xf9
+    };
+    final KeySpec keySpec = hkdfSpec(secret, salt, info, 42, "My42ByteSecretKey");
```
geedo0, line 55: Computing the meaning of life? 😉

src/com/amazon/corretto/crypto/provider/HkdfRfc5869SecretKeyFactorySpi.java
```diff
@@ -0,0 +1,143 @@
...
+  @Override
+  protected SecretKey engineTranslateKey(final SecretKey key) {
+    throw new UnsupportedOperationException();
+  }
+
+  static final Map<String, HkdfRfc5869SecretKeyFactorySpi> INSTANCES = getInstances();
+
+  private static final String HKDF_RFC5869 = "HkdfRfc5869";
+  private static final String WITH = "With";
+  static final String HKDF_RFC5869_WITH_SHA1 = HKDF_RFC5869 + WITH + "SHA-1";
```
geedo0, line 122: Was there a discussion around whether or not to include SHA-1?

geedo0:
Do we have a place to document standard algorithm names for our exclusive algorithms?

src/com/amazon/corretto/crypto/provider/HkdfRfc5869SecretKeyFactorySpi.java
```diff
@@ -0,0 +1,143 @@
...
+    result.put(
+        HKDF_RFC5869_WITH_SHA256.toUpperCase(),
+        new HkdfRfc5869SecretKeyFactorySpi(2, getDigestLen(2)));
+    result.put(
+        HKDF_RFC5869_WITH_SHA384.toUpperCase(),
+        new HkdfRfc5869SecretKeyFactorySpi(3, getDigestLen(3)));
+    result.put(
+        HKDF_RFC5869_WITH_SHA512.toUpperCase(),
+        new HkdfRfc5869SecretKeyFactorySpi(4, getDigestLen(4)));
+    return result;
```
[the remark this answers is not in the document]
amirhosv: done

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -261,6 +271,14 @@ public Object newInstance(final Object constructorParameter) throws NoSuchAlgori
         final String type = getType();
         final String algo = getAlgorithm();
 
+        if ("SecretKeyFactory".equalsIgnoreCase(type)) {
+          final HkdfRfc5869SecretKeyFactorySpi spi =
+              HkdfRfc5869SecretKeyFactorySpi.INSTANCES.get(algo.toUpperCase());
```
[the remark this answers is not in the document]
amirhosv: Done

src/com/amazon/corretto/crypto/provider/HkdfRfc5869SecretKeyFactorySpi.java
amirhosv: It's interesting that in JCE/JCA for [digest names](https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#MessageDigest), they have dash. I like this better since we can just say replace `<digest>` with a valid digest name.

src/com/amazon/corretto/crypto/provider/HkdfRfc5869Spec.java
amirhosv: The Cipher class uses int for modes of operation and I'm just following that convention.

csrc/hkdf.cpp
amirhosv: Added an explicit check.

src/com/amazon/corretto/crypto/provider/HkdfRfc5869SecretKeyFactorySpi.java
amirhosv: Done.

src/com/amazon/corretto/crypto/provider/HkdfRfc5869Spec.java
amirhosv: LC is fine with empty input.

src/com/amazon/corretto/crypto/provider/HkdfRfc5869Spec.java
amirhosv: Empty is fine too. I added a test case for that.

src/com/amazon/corretto/crypto/provider/HkdfRfc5869Spec.java
amirhosv: The output of HKDF is a sequence of byte. This is a name that the client associates to that and we don't care that much about it.

tst/com/amazon/corretto/crypto/provider/test/HkdfRfc5869Test.java
amirhosv: Just some random name :)

src/com/amazon/corretto/crypto/provider/HkdfSpec.java
```diff
@@ -0,0 +1,152 @@
...
+        this.desiredLength = validateDesiredLength(desiredLength);
+        break;
+      default:
+        throw new IllegalArgumentException("mode is not a valid value");
+    }
+    this.mode = mode;
+    this.algorithmName = algorithmName != null ? algorithmName : DEFAULT_ALGORITHM_NAME;
+  }
+
+  private static int validateDesiredLength(final int desiredLength) {
```
geedo0, line 54: Still missing the upper bound check. Otherwise review lgtm

amirhosv: To do the upper bound check, the size of the digest is needed. The HkdfSpec class doesn't have the digest size. The [HkdfSecretKeyFactorySpi](https://github.com/amirhosv/amazon-corretto-crypto-provider/blob/hkdf/src/com/amazon/corretto/crypto/provider/HkdfSecretKeyFactorySpi.java#L84-L91) class has that information and that's where the check happens. The `checkExpandLength` is invoked when HKDF and HKDF_Expand.

src/com/amazon/corretto/crypto/provider/HkdfSecretKeyFactorySpi.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    return new SecretKeySpec(resultBytes, spec.algorithmName);
+  }
+
+  private void checkExpandLength(final long outLen) {
+    final long digestLengthLong = digestLength;
+    final long maxInt = Integer.MAX_VALUE;
+    final long n = (outLen + digestLengthLong - 1L) / digestLengthLong;
+    if (outLen + digestLengthLong > maxInt || n > 255) {
```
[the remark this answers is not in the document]
amirhosv: Since we are using longs, in case of overflow, `outLen + digestLengthLong > maxInt` becomes true.
LC has a similar logic: https://github.com/aws/aws-lc/blob/161e747063dcb16e71d299190f072a26c0a56294/crypto/fipsmodule/hkdf/hkdf.c#L94-L99.

I think `outLen >= digestLenghtLong * 255L`  is sufficient as you pointed out. I'll update the PR with that.

geedo0 (approved):

src/com/amazon/corretto/crypto/provider/HkdfSecretKeyFactorySpi.java
```diff
@@ -0,0 +1,166 @@
...
+    result.put(
+        getSpiFactoryForAlgName(HKDF_WITH_SHA256),
+        new HkdfSecretKeyFactorySpi(SHA256_CODE, getDigestLen(SHA256_CODE)));
+    result.put(
+        getSpiFactoryForAlgName(HKDF_WITH_SHA384),
+        new HkdfSecretKeyFactorySpi(SHA384_CODE, getDigestLen(SHA384_CODE)));
+    result.put(
+        getSpiFactoryForAlgName(HKDF_WITH_SHA512),
+        new HkdfSecretKeyFactorySpi(SHA512_CODE, getDigestLen(SHA512_CODE)));
+    return Collections.unmodifiableMap(result);
```
WillChilds-Klein, line 160: nit: not a big deal, but this method is a bit redundant as we already [have Utils for getting + caching digest len](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/Utils.java#L54-L95). example for going from name to length [here](https://github.com/corretto/amazon-corretto-crypto-provider/blob/74ac74160e426dc973edc0a0cffb108a1f8f7baf/src/com/amazon/corretto/crypto/provider/EvpHmac.java#L316).

it looks like this only gets called once per digest algorithm at JVM start, so the caching aspect isn't very important.

src/com/amazon/corretto/crypto/provider/HkdfSecretKeyFactorySpi.java
```diff
@@ -0,0 +1,166 @@
...
+        throw new AssertionError(
+            "This should not be reachable, since the constructor of HkdfSpec ensures the value is"
+                + " one of the above choices.");
+    }
+
+    return new SecretKeySpec(resultBytes, spec.algorithmName);
+  }
+
+  private void checkExpandLength(final long outLen) {
+    final long upperLimit = ((long) digestLength) * 255L;
```
WillChilds-Klein, line 85: where does the constant 255L come from?

tst/com/amazon/corretto/crypto/provider/test/TestUtil.java
```diff
@@ -60,10 +71,31 @@ public class TestUtil {
         new String[] {"secp256k1", "1.3.132.0.10"},
       };
 
-  public static final boolean isFips() {
+  public static boolean isFips() {
```
WillChilds-Klein, line 74: why was the `final` qualifier removed here?

tst/com/amazon/corretto/crypto/provider/test/TestUtil.java
```diff
@@ -60,10 +71,31 @@ public class TestUtil {
...
     return NATIVE_PROVIDER.isFips();
   }
 
+  public static byte[] intArrayToByteArray(final int[] array) {
+    final byte[] result = new byte[array.length];
+    for (int i = 0; i != array.length; i++) {
+      result[i] = (byte) array[i];
```
WillChilds-Klein, line 81: is this safe? i wonder if we should throw an exception if the array element is >255

src/com/amazon/corretto/crypto/provider/HkdfSecretKeyFactorySpi.java
amirhosv: Thanks for pointing this out. I removed the redundant code and used the APIs in Utils class instead.

src/com/amazon/corretto/crypto/provider/HkdfSecretKeyFactorySpi.java
amirhosv: I commented the code. It's coming from the standard.

tst/com/amazon/corretto/crypto/provider/test/TestUtil.java
amirhosv: It's redundant for static methods.

tst/com/amazon/corretto/crypto/provider/test/TestUtil.java
amirhosv: I was copy-pasting tests from LC and since bytes are signed in Java, for values greater than 127 (or 128), Java wanted the hex values to be typecast to bytes. So I decided to do the copy-pasting as int and do the casting in this method. Having said that, I added the bound checks to this method, despite being only used for testing.

geedo0 (approved):
lgtm

PR #320, opened 2023-08-04
Improve SignatureException handling

SalusaSecondus:
*Description of changes:*

The JCA does not fully specify when SignatureException should be thrown but the ResetAfterException test from OpenJDK indicates that SignatureException should be thrown when [validating a too-short signature](https://github.com/openjdk/jdk/blob/101bf2290da5735fd9624ab647a8183c2c21f22d/test/jdk/java/security/Signature/ResetAfterException.java#L107-L109). This PR modifies ACCP to throw an exception in this case and to always properly reset when exceptions are thrown.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
@@ -423,8 +423,9 @@ abstract class EvpSignatureBase extends SignatureSpi {
     // Right now we only check RSA signatures to ensure they are the proper length
     if (key_ instanceof RSAKey) {
       final RSAKey rsaKey = (RSAKey) key_;
-      if (length != (rsaKey.getModulus().bitLength() + 7) / 8) {
-        throw new SignatureException("RSA Signature of invalid length.");
+      final int expectedLength = (rsaKey.getModulus().bitLength() + 7) / 8;
+      if (length != expectedLength) {
+        throw new SignatureException("RSA Signature of invalid length. Expected " + expectedLength);
       }
     }
   }
--- a/src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
@@ -413,4 +413,19 @@ abstract class EvpSignatureBase extends SignatureSpi {
     System.arraycopy(signature, sStart, result, numLen + numLen - sLen, sLen);
     return result;
   }
+
+  /**
+   * Does an initial check of the signature seeing if it is of the proper format and size. This lets
+   * us quickly reject invalid signatures in a way that the JDK expects.
+   */
+  protected void sniffTest(final byte[] signature, final int offset, final int length)
+      throws SignatureException {
+    // Right now we only check RSA signatures to ensure they are the proper length
+    if (key_ instanceof RSAKey) {
+      final RSAKey rsaKey = (RSAKey) key_;
+      if (length != (rsaKey.getModulus().bitLength() + 7) / 8) {
+        throw new SignatureException("RSA Signature of invalid length.");
+      }
+    }
+  }
 }
--- a/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
@@ -639,6 +639,36 @@ public final class EvpSignatureSpecificTest {
   }
 
   /**
+   * For better compatibility with the JDK, we want to throw SignatureException if the length of the
+   * signature is wrong.
+   */
+  @Test
+  public void rsaSignatureWrongLength() throws Exception {
+    final Signature signer = Signature.getInstance("SHA256withRSA", NATIVE_PROVIDER);
+    signer.initSign(RSA_PAIR.getPrivate());
+    signer.update(MESSAGE);
+    final byte[] validSignature = signer.sign();
+
+    // Sanity check
+    signer.initVerify(RSA_PAIR.getPublic());
+    signer.update(MESSAGE);
+    assertTrue(signer.verify(validSignature));
+
+    // Too long throws an exception
+    final byte[] longSig = Arrays.copyOf(validSignature, validSignature.length + 1);
+    longSig[longSig.length - 1] = 1;
+    signer.initVerify(RSA_PAIR.getPublic());
+    signer.update(MESSAGE);
+    assertThrows(SignatureException.class, () -> signer.verify(longSig));
+
+    // Too short throws an exception
+    final byte[] shortSig = Arrays.copyOf(validSignature, validSignature.length - 1);
+    signer.initVerify(RSA_PAIR.getPublic());
+    signer.update(MESSAGE);
+    assertThrows(SignatureException.class, () -> signer.verify(shortSig));
+  }
+
+  /**
    * This test iterates over every implemented algorithm and ensures that it is compatible with the
    * equivalent BouncyCastle implementation. It doesn't check negative cases as the more detailed
    * tests cover that for algorithm families.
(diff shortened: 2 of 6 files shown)


tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -638,6 +638,36 @@ public void ecdsaSignCorruptsErrorState() throws Exception {
...
+    final byte[] validSignature = signer.sign();
+
+    // Sanity check
+    signer.initVerify(RSA_PAIR.getPublic());
+    signer.update(MESSAGE);
+    assertTrue(signer.verify(validSignature));
+
+    // Too long throws an exception
+    final byte[] longSig = Arrays.copyOf(validSignature, validSignature.length + 1);
+    longSig[longSig.length - 1] = 1;
```
WillChilds-Klein, line 659: why even fill this in?

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -638,6 +638,36 @@ public void ecdsaSignCorruptsErrorState() throws Exception {
     }
   }
 
+  /**
+   * For better compatibility with the JDK, we want to throw SignatureException if the length of the
+   * signature is wrong.
+   */
+  @Test
+  public void rsaSignatureWrongLength() throws Exception {
+    final Signature signer = Signature.getInstance("SHA256withRSA", NATIVE_PROVIDER);
```
WillChilds-Klein, line 647: should we test this over ECDSA as well?

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -638,6 +638,36 @@ public void ecdsaSignCorruptsErrorState() throws Exception {
     }
   }
 
+  /**
+   * For better compatibility with the JDK, we want to throw SignatureException if the length of the
+   * signature is wrong.
```
WillChilds-Klein, line 643: this may help with JDK compatibility, but it diverges from [BC's behavior](https://github.com/bcgit/bc-java/blob/91c3c6018ab353f653749b9b56d6de384a31ec3c/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/rsa/DigestSignatureSpi.java#L146) of returning `false` rather than throwing.

hoping to have further discussion here with greg and the ACCP team on tradeoffs of favoring JDK compatibility vs. BC compatibility.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
SalusaSecondus: To ensure that we don't just consume the valid prefix and incorrectly accept.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
SalusaSecondus: In an ideal world, it would only be `false` or `true`, but unfortunately that isn't the world we're in.

We don't need to worry about "leaking information" because every piece of data handled here is public. (Signatures are considered public and public keys are used to validate them.) The concerns are complexity and compatibility.

Complexity: Developers already (unfortunately) need to handle `SignatureException` and this doesn't make things any worse for them.

Compatibility: The JDK's providers (and tests) define the standard through their behaviors. ACCP cannot match it *perfectsly*, but it can try to minimize the differences. Given a choice between mutually exclusive options of "compatible with JDK providers" and "compatible with BC," I believe that ACCP should *usually* choose to be compatible with the JDK providers. (Obviously, there may be questions of security and correctness, but neither apply here.)

Finally, I believe it is a reasonably generalization to say "Teams use the JDK providers by default and might *choose* to use BC." This means that use of BC is *more likely* to be an intentional decision rather than using the JDK providers. So, if ACCP is being added, we shouldn't change how things work for teams which are just grabbing the default and not paying attention to the details. If a team has chosen to use BC then:

* They are overriding ACCP (so these changes don't matter anyway), OR
* They are making intentional decisions around their cryptography and so can be expected to make adjustments to their code if needed.

Truthfully, many teams use BC without thinking about it (either because of a dependency or just not understanding cryptographic code). But it remains true that using BC is more of a conscious choice than just sticking with the default.

amirhosv: ACCP-FIPS compatibility with BCFIPS is highly desirable. Customers that have FIPS requirements and want to switch to ACCP-FIPS for performance and still satisfy the FIPS requirements would have simpler transition.
P.S. we have just released ACCP-FIPS for experimental purposes. AWS-LC-FIPS is still waiting for it certificate.

SalusaSecondus: I still believe that my above positions hold the vast majority of the time.

Careful code *needs* to handle `SignatureException` (partially because it is a checked exception).

There is an interesting question as to whether there should be some way to tune ACCP's compatibility towards JDK or BC, but I think the answer there should remain "no."

src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
```diff
@@ -413,4 +413,19 @@ protected static byte[] asn1ToiIeeeP1363(byte[] signature, int numLen) throws Si
...
+   * Does an initial check of the signature seeing if it is of the proper format and size. This lets
+   * us quickly reject invalid signatures in a way that the JDK expects.
+   */
+  protected void sniffTest(final byte[] signature, final int offset, final int length)
+      throws SignatureException {
+    // Right now we only check RSA signatures to ensure they are the proper length
+    if (key_ instanceof RSAKey) {
+      final RSAKey rsaKey = (RSAKey) key_;
+      if (length != (rsaKey.getModulus().bitLength() + 7) / 8) {
+        throw new SignatureException("RSA Signature of invalid length.");
```
geedo0, line 427: Would it be helpful to report the expected and received lengths?

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
geedo0: I agree with @SalusaSecondus's points. The incompatibility introduced here is trivially resolved by the customers and returning more contextual information in our SignatureExceptions is more helpful than a boolean failure.

geedo0 (approved):

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
geedo0: I agree with @SalusaSecondus's points. The incompatibility introduced here is trivially resolved by the customers and returning more contextual information in our SignatureExceptions is more helpful than a boolean failure.

amirhosv: Sounds good to me.

src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
SalusaSecondus: Whatever you prefer. I'm happy either way.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
amirhosv: Will's guideline is a great mechanism for deciding how incompatibilities should be handled.

src/com/amazon/corretto/crypto/provider/EvpSignatureBase.java
geedo0: I wouldn't block the merge over it, I'd only encourage it if you're pushing another rev anyway.

WillChilds-Klein (approved):

amirhosv (approved):

PR #321, opened 2023-08-04
Round RSA key sizes up when generating

SalusaSecondus:
AWS-LC will only generate RSA keys with bit-lengths that are multiples of 128 and will round the requested bit-length down to meet that [requirement](https://github.com/aws/aws-lc/blob/25260d785f6e2eaf3c5f5dce83cf92c272f0a8b1/crypto/fipsmodule/rsa/rsa_impl.c#L1168-L1171). This means that when ACCP is used to generate an RSA key, requests such as KeyPairGenerator.initialize(2049) will result in smaller than requested keys. In general, we should try to always provide at least as much security as requested and so rounding down is unexpected.

For example,the above behavior causes tests such as the [DefaultSignatureAlgorithm test from the OpenJDK](https://github.com/openjdk/jdk/blob/dfacda488bfbe2e11e8d607a6d08527710286982/test/jdk/sun/security/tools/keytool/fakegen/DefaultSignatureAlgorithm.java#L51) to fail. It fails because it is checking for behavior differences for keys larger than a threshold (such as 3072) where the default signature algorithm changes from using SHA256 to SHA384. With the current version of ACCP, requesting a 3073 bit key gets a 3072 key (and thus will use SHA256) so the test fails. By changing the logic to rounding up, requesting a 3073 bit key gets a 3200 bit key (which will trigger use of SHA384) and the test passes.

### DIFF
--- a/csrc/rsa_gen.cpp
+++ b/csrc/rsa_gen.cpp
@@ -35,7 +35,7 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_RsaGen_generate
             // round up. We only do this in the non-FIPS branch because in FIPS mode we want to do
             // exactly what the application requests.
             if (bits % 128 != 0) {
-                bits += 128;
+                bits += 128 - (bits % 128);
             }
 
             BigNumObj bne;
--- a/csrc/rsa_gen.cpp
+++ b/csrc/rsa_gen.cpp
@@ -21,14 +21,6 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_RsaGen_generate
     try {
         raii_env env(pEnv);
 
-        // AWS-LC requires that the bitlength be a multiple of 128 and will round down.
-        // We want to guarantee that we return a key of at least the requested strength and so must
-        // round up.
-        jint rounded_bits = bits & ~127;
-        if (rounded_bits < bits) {
-            bits += 128;
-        }
-
         if (FIPS_mode() == 1) {
             // RSA_generate_key_fips performs extra checks so there is no need
             // to run post generation checks. This API generates keys with
@@ -38,6 +30,14 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_RsaGen_generate
                 throw_openssl("Unable to generate key");
             }
         } else {
+            // AWS-LC requires that the bitlength be a multiple of 128 and will round down.
+            // We want to guarantee that we return a key of at least the requested strength and so must
+            // round up. We only do this in the non-FIPS branch because in FIPS mode we want to do
+            // exactly what the application requests.
+            if (bits % 128 != 0) {
+                bits += 128;
+            }
+
             BigNumObj bne;
             jarr2bn(env, pubExp, bne);
 
--- a/csrc/rsa_gen.cpp
+++ b/csrc/rsa_gen.cpp
@@ -21,6 +21,14 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_RsaGen_generate
     try {
         raii_env env(pEnv);
 
+        // AWS-LC requires that the bitlength be a multiple of 128 and will round down.
+        // We want to guarantee that we return a key of at least the requested strength and so must
+        // round up.
+        jint rounded_bits = bits & ~127;
+        if (rounded_bits < bits) {
+            bits += 128;
+        }
+
         if (FIPS_mode() == 1) {
             // RSA_generate_key_fips performs extra checks so there is no need
             // to run post generation checks. This API generates keys with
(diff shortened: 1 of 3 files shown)


WillChilds-Klein (commented):
I really like this change on security merits, and even believe it would be worth sacrificing some compatibility with BC to round the security level up. If compatibility is a goal, should we add a compatibility unit test between ACCP, JDK, and BC so we can codify expected behavior across the different providers?

SalusaSecondus:
I'm not sure if we can really promise full compatibility here. I didn't check to see exactly how the JDK and BC handle this. Part of the problem is that AWS-LC rounds to 128 bits and I don't think that the others do.

I think that rounding down in AWS-LC is reasonable because callers may allocate a static byte array to receive the modulus, so ensuring that the result still fits in the same size makes sense. However, in Java that concern isn't present and I think that passing more JDK tests and always giving "at least as much security" is better.

We should codify *our* behavior (and document it in `DIFFERENCES.md`).

amirhosv:
> I'm not sure if we can really promise full compatibility here. I didn't check to see exactly how the JDK and BC handle this. Part of the problem is that AWS-LC rounds to 128 bits and I don't think that the others do.
> 
> I think that rounding down in AWS-LC is reasonable because callers may allocate a static byte array to receive the modulus, so ensuring that the result still fits in the same size makes sense. However, in Java that concern isn't present and I think that passing more JDK tests and always giving "at least as much security" is better.
> 
> We should codify _our_ behavior (and document it in `DIFFERENCES.md`).
@SalusaSecondus Thanks for the PR and also bringing this strange behavior to our attention. I tried BCFIPS and SunRsaSign and both generate keys with the exact requested bit length. IMO it should not be the responsibility of ACCP or AWS-LC to modify the key length to accommodate the values that they do not support; the right behavior by ACCP should be throwing an exception when the key length is not supported by ACCP/AWS-LC. The client code should make explicit choice for such key sizes and see if they should round up or down or any other logic.
It's a limitation and maybe it's best to be honest about it by throwing an exception.

csrc/rsa_gen.cpp
```diff
@@ -21,6 +21,14 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_RsaGen_generate
     try {
         raii_env env(pEnv);
 
+        // AWS-LC requires that the bitlength be a multiple of 128 and will round down.
+        // We want to guarantee that we return a key of at least the requested strength and so must
+        // round up.
+        jint rounded_bits = bits & ~127;
```
geedo0, line 27: nit: Defining bit mask literals in hex is easier to reason about.

geedo0 (dismissed):
I agree with @amirhosv that the "right" thing to do is to just reject anything unsupported and with @WillChilds-Klein that rounding up is the more secure choice relative to rounding down. Retaining the rounding behavior is more neutral from a compatibility perspective so that's my preference.

SalusaSecondus:
(Not ignoring the discussion here, just a lot to think about.)

I agree that actually rejecting might be a reasonable choice too. I'm not sure which is the better option. Rounding up "just works" more often, but gives a somewhat unexpected result. Rejecting always gives the right answer. Rounding down seems the worst of the options.

Rounding up is less work for me as a general consumer. That said, I don't know which my preference is yet.

csrc/rsa_gen.cpp
```diff
@@ -21,6 +21,14 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_RsaGen_generate
...
+        if (rounded_bits < bits) {
+            bits += 128;
+        }
+
         if (FIPS_mode() == 1) {
```
amirhosv, line 32: It's best to do rounding up in nonFIPS mode only. This would increase the possible values that ACCP allows despite LC not allowing.

SalusaSecondus: Is this the consensus solution? I'll happily do it if we think this is what will make us all happy.

amirhosv: Yes. It's OK not to throw exception and do rounding up for non-FIPS builds. For FIPS builds, no rounding and throwing exception for the key bit lengths that are not supported.

WillChilds-Klein: I'm on board with amir's proposal. We can always relax the FIPS strictness if we encounter a legitimate use-case that requires specifying modulus size `% 128 != 0`.

amirhosv (approved):

csrc/rsa_gen.cpp
```diff
@@ -30,6 +30,14 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_RsaGen_generate
                 throw_openssl("Unable to generate key");
             }
         } else {
+            // AWS-LC requires that the bitlength be a multiple of 128 and will round down.
+            // We want to guarantee that we return a key of at least the requested strength and so must
+            // round up. We only do this in the non-FIPS branch because in FIPS mode we want to do
+            // exactly what the application requests.
+            if (bits % 128 != 0) {
+                bits += 128;
```
WillChilds-Klein, line 38: should we be more precise here? if AWS-LC changes their behavior in the future and starts to allow key sizes `% 128 != 0`, then our behavior will change in a strange way -- we'll allow users to create keys at odd sizes _and_ add 128 bits to the requested size.

```suggestion
                bits += 128 - (bits % 128);
```

WillChilds-Klein: > As opposed to ACCP always rounding up to the nearest multiple of 128 even after it's no longer needed?

Right. It's still kinda weird, but at least we'd remain consistent with current behavior (i.e. current as of this PR).

WillChilds-Klein (approved):

PR #325, opened 2023-09-05
Move benchmarks to an independent project

amirhosv:
## Move benchmarks to an independent project
* Benchmarks now depend on ACCP and there is no need to build ACCP
* Benchmarks can be run against published ACCP artifacts on Maven


benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/AesGcmOneShot.java
```diff
@@ -1,8 +1,10 @@
 // Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-package com.amazon.corretto.crypto.provider;
```
amirhosv, line 3: When using published ACCP for benchmarking, it complains that the benchmark package is not consistent with ACCP's signed JAR. To resolve this issue, the benchmark code is moved to `com.amazon.corretto.crypto.provider.benchmarks` from `com.amazon.corretto.crypto.provider`.

benchmarks/lib/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
...
+    if (project.hasProperty("fips"))
+        "AmazonCorrettoCryptoProvider-FIPS"
+    else
+        "AmazonCorrettoCryptoProvider"
+
+    val publishedAccpVersion =
+    if (project.hasProperty("accpVersion"))
+        accpVersion
+    else
+        "2.3.1"
```
WillChilds-Klein, line 31: do we want to just always take the latest instead of fixing this to a specific version we'll have to remember to update through time?

benchmarks/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+* `-Pfips` flag can be used to use the FIPS artifacts for benchmarking.
+* `-PaccpVerion="ACCP_VERSION"` can be used to run benchmarks for a specific version of ACCP
+
+### Benchmarking locally built ACCP
+
+Use `-PaccpLocalJar="PATH_TO_LOCAL_JAR"`:
+
+```bash
+ ./gradlew -PaccpLocalJar="../../build/cmake/AmazonCorrettoCryptoProvider.jar" lib:jmh
```
WillChilds-Klein, line 21: minor nit: can we also specify the `jmhReport` task in this example?

WillChilds-Klein (commented):
Excellent simplification @amirhosv !

benchmarks/lib/build.gradle.kts
amirhosv: Replaced "2.3.1" with "2.+" so it would use the latest.

benchmarks/README.md
amirhosv: Modified the build.gradle.kts so that the `jmhReport` would always be executed after `jmh`. In this way, we don't have to explicitly run it.

benchmarks/lib/build.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+    jvmArgs.add("-DversionStr=${accpVersion}")
+    jvmArgs.add("-Dcom.amazon.corretto.crypto.provider.registerSecureRandom=true")
+}
+
+jmhReport {
+    jmhReportOutput = "lib/build/results/jmh"
+    jmhResultPath = "${jmhReportOutput}/results.json"
+}
+
+tasks.jmh {
```
amirhosv, line 69: This block ensures `jmhReport` is executed after `jmh`.

geedo0:
Thanks for doing this! This'll solve a lot of annoying problems related to the awkward JMH source set breaking conventions in IDEs and Gradle.

geedo0 (approved):
lgtm

PR #329, opened 2023-09-15
Allow users to control the release of EVP context for AES-GCM

amirhosv:
*Description of changes:*

This change introduces a system property that allows one to control when an EVP_CIPHER_CTX (native context for encryption) is released. For use cases when the same Cipher object is used for multiple encryption/decryption operations, this can result in improved performance since it avoids allocation and release of the native context while the Cipher object is not garbage collected.

There are three strategies for releasing the context: 1) `Hybrid`, 2) `Eager`, 3) `Lazy`.

* `Hybrid`: this is the existing strategy and it's introduced so that existing customers would not be affected. In this strategy, the context is released eagerly, unless it is used with the same key. Benchmark `CipherReuse.reuse` when `newKey` is `false` shows that the existing optimizations are preserved for customers that use the default setting.
* `Eager`: in this strategy, the context is released as soon as possible. Based on `CipherReuse.newInstance` benchmarks, customers that use a fresh instance of `Cipher` class for their operations, could experience 30% performance improvement when `Eager` strategy is used.
* `Lazy`: the context object is preserved while the `Cipher` object is not garbage collected. Customers caching Cipher objects. Compared to default `Hybrid` strategy, could experience 16% performance improvement even if they use different keys for with the same `Cipher` class; however, using the same key, this strategy and `Hybrid` perform the same.

Here is benchmark results:

ACCP 2.3.2 (the latest release of ACCP on Maven)

```
#../gradlew -PaccpVersion="2.3.2" -PincludeBenchmark="CipherReuse" lib:jmh

Benchmark                (newKey)   Mode  Cnt       Score       Error  Units
CipherReuse.newInstance      true  thrpt    5  195961.858 ± 26482.726  ops/s
CipherReuse.newInstance     false  thrpt    5  194516.879 ± 31070.067  ops/s
CipherReuse.reuse            true  thrpt    5  305481.372 ± 13363.712  ops/s
CipherReuse.reuse           false  thrpt    5  440410.914 ± 18570.172  ops/s
```

HYBRID

```
#./gradlew -PaccpLocalJar="../../build/cmake/AmazonCorrettoCryptoProvider.jar" -PnativeContextReleaseStrategy="HYBRID" -PincludeBenchmark="CipherReuse" lib:jmh

Benchmark                (newKey)   Mode  Cnt       Score       Error  Units
CipherReuse.newInstance      true  thrpt    5  191908.943 ± 26892.090  ops/s
CipherReuse.newInstance     false  thrpt    5  197180.321 ± 29013.380  ops/s
CipherReuse.reuse            true  thrpt    5  323944.919 ± 11095.834  ops/s
CipherReuse.reuse           false  thrpt    5  444121.750 ±  9679.539  ops/s
```

EAGER

```
#./gradlew -PaccpLocalJar="../../build/cmake/AmazonCorrettoCryptoProvider.jar" -PnativeContextReleaseStrategy="EAGER" -PincludeBenchmark="CipherReuse" lib:jmh

Benchmark                (newKey)   Mode  Cnt       Score       Error  Units
CipherReuse.newInstance      true  thrpt    5  275440.808 ±  9990.183  ops/s
CipherReuse.newInstance     false  thrpt    5  275176.586 ± 10058.817  ops/s
CipherReuse.reuse            true  thrpt    5  327901.459 ± 12479.909  ops/s
CipherReuse.reuse           false  thrpt    5  324376.765 ± 18051.911  ops/s
```

LAZY

```
#./gradlew -PaccpLocalJar="../../build/cmake/AmazonCorrettoCryptoProvider.jar" -PnativeContextReleaseStrategy="LAZY" -PincludeBenchmark="CipherReuse" lib:jmh

Benchmark                (newKey)   Mode  Cnt       Score       Error  Units
CipherReuse.newInstance      true  thrpt    5  223333.903 ± 44229.461  ops/s
CipherReuse.newInstance     false  thrpt    5  221117.187 ± 34093.029  ops/s
CipherReuse.reuse            true  thrpt    5  382687.274 ± 18950.889  ops/s
CipherReuse.reuse           false  thrpt    5  440212.922 ± 11446.041  ops/s
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,7 +274,7 @@ AmazonCorrettoCryptoProvider.INSTANCE.assertHealthy();
 
 ### Other system properties
 ACCP can be configured via several system properties.
-None of these should be needed for standard deployments and we recommend not touching them.
+None of these should be needed for standard deployments, and we recommend not touching them.
 They are of most use to developers needing to test ACCP or experiment with benchmarking.
 These are all read early in the load process and may be cached so any changes to them made from within Java may not be respected.
 Thus, these should all be set on the JVM command line using `-D`.
@@ -316,7 +316,21 @@ Thus, these should all be set on the JVM command line using `-D`.
   This means that there is a slight "pause" before ACCP FIPS's SecureRandom can produce pseudo-random bytes in highly threaded environments.
   Because, in extreme cases this could present an availability risk, we do not register LibCryptoRng by default in configurations where this initialization cost is incurred (i.e. FIPS mode).
   Non-FIPS AWS-LC does not use CPU jitter for its DRBG seed's entropy, and therefore does not incur this initialization cost, therefore we register LibCryptoRng by default when not in FIPS mode.
-
+* `com.amazon.corretto.crypto.provider.nativeContextReleaseStrategy`
+  Takes in `HYBRID`, `LAZY`, or `EAGER` (defaults ot `HYBRID`). This property only affects
+  AES-GCM cipher for now. AES-GCM associates a native object of type `EVP_CIPHER_CTX`
+  to each `Cipher` object. This property allows users to control the strategy for releasing
+  the native object.
+  * `HYBRID` (default): the structure is released eagerly, unless the same AES key is used. This is the
+     default behavior, and it is consistent with prior releases of ACCP.
+  * `LAZY`: preserve the native object and do not release while the `Cipher` object is not garbage collected.
+  * `EAGER`: release the native object as soon as possible, regardless of using the same key or not.
+  Our recommendation is to set this property to `EAGER` if `Cipher` objects are discarded
+  after use and caching of `Cipher` objects is not needed. When reusing the same `Cipher`
+  object, it would be beneficial to set this system property to `LAZY` so that different
+  encryption/decryption operations would not require allocation and release of `EVP_CIPHER_CTX`
+  structure. A common use case would be having long-running threads that each would get its
+  own instance of `Cipher` class.
 
 # License
 This library is licensed under the Apache 2.0 license although portions of this
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -6,6 +6,9 @@ The benchmarks can use locally built ACCP or published ACCP.
 The `lib:jmh` Gradle task runs the benchmarks and generates reports in JSON and HTML.
 The reports are saved under `lib/build/results/jmh`.
 
+* `-PincludeBenchmark="INCLUDE_BENCHMARK"` would only run the specified matching benchmarks.
+  * `INCLUDE_BENCHMARK` is a regular expression. For example, `CipherReuse|AesKwp` runs these two sets of benchmarks only.
+
 ### Benchmarking published ACCP to Maven
 
 ```bash
--- a/benchmarks/lib/build.gradle.kts
+++ b/benchmarks/lib/build.gradle.kts
@@ -1,6 +1,8 @@
 val accpVersion: String? by project
 val accpLocalJar: String by project
 val fips: Boolean by project
+val includeBenchmark: String by project
+val nativeContextReleaseStrategy: String by project
 
 plugins {
     `java-library`
@@ -46,7 +48,9 @@ java {
 }
 
 jmh {
-    // includes.add("AesXts") // can be used to run a subset of benchmarks
+    if (project.hasProperty("includeBenchmark")) {
+        includes.add(includeBenchmark)
+    }
     fork.set(1)
     benchmarkMode.add("thrpt")
     threads.set(1)
@@ -59,6 +63,9 @@ jmh {
     duplicateClassesStrategy.set(DuplicatesStrategy.WARN)
     jvmArgs.add("-DversionStr=${accpVersion}")
     jvmArgs.add("-Dcom.amazon.corretto.crypto.provider.registerSecureRandom=true")
+    if (project.hasProperty("nativeContextReleaseStrategy")) {
+        jvmArgs.add("-Dcom.amazon.corretto.crypto.provider.nativeContextReleaseStrategy=${nativeContextReleaseStrategy}")
+    }
 }
 
 jmhReport {
--- a/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/CipherReuse.java
+++ b/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/CipherReuse.java
@@ -1,0 +1,75 @@
+package com.amazon.corretto.crypto.provider.benchmarks;
+
+import com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.infra.Blackhole;
+
+import javax.crypto.Cipher;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import java.util.Random;
+
+@State(Scope.Thread)
+public class CipherReuse {
+    private final static String AES = "AES";
+    private final static String AES_GCM = "AES/GCM/NoPadding";
+    Cipher shared;
+    SecretKeySpec key;
+    Random random = new Random();
+    byte[] iv = new byte[12];
+    byte[] input = new byte[1000];
+    byte[] keyb = new byte[32];
+    byte[] aad = new byte[100];
+
+    @Param({"true", "false"})
+    private boolean newKey;
+
+    @Setup
+    public void setup() throws Exception {
+        shared = getAesGcmCipherFromAccp();
+        random.nextBytes(keyb);
+        key = new SecretKeySpec(keyb, AES);
+        random.nextBytes(input);
+        random.nextBytes(aad);
+    }
+
+    @Benchmark
+    public void newInstance(Blackhole blackhole) throws Exception {
+        blackhole.consume(encryptDecrypt(getAesGcmCipherFromAccp()));
+    }
+
+    @Benchmark
+    public void reuse(Blackhole blackhole) throws Exception {
+        blackhole.consume(encryptDecrypt(shared));
+    }
+
+    byte[] encryptDecrypt(final Cipher cipher) throws Exception {
+        iv[0]++;
+        final SecretKey sk;
+        if (newKey) {
+            keyb[0]++;
+            sk = new SecretKeySpec(keyb, AES);
+        } else {
+            sk = key;
+        }
+        cipher.init(Cipher.ENCRYPT_MODE, sk, new GCMParameterSpec(128, iv));
+        cipher.updateAAD(aad);
+        final byte[] cipherText = cipher.doFinal(input);
+        cipher.init(Cipher.DECRYPT_MODE, sk, new GCMParameterSpec(128, iv));
+        cipher.updateAAD(aad);
+        return cipher.doFinal(cipherText);
+    }
+
+    private Cipher getAesGcmCipherFromAccp() {
+        try {
+            return Cipher.getInstance(AES_GCM, AmazonCorrettoCryptoProvider.INSTANCE);
+        } catch (final Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+}
--- a/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
@@ -32,12 +32,6 @@ final class AesGcmSpi extends CipherSpi {
     Loader.load();
   }
 
-  /**
-   * The number a times a key must be reused prior to keeping it in native memory rather than
-   * freeing it each time.
-   */
-  private static final int KEY_REUSE_THRESHOLD = 1;
-
   private static final int DEFAULT_TAG_LENGTH = 16 * 8;
 
   /* Some random notes:
@@ -69,6 +63,7 @@ final class AesGcmSpi extends CipherSpi {
    */
   private static native int oneShotEncrypt(
       long ctxPtr,
+      boolean sameKey,
       long[] ctxPtrOut,
       byte[] input,
       int inputOffset,
@@ -100,6 +95,7 @@ final class AesGcmSpi extends CipherSpi {
    */
   private static native int oneShotDecrypt(
       long ctxPtr,
+      boolean sameKey,
       long[] ctxPtrOut,
       byte[] input,
       int inoffset,
@@ -116,20 +112,14 @@ final class AesGcmSpi extends CipherSpi {
   /**
    * Initializes state for a non-one-shot encryption operation.
    *
+   * @param ptr Context pointer to be used if not zero
+   * @param sameKey true iff the key for initialization has been used before
    * @param key Encryption key
    * @param iv Initialization vector
    * @return Native pointer to context data structure, which must be freed using releaseContext() or
    *     encryptDoFinal()
    */
-  private static native long encryptInit(byte[] key, byte[] iv);
-
-  /**
-   * Reuses an existing EVP context and initializes it for encryption given the new IV.
-   *
-   * @param ptr Context pointer
-   * @param iv Initialization vector
-   */
-  private static native void encryptInit(long ptr, byte[] iv);
+  private static native long encryptInit(long ptr, boolean sameKey, byte[] key, byte[] iv);
 
   /**
    * Processes some plaintext during a non-one-shot encryption operation. This is essentially a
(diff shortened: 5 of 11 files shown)


README.md
```diff
@@ -315,7 +314,14 @@ Thus, these should all be set on the JVM command line using `-D`.
   This means that there is a slight "pause" before ACCP FIPS's SecureRandom can produce pseudo-random bytes in highly threaded environments.
   Because, in extreme cases this could present an availability risk, we do not register LibCryptoRng by default in configurations where this initialization cost is incurred (i.e. FIPS mode).
   Non-FIPS AWS-LC does not use CPU jitter for its DRBG seed's entropy, and therefore does not incur this initialization cost, therefore we register LibCryptoRng by default when not in FIPS mode.
-
+* `com.amazon.corretto.crypto.provider.freeNativeContextEagerly`
+  Takes in `true` or `false` (defaults to `true`). If `true`, the underlying `EVP_CIPHER_CTX`
+  of a GCM `Cipher` is freed as soon as possible; otherwise, the underlying `EVP_CIPHER_CTX`
```
geedo0, line 319: This documentation is misleading. We are naming this property very generically, but it only impacts AES-GCM. It doesn't makes sense to expand the scope of this work to allow re-use for all algorithms. I suggest leaving the naming as-is, but calling out that this only applies to AES-GCM more clearly (This leaves the door open to more context reuse in the future).

benchmarks/README.md
```diff
@@ -6,6 +6,8 @@ The benchmarks can use locally built ACCP or published ACCP.
 The `lib:jmh` Gradle task runs the benchmarks and generates reports in JSON and HTML.
 The reports are saved under `lib/build/results/jmh`.
 
+* `-Pib="INLCUDE_BENCHMARK"` would only run specified benchmark.
```
geedo0, line 9: nit: typo

benchmarks/lib/build.gradle.kts
```diff
@@ -1,6 +1,7 @@
 val accpVersion: String? by project
 val accpLocalJar: String by project
 val fips: Boolean by project
+val ib: String by project
```
geedo0, line 4: nit: The name is ambiguous. Maybe give it a long name instead? I think Gradle is smart enough to shorten a long name automatically so we can have it both ways.

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/CipherReuse.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Benchmark
+    public void reuse(Blackhole blackhole) throws Exception {
+        blackhole.consume(encrypt(shared));
+    }
+
+    byte[] encrypt(Cipher cipher) throws Exception {
+        random.nextBytes(iv);
+        keyb[0]++;
+        final String name = random.nextBoolean() ? "AES" : "aes";
```
geedo0, line 60: Seems unecessary for this benchmark.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -368,39 +359,23 @@ protected void engineInit(
...
-      case Cipher.WRAP_MODE:
-        this.opMode = NATIVE_MODE_ENCRYPT;
-        break;
-      case Cipher.DECRYPT_MODE:
-      case Cipher.UNWRAP_MODE:
-        this.opMode = NATIVE_MODE_DECRYPT;
-        break;
-      default:
-        throw new InvalidAlgorithmParameterException("Unsupported cipher mode " + jceOpMode);
+    this.sameKey = this.key != null && MessageDigest.isEqual(this.key, encodedKey);
```
geedo0, line 366: You don't get anything from comparing the MessageDigest, just compare the keys directly. Computing the hashes still requires traversing both keys and doing math on top of that. We also risk the odd issue of collision by comparing like this.

amirhosv: I think the naming is confusing here. [MessageDigest.isEquals](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/security/MessageDigest.java#L471) is same as Arrays.equals, except it's constant time for both equal and not equal cases.

benchmarks/lib/build.gradle.kts
```diff
@@ -59,6 +62,7 @@ jmh {
     duplicateClassesStrategy.set(DuplicatesStrategy.WARN)
     jvmArgs.add("-DversionStr=${accpVersion}")
     jvmArgs.add("-Dcom.amazon.corretto.crypto.provider.registerSecureRandom=true")
+    //jvmArgs.add("-Dcom.amazon.corretto.crypto.provider.freeNativeContextEagerly=false")
```
WillChilds-Klein, line 65: Was this intentionally left in here?

amirhosv: I'll remove it.

README.md
```diff
@@ -315,7 +314,14 @@ Thus, these should all be set on the JVM command line using `-D`.
...
   Non-FIPS AWS-LC does not use CPU jitter for its DRBG seed's entropy, and therefore does not incur this initialization cost, therefore we register LibCryptoRng by default when not in FIPS mode.
-
+* `com.amazon.corretto.crypto.provider.freeNativeContextEagerly`
+  Takes in `true` or `false` (defaults to `true`). This flag only affects AES-GCM cipher.
+  If `true`, the underlying `EVP_CIPHER_CTX` of a GCM `Cipher` is freed as soon as possible;
+  otherwise, the underlying `EVP_CIPHER_CTX` is not released util the `Cipher` object is
+  garbage collected. When reusing the same `Cipher` object, it would be beneficial to set
+  this system property to `false` so that different encryption/decryption operations would
+  not require allocation and release of `EVP_CIPHER_CTX` structure. A common use case would
+  be having long-running threads that each would get its own instance of `Cipher` object.
```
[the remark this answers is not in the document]
amirhosv: There is no clear winner here. Depending on the usage pattern, each option improves one, and degrades another one. The following are the two usage patterns:
1) The Cipher object is instantiated for each operation and discarded after. Eager approach improves the performance up to 30% in this case.
2) The Cipher object is cached and reused. The lazy approach improves the performance in this case.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -307,103 +314,115 @@ protected void engineInit(
...
-          new GCMParameterSpec(
-              DEFAULT_TAG_LENGTH, ((IvParameterSpec) algorithmParameterSpec).getIV());
-    } else {
-      throw new InvalidAlgorithmParameterException(
-          "I don't know how to handle a " + algorithmParameterSpec.getClass());
-    }
+    final int opMode = checkOperation(jceOpMode);
 
-    byte[] encodedKey = null;
-    if (jceKey != key) {
```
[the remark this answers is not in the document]
amirhosv: This check only saves time if the Cipher object is reused AND the same AES key is use AND the key is the same Java object; in all other cases, keeping `jceKey` around not beneficial. From security perspective, `jceKey` is a cryptographic material and it's not a good practice to keep a reference to a secret object for longer than it is needed.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -33,12 +33,6 @@ final class AesGcmSpi extends CipherSpi {
     Loader.load();
   }
 
-  /**
-   * The number a times a key must be reused prior to keeping it in native memory rather than
-   * freeing it each time.
-   */
-  private static final int KEY_REUSE_THRESHOLD = 1;
```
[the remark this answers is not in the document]
amirhosv: The `HYBRID` mode, which is the default, preserves the existing behavior and our benchmarks show that there is no regression. When threshold is one, the same effect can be achieved by using a single boolean variable (`sameKey`).

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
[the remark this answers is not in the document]
SalusaSecondus: NOT A CONTRIBUTION

---

Consider the common case of a single Cipher object being used to encrypt/decrypt a lot of data with the same key but different parameters (iv or gcm). In this case you can absolutely expect to see the exact same Java key object provided each time. What's more, I would expect that the majority of the time that a cipher object is reused, it is reused with the same Java key object.

You're already keeping the cryptographic materials around in `private byte[] key`, so minimizing it here doesn't help. What's more, as you're calling `getEncoded()` each time for the comparison, you're resulting in more copies of the cryptographic material spilling to memory than before this change. (In other words, your change will result in more copies of the key in memory rather than fewer.)

---

NOT A CONTRIBUTION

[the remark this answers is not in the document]
amirhosv: Despite not having any impact on performance, I brought back the logic. I'm not sure about "common" this pattern is, since we have many internal customers that prefer not reuse a Cipher object so that they don't have to deal with thread-safety issues.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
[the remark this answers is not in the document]
amirhosv: Yes, I did the run the existing benchmarks and as expected, there is no regression. The existing benchmarks utilize the same object so other patterns of usage are not measured. Here are the results of the existing benchmarks:
```
./gradlew -PaccpVersion="2.3.2" -PincludeBenchmark="AesGcm" lib:jmh
Benchmark                         (chunkSize)  (keyBits)   Mode  Cnt     Score    Error  Units
AesGcmOneShot.oneShot1MiBDecrypt          N/A        128  thrpt    5  2600.963 ± 88.948  ops/s
AesGcmOneShot.oneShot1MiBDecrypt          N/A        256  thrpt    5  2179.269 ± 31.994  ops/s
AesGcmOneShot.oneShot1MiBEncrypt          N/A        128  thrpt    5  2606.161 ± 75.525  ops/s
AesGcmOneShot.oneShot1MiBEncrypt          N/A        256  thrpt    5  2190.634 ± 68.482  ops/s
AesGcmUpdate.updateDecrypt                 16        128  thrpt    5   734.160 ± 26.745  ops/s
AesGcmUpdate.updateDecrypt                 16        256  thrpt    5   745.288 ± 24.561  ops/s
AesGcmUpdate.updateDecrypt                256        128  thrpt    5  1550.954 ± 47.719  ops/s
AesGcmUpdate.updateDecrypt                256        256  thrpt    5  1704.490 ± 80.432  ops/s
AesGcmUpdate.updateEncrypt                 16        128  thrpt    5    67.282 ±  5.458  ops/s
AesGcmUpdate.updateEncrypt                 16        256  thrpt    5    66.709 ±  2.305  ops/s
AesGcmUpdate.updateEncrypt                256        128  thrpt    5   789.663 ± 19.540  ops/s
AesGcmUpdate.updateEncrypt                256        256  thrpt    5   737.160 ± 31.125  ops/s

./gradlew -PaccpLocalJar="../../build/cmake/AmazonCorrettoCryptoProvider.jar" -PnativeContextReleaseStrategy="HYBRID" -PincludeBenchmark="AesGcm" lib:jmh
Benchmark                         (chunkSize)  (keyBits)   Mode  Cnt     Score     Error  Units
AesGcmOneShot.oneShot1MiBDecrypt          N/A        128  thrpt    5  2611.790 ± 141.192  ops/s
AesGcmOneShot.oneShot1MiBDecrypt          N/A        256  thrpt    5  2175.562 ±  68.551  ops/s
AesGcmOneShot.oneShot1MiBEncrypt          N/A        128  thrpt    5  2639.222 ±  63.212  ops/s
AesGcmOneShot.oneShot1MiBEncrypt          N/A        256  thrpt    5  2196.239 ± 317.460  ops/s
AesGcmUpdate.updateDecrypt                 16        128  thrpt    5   787.723 ±  28.777  ops/s
AesGcmUpdate.updateDecrypt                 16        256  thrpt    5   675.053 ±  28.231  ops/s
AesGcmUpdate.updateDecrypt                256        128  thrpt    5  1887.667 ±  64.505  ops/s
AesGcmUpdate.updateDecrypt                256        256  thrpt    5  1652.982 ±  95.543  ops/s
AesGcmUpdate.updateEncrypt                 16        128  thrpt    5    67.716 ±   1.974  ops/s
AesGcmUpdate.updateEncrypt                 16        256  thrpt    5    66.032 ±   4.465  ops/s
AesGcmUpdate.updateEncrypt                256        128  thrpt    5   789.979 ±  38.398  ops/s
AesGcmUpdate.updateEncrypt                256        256  thrpt    5   738.422 ±  28.252  ops/s
```

benchmarks/lib/build.gradle.kts
```diff
@@ -1,6 +1,8 @@
 val accpVersion: String? by project
 val accpLocalJar: String by project
 val fips: Boolean by project
+val includeBenchmark: String by project
```
geedo0, line 4: Would be more general if you could specify multiple benchmarks.

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/CipherReuse.java
```diff
@@ -0,0 +1,75 @@
...
+    public void newInstance(Blackhole blackhole) throws Exception {
+        blackhole.consume(encrypt(getAesGcmCipherFromAccp()));
+    }
+
+    @Benchmark
+    public void reuse(Blackhole blackhole) throws Exception {
+        blackhole.consume(encrypt(shared));
+    }
+
+    byte[] encrypt(Cipher cipher) throws Exception {
```
geedo0, line 51: pedantic: the method name says encrypt, but you do both directions.

src/com/amazon/corretto/crypto/provider/Utils.java
```diff
@@ -572,4 +572,27 @@ static <T> T requireNonNull(final T obj, final String message) {
...
   }
+
+  enum NativeContextReleaseStrategy {
+    HYBRID,
+    LAZY,
+    EAGER
+  }
+
+  static NativeContextReleaseStrategy getNativeContextReleaseStrategyProperty(
+      final String propertyName) {
```
geedo0, line 583: This interface doesn't make sense. When would you ever need to call this with a different property name? Supporting that seems error-prone. Suggest referencing a constant within this method and removing the indirection inside AmazonCorrettoCryptoProvider.

geedo0 (approved):

benchmarks/lib/build.gradle.kts
amirhosv: This string is can be a regular expression. I updated the README to reflect this fact.

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/CipherReuse.java
amirhosv: Renamed to `encryptDecrypt`

src/com/amazon/corretto/crypto/provider/Utils.java
amirhosv: Done.

geedo0 (approved):

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/CipherReuse.java
```diff
@@ -0,0 +1,75 @@
...
+        final byte[] cipherText = cipher.doFinal(input);
+        cipher.init(Cipher.DECRYPT_MODE, sk, new GCMParameterSpec(128, iv));
+        cipher.updateAAD(aad);
+        return cipher.doFinal(cipherText);
+    }
+
+    private Cipher getAesGcmCipherFromAccp() {
+        try {
+            return Cipher.getInstance(AES_GCM, AmazonCorrettoCryptoProvider.INSTANCE);
+        } catch (final Exception e) {
```
WillChilds-Klein, line 71: very minor nit: both callers of this function are themselves in functions that throw checked exceptions, so you don't need to wrap `e` in an unchecked exception here.

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -739,16 +775,31 @@ private int engineDecryptFinal(
...
+          output,
+          outputOffset,
+          tagLength,
+          key,
+          iv,
+
+          // The cost of calling decryptAADBuf.getDataBuffer() when its buffer is empty is
+          // significant for 16-byte decrypt operations (approximately a 7% performance hit).
+          // To avoid this, we reuse the same empty array
+          decryptAADBuf.size() != 0 ? decryptAADBuf.getDataBuffer() : EMPTY_ARRAY,
```
WillChilds-Klein, line 800: nit: why use `.size()` here and `.isEmpty()` on L778 and L754? is there a difference?

tst/com/amazon/corretto/crypto/provider/test/AesTest.java
```diff
@@ -1142,61 +1188,110 @@ public void safeReuse() throws Throwable {
...
     c.init(Cipher.DECRYPT_MODE, key, spec2);
     assertFalse((boolean) sneakyGetField(spi, "contextInitialized"));
-    assertNotNull(sneakyGetField(spi, "context"));
     assertArrayEquals(plaintext2, c.doFinal(ciphertext2));
+    if (saveNativeContext(spi)) {
+      assertNotNull(sneakyGetField(spi, "context"));
+    } else {
+      assertNull(sneakyGetField(spi, "context"));
+    }
+
```
WillChilds-Klein, line 1224: very minor nit: would it make sense to pull this (heavily duplicated) assertion logic into a `assertNativeContextOk` function or something?

WillChilds-Klein (approved):

src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
amirhosv: There is no difference. I rebased and this one was not updated. I'll change it.

tst/com/amazon/corretto/crypto/provider/test/AesTest.java
amirhosv: Thanks for the suggestion.

WillChilds-Klein (approved):

geedo0 (approved):

PR #330, opened 2023-09-18
Remove JML Markup

geedo0:
*Issue #, if available:*
N/A

*Description of changes:*
This JML markup is no longer maintained and/or used by the team. It's
not wired up to our build in any way so we fear that it's contributing
more to bitrot than providing reliable coverage.


geedo0:
I was going to do that in a follow-up to avoid adding code formatting changes to this commit.

amirhosv (approved):

PR #333, opened 2023-09-26
Replace MessageDigest.isEqual with our own implementation

amirhosv:
*Description of changes:*

The documentation of [MessageDigest.isEqual](https://docs.oracle.com/javase/8/docs/api/java/security/MessageDigest.html#isEqual-byte:A-byte:A-) does not guarantee constant time despite its implementation for most JDKs is constant time.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/ConstantTime.java
+++ b/src/com/amazon/corretto/crypto/provider/ConstantTime.java
@@ -4,45 +4,47 @@ package com.amazon.corretto.crypto.provider;
 
 /** Contains several constant time utilities */
 final class ConstantTime {
+  private static int LONG_UNSIGNED_SHIFT = Long.SIZE - 1;
+
   private ConstantTime() {
     // Prevent instantiation
   }
 
   /** Equivalent to {@code val != 0 ? 1 : 0} */
-  static final int isNonZero(int val) {
+  static int isNonZero(int val) {
     return ((val | -val) >>> 31) & 0x01; // Unsigned bitshift
   }
 
   /** Equivalent to {@code val == 0 ? 1 : 0} */
-  static final int isZero(int val) {
+  static int isZero(int val) {
     return 1 - isNonZero(val);
   }
 
   /** Equivalent to {@code val < 0 ? 1 : 0} */
-  static final int isNegative(int val) {
+  static int isNegative(int val) {
     return (val >>> 31) & 0x01;
   }
 
   /** Equivalent to {@code x == y ? 1 : 0} */
-  static final int equal(int x, int y) {
+  static int equal(int x, int y) {
     final int difference = x - y;
     // Difference is 0 iff x == y
     return isZero(difference);
   }
 
   /** Equivalent to {@code x > y ? 1 : 0} */
-  static final int gt(int x, int y) {
+  static int gt(int x, int y) {
     // Convert to long to avoid underflow
     final long xl = x;
     final long yl = y;
     final long difference = yl - xl;
     // If xl > yl, then difference is negative.
     // Thus, we can just return the sign-bit
-    return (int) ((difference >>> 63) & 0x01); // Unsigned bitshift
+    return (int) (difference >>> LONG_UNSIGNED_SHIFT); // Unsigned bitshift
   }
 
   /** Equivalent to {@code selector != 0 ? a : b} */
-  static final int select(int selector, int a, int b) {
+  static int select(int selector, int a, int b) {
     final int mask = isZero(selector) - 1;
     // Mask == -1 (all bits 1) iff selector != 0
     // Mask ==  0 (all bits 0) iff selector == 0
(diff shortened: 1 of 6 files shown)


src/com/amazon/corretto/crypto/provider/Utils.java
```diff
@@ -572,4 +574,61 @@ static <T> T requireNonNull(final T obj, final String message) {
...
+   *
+   * @param trusted
+   * @param trustedStart
+   * @param trustedLen
+   * @param other
+   * @param otherStart
+   * @param otherLen
+   * @return true iff all the bytes in the specified ranges are equal
+   */
+  static boolean constantEquality(
```
alexw91, line 595: Nitpick: If this function is intended to be constant time, then it should be reflected in the name. Could this function be renamed to `constantTimeEquality` or `constantTimeEquals`.

amirhosv: I'll rename in the next patch.

src/com/amazon/corretto/crypto/provider/Utils.java
```diff
@@ -572,4 +574,61 @@ static <T> T requireNonNull(final T obj, final String message) {
...
+   *
+   * @param trusted
+   * @param trustedStart
+   * @param trustedLen
+   * @param other
+   * @param otherStart
+   * @param otherLen
+   * @return true iff all the bytes in the specified ranges are equal
+   */
+  static boolean constantTimeEquals(
```
alexw91, line 595: This function should be moved to [`ConstantTime.java`](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/ConstantTime.java)

amirhosv: Done.

src/com/amazon/corretto/crypto/provider/ConstantTime.java
```diff
@@ -51,4 +59,53 @@ static final int select(int selector, int a, int b) {
...
+      final int otherLen) {
+    if (otherLen == 0) {
+      return trustedLen == 0;
+    }
+
+    int result = trustedLen - otherLen;
+
+    for (int i = 0; i < trustedLen; i++) {
+      final int trustedIndex = trustedStart + i;
+      final int otherIndex = otherStart + gtPositive(otherLen, i) * i;
```
apetcher-amazon, line 90: Why is all this necessary? Do we need to hide the length?

amirhosv: The implementation is mostly borrowed from [JDK](https://github.com/corretto/corretto-11/blob/develop/src/java.base/share/classes/java/security/MessageDigest.java#L492). Avoiding branches inside this loop would minimize the risk of compiler/run-time environment optimize this loop by short circuiting it and as a result, this loop would be constant time and only depend on the length of the trusted input. I guess hiding the length is also a nice by-product.

apetcher-amazon: But if you are allowed to leak the length, then you could simplify this by doing some branching calculations at the beginning to make sure the lengths are correct, and then do the constant time comparison of the array contents. Right?

amirhosv: It will definitely simplify it. If the user can't learn the length of the key from one Cipher object, they can compare the timing from two different Cipher objects initialized with different key sizes and get that information. I would rely on your expertise to see if it is OK to leak the length, by first checking if the lengths are different and returning `false`; otherwise proceeding with the rest of the logic.

apetcher-amazon: I can't say whether you need to hide the length---that is up to you. All I'm saying is that the additional complexity of this code is probably not justified unless hiding the length is a requirement. FWIW, I don't think that hiding the length is usually expected for an operation like this, and as long as you document what is leaked to side channels, it should be okay either way. Though I don't have an explanation for why MessageDigest::isEqual in OpenJDK hides the length of the second array. Maybe there is a good reason for it that I'm missing.

src/com/amazon/corretto/crypto/provider/ConstantTime.java
```diff
@@ -51,4 +59,53 @@ static final int select(int selector, int a, int b) {
...
+  static boolean equals(
+      final byte[] trusted, final int trustedStart, final int trustedLen, final byte[] other) {
+    return equals(trusted, trustedStart, trustedLen, other, 0, other.length);
+  }
+
+  static boolean equals(final byte[] trusted, final byte[] other) {
+    if (trusted == other) {
+      return true;
+    }
+    if (trusted == null || other == null) {
```
WillChilds-Klein, line 106: we could also return false here if `trusted.length != other.length`, right?

amirhosv: It could. I moved it to the "main" one.

src/com/amazon/corretto/crypto/provider/ConstantTime.java
```diff
@@ -51,4 +53,53 @@ static final int select(int selector, int a, int b) {
...
+      final byte[] other,
+      final int otherStart,
+      final int otherLen) {
+
+    if (trustedLen != otherLen) {
+      return false;
+    }
+
+    int result = 0;
+
```
geedo0, line 83: Not entirely sure what you're definition of trust is, but we're extending a lot of it here. No size checking, no checking for negative indices... It would help to define exactly what we're trusting.

amirhosv: In the first implementation, `equals(a, b)` would not necessarily take the same time as `equals(b, a)` since the the length of the first input would determine the number of iterations. In that implementation, one needed to ensure the first argument is something that has been vetted: a large untrusted buffer as first arg could potentially be problematic.
With the new implementation, the performance of `equals(a, b)` is exactly the same as `equals(b, a)` so the "trusted" concept is irrelevant now. I forgot to update the comment and will do it now.

geedo0 (approved):

src/com/amazon/corretto/crypto/provider/ConstantTime.java
```diff
@@ -51,4 +53,44 @@ static final int select(int selector, int a, int b) {
...
+    for (int i = 0; i < aLen; i++) {
+      result |= a[aStart + i] ^ b[bStart + i];
+    }
+
+    return result == 0;
+  }
+
+  static boolean equals(final byte[] a, final int aStart, final int aLen, final byte[] b) {
```
WillChilds-Klein, line 83: nit: do we want to do null checks here too? right now we'd NPE on L84 if `b` is null

amirhosv: It's a method used internally, but I'll add the check. We should invest in CheckerFramework to avoid these runtime checks for internal methods and statically ensure that all use cases pass none null values.

WillChilds-Klein (approved):

geedo0 (approved):

PR #334, opened 2023-09-27
Let ECDSA Signature objects accept parameters

SalusaSecondus:
The OpenJDK expects ECDSA signature objects to support calls to `setParameter()`. This can be seen in their [SignatureParameters](https://github.com/openjdk/jdk/blob/master/test/jdk/sun/security/ec/SignatureParameters.java) test.  The proposed change matches the default JDK behavior in that it requires that the set parameters be equal to those from the key.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/EcUtils.java
+++ b/src/com/amazon/corretto/crypto/provider/EcUtils.java
@@ -220,6 +220,18 @@ final class EcUtils {
     return name;
   }
 
+  static boolean ecParameterSpecsAreEqual(final ECParameterSpec a, final ECParameterSpec b) {
+    if (a == b) {
+      return true;
+    } else if (a == null || b == null) {
+      return false;
+    }
+    return a.getCofactor() == b.getCofactor()
+        && a.getOrder().equals(b.getOrder())
+        && a.getCurve().equals(b.getCurve())
+        && a.getGenerator().equals(b.getGenerator());
+  }
+
   static final class ECInfo {
     private final ThreadLocal<NativeGroup> group =
         new ThreadLocal<NativeGroup>() {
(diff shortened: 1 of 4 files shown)


tst/com/amazon/corretto/crypto/provider/test/EcGenTest.java
```diff
@@ -358,10 +362,13 @@ public static void assertECEquals(
...
-    assertEquals(expected.getOrder(), actual.getOrder(), message);
-    assertEquals(expected.getGenerator(), actual.getGenerator(), message);
-    assertEquals(expected.getCurve(), actual.getCurve(), message);
+    boolean result = false;
+    try {
+      result = TestUtil.sneakyInvoke(EC_UTILS_CLASS, "ecParameterSpecsAreEqual", expected, actual);
+    } catch (final Throwable t) {
+      throw new AssertionError(message, t);
+    }
+    assertTrue(message, result);
```
WillChilds-Klein, line 371: should this be `assertTrue(result, message)`?

https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertTrue-boolean-java.lang.String-

SalusaSecondus: Somehow the wrong `assertTrue` also got pulled in. Not sure how I did that. Thanks for the catch. I'll squash that patch and force push shortly.

src/com/amazon/corretto/crypto/provider/EcUtils.java
```diff
@@ -220,6 +220,18 @@ static String getNameByEncoded(final byte[] encoded) {
     return name;
   }
 
+  static boolean ecParameterSpecsAreEqual(final ECParameterSpec a, final ECParameterSpec b) {
+    if (a == b) {
+      return true;
+    } else if (a == null || b == null) {
```
WillChilds-Klein, line 226: nit: i doubt this will come up in practice, but should this function return `true` if _both_`a` and `b` are null?

WillChilds-Klein (approved):

src/com/amazon/corretto/crypto/provider/EcUtils.java
SalusaSecondus: Yes it should. This is both standard in Java and also matches the general assumption that the equality is symmetric and so having `ecParameterSpecsAreEqual(param, param)` ever *not* equal `true` would be an unwelcome surprise.

geedo0 (approved):

PR #336, opened 2023-09-27
Fix ByteBuffer position handling

SalusaSecondus:
PR #296 introduced a bug into `ByteBuffer` handling for the AES/GCM decrypt case. Previously, it would correctly update the position of the `ByteBuffer` to mark that all data was consumed but now it leaves it unchanged. This PR makes the following changes:

1. Updates unit tests to prevent regression
2. Moves data directly from the `ByteBuffer` to the underlying `AccessibleByteArrayOutputStream` (which is logically equivalent to what `engineUpdate(byte[], int, int, byte[], int)` does anyway).
    1. This method fixes the `ByteBuffer` position issue
    2. Eliminates yet another byte array copy (which was the motivation for PR # 296 in the first place)

### DIFF
--- a/tst/com/amazon/corretto/crypto/provider/test/AESGenerativeTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/AESGenerativeTest.java
@@ -281,7 +281,9 @@ public class AESGenerativeTest {
       case 0:
         {
           jceCipherEncrypt.updateAAD(buf.duplicate());
-          amzCipherEncrypt.updateAAD(buf.duplicate());
+          final ByteBuffer amzBuf = buf.duplicate();
+          amzCipherEncrypt.updateAAD(amzBuf);
+          assertByteBufferEmpty(amzBuf);
 
           Channels.newChannel(aadData).write(buf);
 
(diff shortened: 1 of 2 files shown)


tst/com/amazon/corretto/crypto/provider/test/AESGenerativeTest.java
```diff
@@ -281,7 +281,9 @@ private void updateAAD(Random r, ByteArrayOutputStream aadData) throws Throwable
       case 0:
         {
           jceCipherEncrypt.updateAAD(buf.duplicate());
-          amzCipherEncrypt.updateAAD(buf.duplicate());
+          final ByteBuffer amzBuf = buf.duplicate();
+          amzCipherEncrypt.updateAAD(amzBuf);
+          assertByteBufferEmpty(amzBuf);
```
amirhosv, line 286: The change is for decryption. Shouldn't assertions be about buffers for decryption rather than encryption?

SalusaSecondus: I've updated *all* the buffer test cases that I could find. This bug crept in because our test coverage missed this case originally. So, my change not only introduces tests for this specific bug (failing to update position on decryption), it also improves the tests for all other cases to ensure we don't introduce a variant of this bug.

amirhosv: It would be good to add it for aad update too: https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/tst/com/amazon/corretto/crypto/provider/test/AESGenerativeTest.java#L524

SalusaSecondus: Thank you. I missed that one!

geedo0 (approved):
Nice catch!

amirhosv (approved):

PR #338, opened 2023-10-03
Avoid buffering cipher text for one-shot AES-GCM decrypt

amirhosv:
*Description of changes:*

AES-GCM decrypt buffers the ciphertext before decryption. This buffering can be avoided for one-shot decryption. Our benchmarks show performance improvement of AES-GCM one-shot decryption by 33% for 128 bit key and 26% for 256 bit key.

Before this PR:
```
Benchmark                         (keyBits)   Mode  Cnt     Score     Error  Units
AesGcmOneShot.oneShot1MiBDecrypt        128  thrpt    5  1953.188 ±  59.826  ops/s
AesGcmOneShot.oneShot1MiBDecrypt        256  thrpt    5  1710.003 ± 108.368  ops/s
```

After this PR:
```
Benchmark                         (keyBits)   Mode  Cnt     Score    Error  Units
AesGcmOneShot.oneShot1MiBDecrypt        128  thrpt    5  2606.189 ± 61.312  ops/s
AesGcmOneShot.oneShot1MiBDecrypt        256  thrpt    5  2161.426 ± 41.597  ops/s
```

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AccessibleByteArrayOutputStream.java
+++ b/src/com/amazon/corretto/crypto/provider/AccessibleByteArrayOutputStream.java
@@ -49,6 +49,16 @@ class AccessibleByteArrayOutputStream extends OutputStream implements Cloneable 
     count += len;
   }
 
+  /**
+   * Same as write, except it does not grow the buffer size beyond what's needed. Use this method if
+   * you know it would be the last time something needs to be written to the buffer.
+   */
+  void finalWrite(final byte[] b, final int off, final int len) {
+    growCapacity(count + len, true);
+    System.arraycopy(b, off, buf, count, len);
+    count += len;
+  }
+
   @Override
   public void write(final int b) {
     growCapacity(count + 1);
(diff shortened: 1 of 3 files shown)


geedo0:
There are two optimizations here. Do you know how big of an impact the logic for avoiding the array resize had? More curious than anything.

geedo0 (approved):

amirhosv:
> There are two optimizations here. Do you know how big of an impact the logic for avoiding the array resize had? More curious than anything.

I run the benchmarks for multi-part decryption, but no measurable impact there.

amirhosv:
> > There are two optimizations here. Do you know how big of an impact the logic for avoiding the array resize had? More curious than anything.
> 
> I run the benchmarks for multi-part decryption, but no measurable impact there.

After looking into our benchmark, I see why there was no impact: [the benchmarks](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/AesGcmUpdate.java#L43) use the same `Cipher` object with the same payload size. When the buffer is [reset](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/AccessibleByteArrayOutputStream.java#L71) it's not de-allocated, which means during the warm-up of the benchmark the buffer gets allocated and we don't see its cost when the measurement actually happens.

WillChilds-Klein (approved):

PR #345, opened 2023-11-02
Control the version of AWS-LC used for FIPS and non-FIPS builds

amirhosv:
*Description of changes:*
ACCP FIPS uses a different version of AWS-LC. In this PR, we allow the CI tasks to switch to the required version when building in FIPS or non-FIPS mode.

Our legacy build, still requires the use of Git submodules to track the version of AWS-LC used. Once that is deprecated, we can stop using Git submodules and only rely on the value of `ext.awLcGitVersionId` in `build.gradle` as the only place for tracking AWS-LC's version using for ACCP and ACCP-FIPS.

With this PR, we do not need to update the Git submodule for AWS-LC since the build always check the version specified by `ext.awLcGitVersionId` is checked out.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,11 @@ plugins {
 group = 'software.amazon.cryptools'
 version = '2.3.2'
 ext.isFips = Boolean.getBoolean('FIPS')
+if (ext.isFips) {
+    ext.awsLcGitVersionId = 'AWS-LC-FIPS-2.0.0'
+} else {
+    ext.awsLcGitVersionId = 'v1.17.0'
+}
 ext.isLegacyBuild = Boolean.getBoolean('LEGACY_BUILD')
 
 ext.lcovIgnore = System.properties['LCOV_IGNORE']
@@ -170,6 +175,17 @@ task buildAwsLc {
     doFirst {
         if (file(awslcSrcPath).list().size() == 0) {
             throw new GradleException("aws-lc dir empty! run 'git submodule update --init --recursive' to populate.")
+        }
+
+        if (!isLegacyBuild) {
+            exec {
+                workingDir awslcSrcPath
+                commandLine "git", "fetch", "--tags"
+            }
+            exec {
+                workingDir awslcSrcPath
+                commandLine "git", "checkout", awsLcGitVersionId
+            }
         }
         mkdir "${buildDir}/awslc"
         mkdir sharedObjectOutDir
--- a/README.md
+++ b/README.md
@@ -209,7 +209,9 @@ Building this provider requires a 64 bit Linux or MacOS build system with the fo
 **FIPS builds are still experimental and are not yet ready for production use.**
 
 By providing `-DFIPS=true` to `gradlew` you will cause the entire build to be for a "FIPS mode" build.
-The only significant difference is that AWS-LC is built with `FIPS=1`.
+The FIPS builds use a different version of AWS-LC along with `FIPS=1` build flag. Not all releases of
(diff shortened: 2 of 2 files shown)


build.gradle
```diff
@@ -15,6 +15,11 @@ plugins {
 group = 'software.amazon.cryptools'
 version = '2.3.2'
 ext.isFips = Boolean.getBoolean('FIPS')
+if (ext.isFips) {
+    ext.awsLcGitVersionId = 'AWS-LC-FIPS-2.0.0'
+} else {
+    ext.awsLcGitVersionId = 'v1.17.0'
```
WillChilds-Klein, line 21: have we updated the submodule commit ID to match this?

WillChilds-Klein (dismissed):

build.gradle
amirhosv: Updated it to the latest version, v1.17.1.

amirhosv: It turns out that in v1.17.1, the RSA struct is opaque and ACCP has problem building with it. Upgrading to v1.17.1 will be done in another PR.

geedo0: Why did this else statement need to exist? This PR purports to only make changes to the FIPS build. Shouldn't the non-FIPS build continue to do everything as it used to do? I don't like that we now have to track the git version tag of LC to use in the gradle script, that seems really brittle and we're already seeing evidence of that in this thread.

build.gradle
```diff
@@ -171,6 +176,17 @@ task buildAwsLc {
         if (file(awslcSrcPath).list().size() == 0) {
             throw new GradleException("aws-lc dir empty! run 'git submodule update --init --recursive' to populate.")
         }
+
+        if (!isLegacyBuild) {
```
geedo0, line 180: Again, why do we have to do this for the non-FIPS build as well?

geedo0 (changes requested):

build.gradle
amirhosv: Ideally we want to stop using Git submodule to track our dependency on AWS-LC and use the values here as the source of truth. What's blocking us is a legacy build that has a hard time reaching out and pulling the right version of AWS-LC from Github. Once that's done, then we will drop submodule and only rely on this value.

Since we are using different version of AWS-LC for FIPS and non-FIPS builds, Git submodule is not a suitable solution.

build.gradle
amirhosv: We do this only for non-Legacy builds. There is only one legacy build that tests ACCP with an old version GCC using an Ubuntu 10.04 image.

build.gradle
geedo0: If that's the case, have we considered how this impacts our agility? AWS-LC iterates on these versions pretty infrequently based on a cursory glance at their [releases](https://github.com/aws/aws-lc/tags). Far less frequently then we pulled in submodule changes.

amirhosv: The mechanism for updating submodule is the same as updating these values in the code: create a PR and get it merged.

geedo0: What I'm saying is that previously, you can point to /any/ AWS-LC commit. Now we're limited to whatever they decide to tag with a release. Though I suppose you could still use a commit hash as a git ref and achieve the same effect.

geedo0 (approved):

build.gradle
WillChilds-Klein: As a matter of hygiene, I think it would be desirable to track LC by releases rather than individual commits (although it's good to have "break glass" options to reference individual commits). Using specific versions should help with stability, bug reports, etc.

WillChilds-Klein (approved):

PR #347, opened 2023-11-06
Add Cryptofuzz to CI

geedo0:
*Issue #, if available:*
CryptoAlg-2133

*Description of changes:*
This adds a new CI workflow to ACCP which runs Cryptofuzz for a fixed amount of time on each PR and reports any failures. Cryptofuzz is a fuzzing framework backed by LLVM's LibFuzzer project. It performs fuzz testing on cryptographic libraries at the granularity of cryptographic algorithms and also performs differential testing against the cryptographic libraries that it fuzzes.

The implementation here is unapologetically derivative of the AWS-LC integration with Cryptofuzz. It follows an architecture reminiscent of our existing CI stacks, but differs in that it runs on a newly defined docker image `amazonlinux-2_clang-11x_cryptofuzz_x86` which uses clang as the compiler and has Cryptofuzz related assets baked into it. Clang is necessary here because LibFuzzer (invoked via -fsantize=fuzzer) is only available from LLVM based compilers.

The ACCPGitHubFuzzCIStack spins up a VPC containing an EFS filesystem which contains the outputs of any Cryptofuzz findings as well as a running corpus of interesting inputs discovered over time. Codebuild containers of the fuzzer mount this filesystem at runtime.

In the CI workflow, it builds ACCP and "links" it into the Cryptofuzz binary. It then runs Cryptofuzz to completion. We depart from the AWS-LC template by explicitly specifying the algorithms to fuzz because ACCP supports a much smaller subset of algorithms as OpenSSL. Without this, the fuzzer will end up fuzzing OpenSSL more than ACCP due to the random nature of LibFuzzer.

To get ACCP working with Cryptofuzz, we adapt the existing Java module in CryptoFuzz to support installing ACCP as a JCE provider. We also configure LSAN suppressions to avoid false positives that would break the build.


tests/ci/run_cryptofuzz.sh
```diff
@@ -11,50 +11,56 @@ set -u
...
+export LINK_FLAGS="${LINK_FLAGS} -Wl,-rpath=${JDK_PATH}/lib/server/:${AWSLC_LIBRARY_PATH}"
+export CXXFLAGS="$CXXFLAGS -DCRYPTOFUZZ_JAVA -DJAVA_WITH_ECDSA -DCRYPTOFUZZ_ACCP"
+cd "${CRYPTOFUZZ_SRC}/modules/java/"
+make -j$(nproc)
+
+# LSAN suppressions due to false positives
+cd ${FUZZ_ROOT}
+echo 'leak:libjvm.so' >> lsan.supp
+echo 'leak:libz.so' >> lsan.supp
+echo 'leak:libzip.so' >> lsan.supp
```
WillChilds-Klein, line 38: i understand why libjvm would trigger false positives for LSAN, but why would compression libraries do so?

tests/ci/run_cryptofuzz.sh
```diff
@@ -11,50 +11,56 @@ set -u
...
-# Build the common OpenSSL module with AWS-LC
-cd "${CRYPTOFUZZ_SRC}/modules/openssl"
-make "-j${NUM_CPU_THREADS}"
-
-export CFLAGS="${CFLAGS} -I $OPENSSL_INCLUDE_PATH"
-export CXXFLAGS="${CXXFLAGS} -I $OPENSSL_INCLUDE_PATH"
+# Required linker flag set via this variable
 export LIBFUZZER_LINK="-fsanitize=fuzzer"
+# Explicitly disable OpenSSL module since it creates symbol conflicts
+export CXXFLAGS="$CXXFLAGS -DCRYPTOFUZZ_NO_OPENSSL"
```
WillChilds-Klein, line 55: is the implication here that we won't be able to do differential fuzzing against OpenSSL until we statically link AWS-LC to libaccp?

tests/ci/run_cryptofuzz.sh
```diff
@@ -67,12 +73,18 @@ SRC_CORPUS="$CRYPTOFUZZ_SEED_CORPUS"
 # 5 minutes for building AWS-LC and Cryptofuzz
 # 16 minutes (1000 seconds) of fuzzing
 # 24 minutes of cleanup and merging in new inputs
-TIME_FOR_EACH_FUZZ=1000
+TIME_FOR_EACH_FUZZ=${TIME_FOR_EACH_FUZZ:=1000}
 
 # Some fuzz tests can take a while but still pass. This is a tradeoff: less false positive noise, but some inputs that take
 # a long time could lead to a denial of service avenue. We're mostly interested in correctness and memory safety at this
 # time so we're willing to take the fit on fuzz speed
```
WillChilds-Klein, line 80: minor nit: s/fit/hit/

tests/ci/run_cryptofuzz.sh
```diff
@@ -67,12 +73,18 @@ SRC_CORPUS="$CRYPTOFUZZ_SEED_CORPUS"
...
-FUZZ_TEST_TIMEOUT=30
+FUZZ_TEST_TIMEOUT=${FUZZ_TEST_TIMEOUT:=30}
+
+# Cryptofuzz lets us specify the operations and algorithms to fuzz
+# We use this to narrow the fuzzing run to only what's supported by the Java/ACCP module rather than all of OpenSSL
+FUZZ_TEST_ADDITIONAL_ARGS="--operations=Digest,HMAC,ECDSA_Verify"
+FUZZ_TEST_ADDITIONAL_ARGS="${FUZZ_TEST_ADDITIONAL_ARGS} --digests=NULL,MD5,SHA1,SHA256,SHA384,SHA512"
```
WillChilds-Klein, line 86: what does NULL digest do here?

tests/ci/run_cryptofuzz.sh
```diff
@@ -67,12 +73,18 @@ SRC_CORPUS="$CRYPTOFUZZ_SEED_CORPUS"
...
+FUZZ_TEST_ADDITIONAL_ARGS="${FUZZ_TEST_ADDITIONAL_ARGS} --curves=secp192r1,secp256r1,secp384r1,secp256k1"
```
WillChilds-Klein, line 87: i don't think we [support](https://github.com/search?q=repo%3Acorretto%2Famazon-corretto-crypto-provider+secp192r1&type=code) secp192r1 anymore. suggest replacing that with secp521r1
```suggestion
FUZZ_TEST_ADDITIONAL_ARGS="${FUZZ_TEST_ADDITIONAL_ARGS} --curves=secp521r1,secp256r1,secp384r1,secp256k1"
```

WillChilds-Klein (changes requested):

tests/ci/run_cryptofuzz.sh
geedo0: These libraries are invoked by libjvm and they get flagged by LSAN if you don't suppress them. My theory is that libjvm's leakiness gets picked up by its dependencies because of similar issues around how it deals with memory.

tests/ci/run_cryptofuzz.sh
geedo0: Yes, but once we statically link AWS-LC it becomes a lot more clear cut how ld ties the symbols at run time and that's a straightforward PR. That said, we won't be able to reliably differential test against OpenSSL because of the symbol conflicts. If we bang at it for a lot longer, we can probably get it to work but I do question the value prop when the simple thing is to do is to just differential fuzz against something that has none of these problems.

tests/ci/run_cryptofuzz.sh
geedo0: We'll need to get this fixed upstream and then pull it in. I'll cut an issue against the LC repo.

tests/ci/run_cryptofuzz.sh
geedo0: This is how you exercise the `NONEwithECDSA` variant of the ECDSA operation which just EC signs the data you give it.

tests/ci/run_cryptofuzz.sh
geedo0: Anything we specify here needs to be the intersection of what we support and what [cryptofuzz supports](https://github.com/guidovranken/cryptofuzz/blob/master/modules/java/module.cpp#L390). Would be easy enough to develop, but that's a different project around uplifting their operation/algorithm support.

WillChilds-Klein: gotcha. but we don't support secp192r1 so can we drop it here?

tests/ci/run_cryptofuzz.sh
WillChilds-Klein: hm, it's a pretty inconsequential typo. not sure if it's worth chasing down

tests/ci/run_cryptofuzz.sh
WillChilds-Klein: > I do question the value prop when the simple thing is to do is to just differential fuzz against something that has none of these problems.

good point, and differential fuzzing against OpenSSL would be a little redundant if AWS-LC already does that. i suppose differential fuzzing would only be useful for us for compatibility checks against other java providers, but we've already discussed the limtations around that.

WillChilds-Klein (approved):

amirhosv (approved):

PR #353, opened 2023-11-22
Static linking to libcrypto

amirhosv:
*Description of changes:*

+ Static linking to libcrypto


WillChilds-Klein:
Out of curiosity, what's the size difference before/after this change for AmazonCorrettoCryptoProvider.jar and liaAmazonCorrettoCryptoProvider.so?

geedo0:
Can you provide some additional details on any manual validation you performed? In particular, I'd like to see that you unpacked the jar and saw a big accp.so file and did some validation to prove that LC is present in the symbol table (e.g. using readelf).

alexw91 (commented):
I think you should also update the documentation in `Loader.java` (and also fix the ASCII art diagrams that got mangled in a previous auto-format PR): 

https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/Loader.java#L192-L237

amirhosv:
> Can you provide some additional details on any manual validation you performed? In particular, I'd like to see that you unpacked the jar and saw a big accp.so file and did some validation to prove that LC is present in the symbol table (e.g. using readelf).

I updated the description with the info you asked.

amirhosv:
> Out of curiosity, what's the size difference before/after this change for AmazonCorrettoCryptoProvider.jar and libamazonCorrettoCryptoProvider.so?

The size of the jar is reduced by 19%.

build.gradle
```diff
@@ -558,7 +556,7 @@ task coverage_cpp_report {
         }
         exec {
             workingDir "${buildDir}/cmake-coverage"
-            commandLine 'gcovr', '-r', "${projectDir}", '--xml'
+            commandLine 'gcovr', '-r', "${projectDir}", '--xml', '--gcov-ignore-parse-errors'
```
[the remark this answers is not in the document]
amirhosv: The tool has issue parsing generated files as part of coverage calculation. It's a known issue: https://gcovr.com/en/master/guide/gcov_parser.html

geedo0 (approved):

build.gradle
```diff
@@ -523,30 +521,32 @@ task coverage_cpp_report {
...
-        }
-        exec {
-            workingDir "${buildDir}/cmake-coverage"
-            commandLine 'gcovr', '-r', "${projectDir}", '--xml'
-            standardOutput = new FileOutputStream("${buildDir}/reports/cpp/cobertura.xml")
-        }
-        exec {
-             workingDir projectDir
-             commandLine 'genhtml', '-o', "${buildDir}/reports/cpp", '--rc', 'genhtml_branch_coverage=1', "${buildDir}/reports/cpp/coverage.info", '--ignore-errors', lcovIgnore
+        if (!isLegacyBuild) {
```
WillChilds-Klein, line 524: why does linkage affect our coverage collection and reporting logic?

also, why do we now condition this on `isLegacyBuild` where we didn't before?

amirhosv: As part of this change, we started running coverage during legacy build. Legacy build has old versions of lcov and gcov, which are not very reliable. Since we already perform coverage in other CI tasks, we don't need to do it for legacy build.
Having said that, this change is reverted (including adding `--gcov-ignore-parse-errors`) and instead for regular builds we add `-fprofile-update=atomic` compile flag when computing coverage.

WillChilds-Klein (approved):

geedo0 (approved):

PR #360, opened 2024-02-06
Benchmarking ACCP when its bundled with JDK

amirhosv:
*Description of changes:*

The change allows to run benchmarks with ACCP when it's bundled with JDK.


geedo0 (dismissed):

geedo0 (approved):

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/BenchmarkUtils.java
```diff
@@ -2,75 +2,38 @@
...
-    }
-  }
-
-  static void installDefaultProviders() {
-    for (Provider provider : DEFAULT_PROVIDERS) {
-      Security.addProvider(provider);
-    }
-  }
-
-  static void removeAllProviders() {
```
WillChilds-Klein, line 71: I believe the original reason to remove/setup individual providers was to isolate utilities classes or other crypto instances that may be requested internally (_without_ specifying the provider explicitly) by providers. One instance of that can be found in [JDK11's MGF1](https://github.com/corretto/corretto-11/blob/1b08b745e538fb8f2c89d379382459622260f043/src/java.base/share/classes/sun/security/rsa/MGF1.java#L47), some others [here](https://github.com/search?q=repo%3Acorretto%2Fcorretto-11+%22.getInstance%22+path%3Asrc%2Fjava.base%2Fshare%2Fclasses%2F&type=code).

My concern here is that without isolating individual providers/provider suites, then we'll get less accurate measurements due to cross-provider instance sourcing. Do we need this particular change (installing all providers all the time) to satisfy the overall intent of the PR?

amirhosv: Thanks for catching that.

WillChilds-Klein (approved):

geedo0 (approved):

Issue #359, opened 2024-01-26
Allow configuration of tmpdir to avoid issues with noexec on java.io.tmpdir


christopherincanada:
**NOT A CONTRIBUTION**

ACCP uses `java.io.tmpdir` as the location to extract native libraries to. However, if that directory has been mounted with the `noexec` option, as is the case with the `/tmp` directory on many Linux systems, ACCP will fail to load with an error like the following:

```
Exception in thread "main" com.amazon.corretto.crypto.provider.RuntimeCryptoException: Unable to load native library
at Caused by: 
com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider.assertHealthy(AmazonCorrettoCryptoProvider.java:517)
java.lang.UnsatisfiedLinkError: /tmp/libamazonCorrettoCryptoProvider87cd5202885576a3.so: failed to map segment from shared object: Operation not permitted
```

The above results from a call to `AmazonCorrettoCryptoProvider.INSTANCE.assertHealthy()`.

To workaround this issue it would nice if ACCP would provide a system property like `com.amazon.corretto.crypto.provider.tmpdir` that could be used to specify which temp directory for ACCP to use. When not provided it could default to the value of the `java.io.tmpdir` system property.

Also open to other ideas for how to address the issue. Another solution is to simply update `java.io.tmpdir` to point it to a different directory not mounted with `noexec`, but that effects everything not just ACCP, which may not be desirable.

amirhosv:
@christopherincanada Thanks for reporting this. The proposal to add a new system property to specify a custom temp directory for ACCP is reasonable. I'll submit a PR that and it will be included in our next release.

amirhosv:
The PR for addressing the issue has been merged.

PR #361, opened 2024-02-06
Github issue #359

amirhosv:
*Issue #359*

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -669,6 +669,15 @@ add_custom_target(check-junit-AesGcmEager
 
     DEPENDS accp-jar tests-jar)
 
+add_custom_target(check-junit-DifferentTempDir
+    COMMAND ${TEST_JAVA_EXECUTABLE}
+    -Dcom.amazon.corretto.crypto.provider.tmpdir=${CMAKE_BINARY_DIR}/tmpdir
+        ${TEST_RUNNER_ARGUMENTS}
+        --select-class=com.amazon.corretto.crypto.provider.test.AesTest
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -669,6 +669,15 @@ add_custom_target(check-junit-AesGcmEager
 
     DEPENDS accp-jar tests-jar)
 
+add_custom_target(check-junit-DifferentTempDir
```
geedo0, line 672: Not convinced that this is a high signal test. There's no assertion anywhere that a different tmpdir was actually used. Just that these two tests succeeded when this property was set. I'm inclined to drop this test target.

README.md
```diff
@@ -344,6 +344,9 @@ Thus, these should all be set on the JVM command line using `-D`.
   encryption/decryption operations would not require allocation and release of `EVP_CIPHER_CTX`
   structure. A common use case would be having long-running threads that each would get its
   own instance of `Cipher` class.
+* `com.amazon.corretto.crypto.provider.tmpdir`
```
WillChilds-Klein, line 347: what happens if this points to a directory containing the libs we unpack into it? do they get overwritten or do the writes fail?

amirhosv: We actually create a new directory here first and extract the files in there. Upon exit, the directory is marked to be deleted.

WillChilds-Klein (approved):

alexw91 (approved):

CMakeLists.txt
amirhosv: Since the PR already has the two approvals, I'll drop this in another PR.

PR #363, opened 2024-02-13
Fix NullPointerException on invalid keys

SalusaSecondus:
ACCP can throw an incorrect `NullPointerException` when attempting to translate invalid keys with `null` formats. Per the JDK documentation this should throw an `InvalidKeySpecException` instead. By throwing the incorrect exception ACCP prevents the JCA from correctly falling back to providers which might support the keys.


geedo0 (approved):
Thanks!

WillChilds-Klein (approved):

PR #368, opened 2024-03-07
Fix IllegalArgumentException on Cipher.getParameters

SalusaSecondus:
`Cipher.getParameters()` is documented as returning “a combination of default and random parameter values used by the underlying cipher implementation if this cipher requires algorithm parameters but was not initialized with any.” However, if called on an AES/GCM/NoPadding cipher in ACCP, we get an exception because it expects the `iv` to be non-null.

This change detects an uninitialized cipher and generates a random 12 byte (96 bit) IV for use in the returned parameters when this happens. It also has a corresponding unit test.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
@@ -33,6 +33,7 @@ final class AesGcmSpi extends CipherSpi {
   }
 
   private static final int DEFAULT_TAG_LENGTH = 16 * 8;
+  private static final int DEFAULT_IV_LENGTH_BYTES = 12;
 
   /* Some random notes:
    * For decrypt mode, we buffer all data and process the decryption in doFinal;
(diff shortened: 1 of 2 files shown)


src/com/amazon/corretto/crypto/provider/AesGcmSpi.java
```diff
@@ -285,7 +286,13 @@ protected byte[] engineGetIV() {
   protected AlgorithmParameters engineGetParameters() {
     try {
       AlgorithmParameters parameters = AlgorithmParameters.getInstance("GCM");
-      parameters.init(new GCMParameterSpec(tagLength * 8, iv));
+      byte[] ivForParams = iv;
+      if (ivForParams == null) {
+        // We aren't initialized so we return default and random values
+        ivForParams = new byte[DEFAULT_IV_LENGTH_BYTES];
+        new SecureRandom().nextBytes(ivForParams);
```
amirhosv, line 293: It's better to use [LibCryptoRng](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/LibCryptoRng.java#L9) to ensure the secure random provider is backed by AWS-LC.

geedo0 (approved):

PR #370, opened 2024-03-20
Avoid accessing RSA struct's internals directly

amirhosv:
*Description of changes:*

AWS-LC has made the RSA struct opaque. This PR replaces direct accesses to internals of RSA struct with API calls. The FIPS builds do not have an API to disable blinding, therefore, for FIPS builds we still keep accessing the internals until a similar API is provided.

### DIFF
--- a/csrc/java_evp_keys.cpp
+++ b/csrc/java_evp_keys.cpp
@@ -496,11 +496,10 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_r
         if (privateExponentArr) {
             BigNumObj privExp = BigNumObj::fromJavaArray(env, privateExponentArr);
 
-            int res;
+            int res = 1;
             if (BN_is_zero(pubExp)) {
-                // RSA blinding can't be performed without |e|; 0 indicates |e|'s absence.
-                rsa->flags |= RSA_FLAG_NO_BLINDING;
-                res = RSA_set0_key(rsa, modulus, NULL, privExp);
+                // RSA blinding can't be performed without |e|.
+                rsa.set(RSA_new_private_key_no_e(modulus, privExp));
             } else {
                 res = RSA_set0_key(rsa, modulus, pubExp, privExp);
             }
--- a/csrc/keyutils.cpp
+++ b/csrc/keyutils.cpp
@@ -1,6 +1,8 @@
 // Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 #include "keyutils.h"
+#include "auto_free.h"
+#include "bn.h"
 #include <openssl/ec.h>
 #include <openssl/err.h>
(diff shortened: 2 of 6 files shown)


csrc/java_evp_keys.cpp
```diff
@@ -496,11 +496,10 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_r
         if (privateExponentArr) {
             BigNumObj privExp = BigNumObj::fromJavaArray(env, privateExponentArr);
 
-            int res;
+            int res = 1;
             if (BN_is_zero(pubExp)) {
-                // RSA blinding can't be performed without |e|; 0 indicates |e|'s absence.
-                rsa->flags |= RSA_FLAG_NO_BLINDING;
-                res = RSA_set0_key(rsa, modulus, NULL, privExp);
+                // RSA blinding can't be performed without |e|.
```
WillChilds-Klein, line 501: nit: suggest moving this comment to L182

csrc/keyutils.cpp
```diff
@@ -169,4 +169,36 @@ const EVP_MD* digestFromJstring(raii_env& env, jstring digestName)
...
+{
+    RSA_auto rsa = RSA_auto::from(RSA_new());
+    if (rsa.get() == nullptr) {
+        throw_openssl("RSA_new failed");
+    }
+    // The FIPS builds of AWS-LC do not have an API to disable blinding.
+#ifdef FIPS_BUILD
+
+    // RSA struct is not opaque in FIPS mode.
+    rsa->flags |= RSA_FLAG_NO_BLINDING;
```
WillChilds-Klein, line 183: why not also also set this flag in the non-FIPS case as we were doing before?

csrc/keyutils.cpp
```diff
@@ -169,4 +169,36 @@ const EVP_MD* digestFromJstring(raii_env& env, jstring digestName)
...
+
+#else
+
+    // RSA_blinding_off_temp_for_accp_compatibility is marked deprecated, so we need to disable the
+    // "deprecated-declarations" check for this invocation.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+    // We need to change this API once AWS-LC provides a method similar to the following:
+    // https://github.com/google/boringssl/blob/master/include/openssl/rsa.h#L630
+    RSA_blinding_off_temp_for_accp_compatibility(rsa);
```
WillChilds-Klein, line 193: should we file an Issue against AWS-LC for a stable API to accomplish this?

WillChilds-Klein (approved):

csrc/java_evp_keys.cpp
amirhosv: I'll fix that in the next PR.

csrc/keyutils.cpp
amirhosv: AWS-LC has this on their radar.

PR #375, opened 2024-04-03
Improve SecureRandom benchmarks (especially in multi-threaded settings)

fabrice102:
*Issue #, if available:* N/A

*Description of changes:*

1. Rename Drbg.java benchmark to Random.java and add more benchmarks (SUN/NativePrng, java.util.random, and some baseline using copy instead of randomness generation). SUN/NativePrng is the default Java implementation of SecureRandom when ACCP is not installed.
2. Remove configuration line in build.gradle.kts forcing single thread in all benchmarks. This is because command line arguments override annotations, and we need to use multiple threads in some benchmarks.
3. Use average measurement (ns/op) in build.gradle.kts instead of throughput (op/ns). This is because for multi-threaded benchmark of randomness generation, we want to know the duration of a single randomness generation on a single thread (and we want to check if performance drops because of multi-threading).
4. Make the data variable (the target of randomness generation)  thread-local to avoid L1 cache contention.
5. Add a few more benchmarks: 
    a. multiThreadedLocal where the SecureRandom instance is thread local, instead of global: this is useful for users of ACCP to know whether they should have SecureRandom be thread local or not. 
    b. singleThreadedNew creating a new instance of SecureRandom: this is useful for users of ACCP to know whether it is ok to instantiate a SecureRandom every time they need it, or to instead create a global instance. 
    c. A variant of the benchmarks generating 4 bytes of random data, instead of 1,024 bytes. For example, generating a random int requires 4 bytes of random data.


geedo0 (approved):
These are great and well thought out improvements, thanks!

benchmarks/lib/build.gradle.kts
```diff
@@ -47,9 +47,11 @@ jmh {
         includes.add(includeBenchmark)
     }
     fork.set(1)
-    benchmarkMode.add("thrpt")
```
amirhosv, line 50: Many of the benchmarks assume a single threaded execution. It's best not to change this global setting and just modify it at a benchmark level.

amirhosv (changes requested):

benchmarks/lib/build.gradle.kts
fabrice102: Fixed by new commit: I've removed the timeUnit and benchmarkMode completely so they default to ops/s. And I've added annotations in the Random benchmarks to force ns/op there.

fabrice102:
I've rebased to split the commits in two:

1. rename `Drbg.java` into `Random.java`
2. apply other changes

This is to facilitate the comparison with diff in Github (using the diff between the second commit and the first commit).

geedo0 (approved):

amirhosv (approved):

PR #376, opened 2024-04-03
Workaround to mark the DEFAULT SecureRandom algorithm thread-safe

fabrice102:
*Issue #, if available:* N/A

*Description of changes:*

This is a workaround to ensure that the `nextBytes` calls to SecureRandom are not synchronized when specifying the alias algorithm `DEFAULT` of ACCP SecureRandom provider. Without this workaround, the `nextBytes` calls are synchronized and the multi-threaded performance of ACCP SecureRandom drops: concretely, with x threads concurrently generating randomness, average time per thread gets multiplied by x.

In more details, OpenJDK/Corretto requires thread safe implementations of SecureRandom to specify the attribute ThreadSafe (https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/security/SecureRandom.html). If this attribute is not set to true, the nextBytes calls are synchronized. ACCP sets this attribute when registering SecureRandom. It also registers an alias `DEFAULT` to the algorithm `LibCryptoRng` that it registers.

Attributes of services are converted to attributes at the provider level, using the naming convention: `Service.Algorithm Attribute`. Thus, setting the `ThreadSafe` attribute leads to setting `SecureRandom.LibCryptoRng ThreadSafe=true`, but not `SecureRandom.DEFAULT ThreadSafe=true`.

Unfortunately, the JDK does not follow aliases and thus the DEFAULT algorithm is not considered thread safe. See, e.g., https://github.com/corretto/corretto-17/blob/e0d038c50d5c61683a9270218b32c5aed2b0b466/src/java.base/share/classes/java/security/SecureRandom.java#L229-L236

FAQ:

> Who is affected by the performance drop if this workaround is not applied?

Anyone instantiating explicitly the `DEFAULT` algorithm of ACCP SecureRandom (i.e., `SecureRandom.getInstance("DEFAULT", "AmazonCorrettoCryptoProvider")`) and using it concurrently in multiple threads.

This includes the benchmarks in the subfolder `benchmarks/` of this repo.

> I am using `new SecureRandom`, am I affected?

No. `new SecureRandom()` uses the algorithm `LibCryptoRng` and not the alias `DEFAULT` (assuming ACCP is set as first cryptographic provider).

> I did not register ACCP SecureRandom, am I affected?

No.

> Can this workaround impact other crypto providers like BouncyCastle (which also provides a `DEFAULT` algorithm for SecureRandom)?

No, attributes are stored at the provider level. This change will not affect other providers.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
+++ b/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
@@ -121,6 +121,9 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
               singletonMap("ThreadSafe", "true"),
               "DEFAULT")
           .setSelfTest(LibCryptoRng.SPI.SELF_TEST);
+      // Following line is a workaround to ensure that the SecureRandom service
+      // is seens as ThreadSafe by SecureRandom, when using the alias name DEFAULT
+      setProperty("SecureRandom.DEFAULT ThreadSafe", "true");
     }
 
     addSignatures();
--- a/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
+++ b/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
@@ -121,9 +121,24 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
               singletonMap("ThreadSafe", "true"),
               "DEFAULT")
           .setSelfTest(LibCryptoRng.SPI.SELF_TEST);
-      // Following line is a workaround to ensure that the SecureRandom service
-      // is seens as ThreadSafe by SecureRandom, when using the alias name DEFAULT
-      setProperty("SecureRandom.DEFAULT ThreadSafe", "true");
+
+      // Following lines are a workaround to ensure that the SecureRandom service
+      // is seen as ThreadSafe by SecureRandom, when using the alias name DEFAULT
+      // See https://bugs.openjdk.org/browse/JDK-8329754
+
+      // We add additional tests to confirm the DEFAULT algorithm is actually ThreadSafe
+      // This is to prevent issues in case a future code change set DEFAULT to a non-ThreadSafe
+      // algorithm
+
+      // Get the name of the algorithm pointed by the alias name DEFAULT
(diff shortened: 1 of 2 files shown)


src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -121,6 +121,9 @@ private void buildServiceMap() {
               singletonMap("ThreadSafe", "true"),
               "DEFAULT")
           .setSelfTest(LibCryptoRng.SPI.SELF_TEST);
+      // Following line is a workaround to ensure that the SecureRandom service
+      // is seens as ThreadSafe by SecureRandom, when using the alias name DEFAULT
```
geedo0, line 125: small typo: "seen as ThreadSafe"

geedo0 (approved):

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
nebeid, line 125: Is there a way to check that DEFAULT is still LibCryptoRng before setting ThreadSafe to true? Could there be a scenario where the DEFAULT becomes another provider that's not ThreadSafe and the following line is forgotten?
If there isn't a programmatic check, then maybe add a warning to the documentation, e.g.
"Warning: The following line assumes that DEFAULT is an alias of LibCryptoRng. Change it if the DEFAULT changes."

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -121,6 +121,9 @@ private void buildServiceMap() {
...
+      setProperty("SecureRandom.DEFAULT ThreadSafe", "true");
```
amirhosv, line 126: Could you add a test that verifies both "DEFAUTL" and "LibCryptoRng" have "ThreadSafe" property set to true?

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
fabrice102: Good point. I've added a check that DEFAULT points to an algorithm that is itself marked ThreadSafe.

geedo0 (approved):

amirhosv (approved):

PR #378, opened 2024-04-12
Bundling script to add ACCP to JDK

amirhosv:
*Description of changes:*

Two scripts that allow one to add ACCP to their JDKs. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,26 @@ could be used so that one does not have to explicitly specify the platform.
 [Here](https://github.com/corretto/amazon-corretto-crypto-provider/blob/f1d54b34cf4765789314941dbeefdafd35a4da58/examples/gradle-kt-dsl/lib/build.gradle.kts#L30)
 is an example.
 
+### Bundle ACCP with JDK
+We provide two scripts that allow one to add ACCP to their JDKs: one for JDK8 and one for JDKs 11+.
+Please note that these scripts are provided as examples and for testing only.
+
+These scripts take the version of ACCP and the classifier as input. Optionally, one can pass `-FIPS`
+as the third argument to bundle the FIPS artifacts. To use these scripts, please set `JAVA_HOME` to
+the path of your desired JDK.
+
+Usage example:
+```bash
+./bin/bundle-accp.sh 2.3.3 linux-x86_64
+```
+
+To find the the available versions and classifiers, please checkout Maven central.
+
+Some notes on the bundling scripts:
+* One needs to run the bundling script only once.
+* The bundling is not idempotent: runing the script on a JDK that has ACCP bundled in it could result in undefined behavior.
+* There is no unbundling. Please do a fresh install of the JDK if you need to remove ACCP from your JDK.
+
 ### Manual
 Manual installation requires acquiring the provider and adding it to your classpath.
 You can either download a prebuilt version of the provider or build it yourself.
--- a/bin/bundle-accp-jdk8.sh
+++ b/bin/bundle-accp-jdk8.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+set -e
+set -x
+
+# Use this script for JDKs 8. Please note that this script does not modify the security file,
+# so one still needs to programmatically install ACCP.
+#
+# This script requires JAVA_HOME set properly.
+#
+# usage: ./bundle-accp-jdk8.sh 2.3.3 linux-x86_64
+
+ACCP_VERSION=$1 # for example 2.3.3
+ACCP_PLATFORM=$2 # for example linux-x86_64 or osx-aarch64
+FIPS_EXTENSION=${3:""} # for FIPS, pass -FIPS
+BUILD_DIR="$(pwd)/build"
+TMP_DIR="$BUILD_DIR/private/tmp"
+ACCP_PATH="$TMP_DIR/accp"
+JDK_PATH="$JAVA_HOME"
+JAVA_BIN="$JDK_PATH/bin"
+
+function download_accp {
+    wget -O "$ACCP_PATH/accp.jar" "https://repo1.maven.org/maven2/software/amazon/cryptools/AmazonCorrettoCryptoProvider${FIPS_EXTENSION}/${ACCP_VERSION}/AmazonCorrettoCryptoProvider${FIPS_EXTENSION}-${ACCP_VERSION}-${ACCP_PLATFORM}.jar"
+}
+
+function mk_all_dirs {
+    clean_all_dirs
+    mkdir -p "$ACCP_PATH"
+}
+
+function clean_all_dirs {
+    rm -rf "$ACCP_PATH"
+}
+
+function repackage_jdk {
+    # Repackage JDK
+    cp "$ACCP_PATH/accp.jar" "$JDK_PATH/jre/lib/ext"
+    echo "Integration completed."
+}
+
+function test_integration {
+    pushd examples/gradle-kt-dsl > /dev/null
+    ./gradlew -PuseBundledAccp -Dorg.gradle.java.home="$JDK_PATH" lib:test
+    popd > /dev/null
+}
+
+function main {
+    mk_all_dirs
+    download_accp
+    repackage_jdk
+    clean_all_dirs
+    test_integration
+}
+
+main
--- a/bin/bundle-accp.sh
+++ b/bin/bundle-accp.sh
@@ -1,0 +1,81 @@
+#!/bin/bash
+set -e
+set -x
+
+# Use this script for JDKs 11+. For JDK 8, bundling requires copying the jar into lib/ext directory of JDK.
+# Please note that this script does not modify the security file, so one still needs to programmatically
+# install ACCP.
+#
+# This script requires JAVA_HOME set properly.
+#
+# usage: ./bundle-accp.sh 2.3.3 linux-x86_64
+
+ACCP_VERSION=$1 # for example 2.3.3
+ACCP_PLATFORM=$2 # for example linux-x86_64 or osx-aarch64
+FIPS_EXTENSION=${3:""} # for FIPS, pass -FIPS
+BUILD_DIR="$(pwd)/build"
+TMP_DIR="$BUILD_DIR/private/tmp"
+ACCP_PATH="$TMP_DIR/accp"
+ACCP_JMOD_DIR="$TMP_DIR/accp_jmod"
+CUSTOM_JDK_PATH="$TMP_DIR/customjdk"
+JDK_PATH="$JAVA_HOME"
+JAVA_BIN="$JDK_PATH/bin"
+
+function download_accp {
+    wget -O "$ACCP_PATH/accp.jar" "https://repo1.maven.org/maven2/software/amazon/cryptools/AmazonCorrettoCryptoProvider${FIPS_EXTENSION}/${ACCP_VERSION}/AmazonCorrettoCryptoProvider${FIPS_EXTENSION}-${ACCP_VERSION}-${ACCP_PLATFORM}.jar"
+}
+
+function mk_all_dirs {
+    clean_all_dirs
+    mkdir -p "$ACCP_PATH"
+    mkdir -p "$ACCP_JMOD_DIR"
+}
+
+function clean_all_dirs {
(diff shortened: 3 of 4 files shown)


bin/bundle-accp-jdk8.sh
```diff
@@ -0,0 +1,48 @@
...
+ACCP_PLATFORM=$2 # for example linux-x86_64 or osx-aarch64
+FIPS_EXTENTION=${3:""} # for FIPS, pass -FIPS
+BUILD_DIR="$(pwd)/build"
+TMP_DIR=$BUILD_DIR/private/tmp
+ACCP_PATH=$TMP_DIR/accp
+JDK_PATH=$JAVA_HOME
+JAVA_BIN=$JDK_PATH/bin
+
+function download_accp {
+    wget -O $ACCP_PATH/accp.jar https://repo1.maven.org/maven2/software/amazon/cryptools/AmazonCorrettoCryptoProvider${FIPS_EXTENTION}/${ACCP_VERSION}/AmazonCorrettoCryptoProvider${FIPS_EXTENTION}-${ACCP_VERSION}-${ACCP_PLATFORM}.jar
```
fabrice102, line 22: Here and everywhere where a path is constructed from variables like `$ACCP_PATH`, I would add double quotes.

Concretely:

```
wget -O "$ACCP_PATH/accp.jar" https://repo1.maven.org/maven2/software/amazon/cryptools/AmazonCorrettoCryptoProvider${FIPS_EXTENTION}/${ACCP_VERSION}/AmazonCorrettoCryptoProvider${FIPS_EXTENTION}-${ACCP_VERSION}-${ACCP_PLATFORM}.jar
```

or below

```
./gradlew -PuseBundledAccp -Dorg.gradle.java.home="$JDK_PATH" lib:test
```

Without double quotes, the script will fail when the current path contains spaces.

In the assignments `TMP_DIR=$BUILD_DIR/private/tmp`, quotes are not mandatory, but I always prefer to put them there too, to be on the safe side.

See https://unix.stackexchange.com/a/68748

examples/gradle-kt-dsl/lib/build.gradle.kts
```diff
@@ -37,10 +37,12 @@ dependencies {
         else
             "AmazonCorrettoCryptoProvider"
 
-    if (project.hasProperty("accpLocalJar")) {
-        testImplementation(files(accpLocalJar))
-    } else {
-        testImplementation("software.amazon.cryptools:${accpArtifactId}:$accpVersion:${osdetector.classifier}")
+    if (!project.hasProperty("useBundledAccp")) {
```
WillChilds-Klein, line 40: does this allow the tests to still run against the bundled JAR?

bin/bundle-accp-jdk8.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Use this script for JDKs 8. Please note that this script does not modify the security file,
+# so one still needs to programmatically install ACCP.
+#
+# This script requires JAVA_HOME set properly.
+#
+# usage: ./bundle-script 2.3.3 linux-x86_64
+
+ACCP_VERSION=$1 # for example 2.3.3
+ACCP_PLATFORM=$2 # for example linux-x86_64 or osx-aarch64
```
WillChilds-Klein, line 13: nit: should we fail out with an error if nothing here is specified or default to something reasonable (perhaps grepping for latest version and using `uname` for `$1` and `$2` respectively)?

bin/bundle-accp-jdk8.sh
```diff
@@ -0,0 +1,48 @@
...
+    wget -O $ACCP_PATH/accp.jar https://repo1.maven.org/maven2/software/amazon/cryptools/AmazonCorrettoCryptoProvider${FIPS_EXTENTION}/${ACCP_VERSION}/AmazonCorrettoCryptoProvider${FIPS_EXTENTION}-${ACCP_VERSION}-${ACCP_PLATFORM}.jar
+}
+
+function mk_all_dirs {
+    mkdir -p $ACCP_PATH
+}
+
+function repackage_jdk {
+    # Repackage JDK
+    cp $ACCP_PATH/accp.jar $JDK_PATH/jre/lib/ext
```
WillChilds-Klein, line 31: is the `jre/lib/ext` path the same across different operating systems?

bin/bundle-accp.sh
```diff
@@ -0,0 +1,73 @@
...
+    mv com.amazon.corretto.crypto.provider.jmod $JDK_PATH/jmods/
+}
+
+function repackage_jdk {
+    # Repackage JDK
+    JMOD_LIST=$(ls -1 $JDK_PATH/jmods | sed 's/.jmod$//g' |  tr '\n' ',' | sed 's/,$//')
+    $JAVA_BIN/jlink -p $JDK_PATH/jmods --add-modules $JMOD_LIST --output $CUSTOM_JDK_PATH
+
+    # Overwrite files on original JDK
+    cp -rf $CUSTOM_JDK_PATH/* $JDK_PATH/
```
WillChilds-Klein, line 55: should we clean up the tmp JDK?

bin/bundle-accp.sh
```diff
@@ -0,0 +1,73 @@
...
+#
+# This script requires JAVA_HOME set properly.
+#
+# usage: ./bundle-script 2.3.3 linux-x86_64
+
+ACCP_VERSION=$1 # for example 2.3.3
+ACCP_PLATFORM=$2 # for example linux-x86_64 or osx-aarch64
+FIPS_EXTENTION=${3:""} # for FIPS, pass -FIPS
+BUILD_DIR="$(pwd)/build"
+TMP_DIR=$BUILD_DIR/private/tmp
```
WillChilds-Klein, line 17: `$(mktemp -d)` might be more portable

bin/bundle-accp-jdk8.sh
WillChilds-Klein: if these scripts are mac-specific, maybe we can error out if they're run on another platform?

bin/bundle-accp-jdk8.sh
```diff
@@ -0,0 +1,48 @@
...
+# Use this script for JDKs 8. Please note that this script does not modify the security file,
+# so one still needs to programmatically install ACCP.
+#
+# This script requires JAVA_HOME set properly.
+#
+# usage: ./bundle-script 2.3.3 linux-x86_64
+
+ACCP_VERSION=$1 # for example 2.3.3
+ACCP_PLATFORM=$2 # for example linux-x86_64 or osx-aarch64
+FIPS_EXTENTION=${3:""} # for FIPS, pass -FIPS
```
fabrice102, line 14:
```suggestion
FIPS_EXTENSION=${3:""} # for FIPS, pass -FIPS
```

fabrice102: Nit: typo "extention", same issue below

bin/bundle-accp-jdk8.sh
fabrice102: Depending on the audience, I agree that more input sanitizing may be helpful to the user, including:

* Check that the third argument is `-FIPS` or not, and error otherwise. This is because people can easily switch order of argument and this would result in hard to understand errors.
* Check number of arguments is 2 or 3 and display usage if not valid.
* Maybe check that ACCP_PLATFORM is in a list? Another option is to have an error message when downloading fails asking to check the version/platform/FIPS.
* Check that JAVA_HOME is not empty and is a folder.

examples/gradle-kt-dsl/lib/build.gradle.kts
amirhosv: Passing `-PuseBundledAccp` will cause the project to use ACCP that is bundled. Not specifying this option, it will either use what's on Maven or the JAR that has been built locally.

bin/bundle-accp-jdk8.sh
amirhosv: If any of these options is not set properly, the URL for download wouldn't be valid and download would fail. These scripts are provided as examples.

bin/bundle-accp-jdk8.sh
amirhosv: I tested the scripts for Mac and Linux. I guess once we support Windows, then we can revisit this issue.

bin/bundle-accp.sh
amirhosv: Sure. The bundling happens in the build directory and that's why I didn't include a "clean".

WillChilds-Klein: ah, if this is all in `./build`, then yeah explicit cleanup isn't really important

WillChilds-Klein (approved):

bin/bundle-accp-jdk8.sh
```diff
@@ -0,0 +1,53 @@
...
+ACCP_PATH="$TMP_DIR/accp"
+JDK_PATH="$JAVA_HOME"
+JAVA_BIN="$JDK_PATH/bin"
+
+function download_accp {
+    wget -O "$ACCP_PATH/accp.jar" "https://repo1.maven.org/maven2/software/amazon/cryptools/AmazonCorrettoCryptoProvider${FIPS_EXTENSION}/${ACCP_VERSION}/AmazonCorrettoCryptoProvider${FIPS_EXTENSION}-${ACCP_VERSION}-${ACCP_PLATFORM}.jar"
+}
+
+function mk_all_dirs {
+    mkdir -p "$ACCP_PATH"
```
alexw91, line 26: In the JDK 11+ script you run `clean_all_dirs` inside `mk_all_dirs`, did you want to do the same for this JDK8 script?

bin/bundle-accp.sh
```diff
@@ -0,0 +1,81 @@
...
+    echo "Integration completed."
+}
+
+function test_integration {
+    pushd examples/gradle-kt-dsl > /dev/null
+    ./gradlew -PuseBundledAccp -Dorg.gradle.java.home="$JDK_PATH" lib:test
+    popd > /dev/null
+}
+
+function main {
```
geedo0, line 72: I strongly believe that we should include unbundling functionality in this script to restore the user's JDK installation to how it was since this impacts the system JDK installation. It's not immediately apparent to me how I'd go about undo'ing this script. It's also not clear how updates work. Will running this script with newer versions of ACCP add additional ACCP's to my JDK installation or will they replace eachother?

README.md
```diff
@@ -176,6 +176,21 @@ could be used so that one does not have to explicitly specify the platform.
...
+We provide two scripts that allow one to add ACCP to their JDKs: one for JDK8 and one for JDKs 11+.
+Please note that these scripts are provided as examples and for testing only.
+
+These scripts take the version of ACCP and the classifier as input. Optionally, one can pass `-FIPS`
+as the third argument to bundle the FIPS artifacts. To use these scripts, please set `JAVA_HOME` to
+the path of your desired JDK.
+
+Usage example:
+```bash
+./bin/bundle-accp 2.3.3 linux-x86_64
```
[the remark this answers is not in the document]
amirhosv: done.

bin/bundle-accp-jdk8.sh
amirhosv: Thanks for catching this. Done.

README.md
```diff
@@ -176,6 +176,21 @@ could be used so that one does not have to explicitly specify the platform.
...
+```
+
+To find the the available versions and classifiers, please checkout Maven central.
```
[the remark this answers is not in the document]
amirhosv: There are many mirrors and they keep changing.

bin/bundle-accp.sh
geedo0: In that case, please document that //somewhere// so we know how to get back to a known good state. Also, is that also the case for updating the bundled ACCP? Or can the bundling script be called with a newer ACCP version and the bundle will get updated? Either way, please make it clear how updates/uninstallation is handled.

WillChilds-Klein (dismissed):

bin/bundle-accp.sh
amirhosv: I've documented this in the README.md.

geedo0 (approved):

WillChilds-Klein (approved):

PR #380, opened 2024-04-30
AES-CBC

amirhosv:
*Description of changes:*

Adding support for AES-CBC with no padding and with PKCS7 padding.


src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
```diff
@@ -0,0 +1,685 @@
...
+
+  private static int checkOperation(final int opMode) throws InvalidParameterException {
+    return ((opMode == Cipher.ENCRYPT_MODE) || (opMode == Cipher.WRAP_MODE)) ? ENC_MODE : DEC_MODE;
+  }
+
+  private static byte[] checkAesCbcIv(final AlgorithmParameterSpec params)
+      throws InvalidAlgorithmParameterException {
+    if (!(params instanceof IvParameterSpec)) {
+      throw new InvalidAlgorithmParameterException(
+          "I don't know how to handle a " + params.getClass());
```
geedo0, line 212: Let's not give ACCP a first person voice.

geedo0:
I did a first pass, but still need to go through how the buffering works and how we handle the padding/non-padding cases.

I normally don't like to police commit messages and PR descriptions, but this PR could use some more context and explanation around what's going on. This helps with the review process and also helps future developers figure out what happened in the code. This is especially important for large diffs like this. Things that would be helpful to explain:
- Everything that's going on with the buffers.
- Why PKCS5 aliases PKCS7.
- Why we're adding this algorithm or a reference to a ticket.
- What testing you're doing and why.
- The refactoring you did around AES GCM and this new CBC class.

The README.md file needs to be updated to include the new algorithm.

Curious on your thoughts why we pulled in the NIST test vectors for this algorithm. ACCP doesn't consistently pull-in NIST vectors and tests for all algorithms. This PR accounts for half the rsp.gz files that we've imported. It's unclear to me what decision criteria is used for when we do vector testing in ACCP in addition to AWS-LC's extensive vector testing.

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
```diff
@@ -0,0 +1,685 @@
...
+    // There is no need to check if the Cipher is initialized since
+    // javax.crypto.Cipher::getOutputSize checks that.
+    final long all = inputLen + unprocessedInput;
+
+    final long rem = all % BLOCK_SIZE_IN_BYTES;
+
+    // When there is no padding, the output size for enc/dec is at most all.
+    if (padding == NO_PADDING) {
+      return (int) (all);
+    }
```
geedo0, line 124: I don't think this is the case. It should still be up to the last complete block with any remainder left in unprocessed input.

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
```diff
@@ -0,0 +1,685 @@
...
+      throw new ShortBufferException(outputLen + "<" + getOutputSizeFinal(inputLen));
+    }
+  }
+
+  private int getOutputSizeFinal(final int inputLen) throws IllegalBlockSizeException {
+    final long all = ((long) inputLen) + ((long) unprocessedInput);
+    final long rem = all % BLOCK_SIZE_IN_BYTES;
+    // If there is no padding or if we are decrypting, all the data must be aligned with block size.
+    if ((opMode == DEC_MODE || padding == NO_PADDING) && rem != 0) {
+      throw new IllegalBlockSizeException();
```
geedo0, line 471: Should we add an actionable message here for the users? Just informing them that padding is disabled and that doFinal was invoked with n-bytes remaining instead of the block size.

This is also a bit of an unexpected place for the boundary check. Suggest moving it somewhere more apparent.

tst/com/amazon/corretto/crypto/provider/test/TestUtil.java
```diff
@@ -448,4 +461,284 @@ public static synchronized void restoreProviders(final Provider[] providers) {
...
+    }
+    final int len = a.remaining();
+    for (int i = 0; i != len; i++) {
+      if (a.get(a.position() + i) != b.get(b.position() + i)) {
+        return false;
+      }
+    }
+    return true;
+
+     */
```
geedo0, line 521: Don't check in commented code

tst/com/amazon/corretto/crypto/provider/test/AesCbcTest.java
```diff
@@ -0,0 +1,894 @@
...
+import org.junit.jupiter.api.parallel.ResourceAccessMode;
+import org.junit.jupiter.api.parallel.ResourceLock;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+@ExtendWith(TestResultLogger.class)
+@Execution(ExecutionMode.SAME_THREAD)
+@ResourceLock(value = TestUtil.RESOURCE_GLOBAL, mode = ResourceAccessMode.READ)
+public class AesCbcTest {
```
geedo0, line 58: Nice testing in here, really thorough!

tst/com/amazon/corretto/crypto/provider/test/AesCbcNistTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.jupiter.api.parallel.ExecutionMode;
+import org.junit.jupiter.api.parallel.ResourceAccessMode;
+import org.junit.jupiter.api.parallel.ResourceLock;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+@ExtendWith(TestResultLogger.class)
+@Execution(ExecutionMode.CONCURRENT)
+@ResourceLock(value = TestUtil.RESOURCE_GLOBAL, mode = ResourceAccessMode.READ)
+public class AesCbcNistTest {
```
geedo0, line 25: I'm unsure about whether or not we should be including NIST test vectors in ACCP. This is not something we do for all algorithms in ACCP. More testing is always great, but I think it's a bit redundant with AWS-LC's testing. Our focus should be more on tests that surface issues around interfacing with the native layer and less so around exhaustive validation of the cryptographic implementation.

tst/com/amazon/corretto/crypto/provider/test/AesCbcTest.java
```diff
@@ -0,0 +1,894 @@
...
+    final ByteBuffer inputBuffer = ByteBuffer.allocate(input.length);
+    inputBuffer.put(input).flip();
+    final IvParameterSpec iv =
+        new IvParameterSpec(TestUtil.decodeHex("000102030405060708090A0B0C0D0E0F"));
+    final Cipher cbcPadding = accpAesCbcCipher(true);
+    cbcPadding.init(Cipher.ENCRYPT_MODE, key, iv);
+    // final ByteBuffer cipherText = ByteBuffer.allocate(input.length + 16);
+    // final int cipherLen = cbcPadding.doFinal(directInput, cipherText);
+    // cipherText.flip();
+    // assertEquals(cipherLen, cipherText.remaining());
```
alexw91, line 492: Should this code be uncommented/deleted?

.github/workflows/macos_ci_jobs.yml
```diff
@@ -22,9 +22,9 @@ jobs:
     - uses: actions/checkout@v3
       with:
         submodules: true
-    - name: Install lcov
+    - name: Install lcov golang
```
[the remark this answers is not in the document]
amirhosv: Done.

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
```diff
@@ -0,0 +1,685 @@
...
+    AES_CBC_NO_PADDING_NAMES.add("AES_128/CBC/NoPadding".toLowerCase());
+    AES_CBC_NO_PADDING_NAMES.add("AES_192/CBC/NoPadding".toLowerCase());
+    AES_CBC_NO_PADDING_NAMES.add("AES_256/CBC/NoPadding".toLowerCase());
+
+    AES_CBC_PKCS7_PADDING_NAMES = new HashSet<>();
+    AES_CBC_PKCS7_PADDING_NAMES.add("AES/CBC/PKCS7Padding".toLowerCase());
+    AES_CBC_PKCS7_PADDING_NAMES.add("AES_128/CBC/PKCS7Padding".toLowerCase());
+    AES_CBC_PKCS7_PADDING_NAMES.add("AES_192/CBC/PKCS7Padding".toLowerCase());
+    AES_CBC_PKCS7_PADDING_NAMES.add("AES_256/CBC/PKCS7Padding".toLowerCase());
+    // PKCS5Padding with AES/CBC should be treated as PKCS7Padding
```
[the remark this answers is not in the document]
amirhosv: Updated the comment and added a test case that shows BC recognizes PKCS7Padding and treats it the same as PKCS5Padding; however, SunJCE only recognizes PKCS5Padding name.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -309,6 +326,22 @@ public Object newInstance(final Object constructorParameter) throws NoSuchAlgori
           return new AesXtsSpi();
         }
 
+        if ("Cipher".equalsIgnoreCase(type)
+            && AES_CBC_PKCS7_PADDING_NAMES.contains(algo.toLowerCase())) {
+          final boolean saveContext =
+              AmazonCorrettoCryptoProvider.this.nativeContextReleaseStrategy
```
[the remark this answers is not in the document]
amirhosv: This is happening inside ACCPService class, which is a non-static nested class inside AmazonCorrettoCryptoProvider.

src/com/amazon/corretto/crypto/provider/ShimByteBuffer.java
```diff
@@ -0,0 +1,51 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider;
+
+import java.nio.ByteBuffer;
+
+// TODO: merge this class and AesGcmSpi.ShimArray
+class ShimByteBuffer {
```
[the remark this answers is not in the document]
amirhosv: Added comments to this class. The comments capture your points.

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
amirhosv: Sure.

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
amirhosv: You are right, but this method cannot assume if the next call is going to be update or doFinal. It also should not check alignment since that check only happens during doFinal.

tst/com/amazon/corretto/crypto/provider/test/TestUtil.java
amirhosv: My bad. Forgot to remove this.

tst/com/amazon/corretto/crypto/provider/test/AesCbcTest.java
amirhosv: My bad. I forgot to remove the commented code.

amirhosv:
> I did a first pass, but still need to go through how the buffering works and how we handle the padding/non-padding cases.
> 
> I normally don't like to police commit messages and PR descriptions, but this PR could use some more context and explanation around what's going on. This helps with the review process and also helps future developers figure out what happened in the code. This is especially important for large diffs like this. Things that would be helpful to explain:
> 
> * Everything that's going on with the buffers.
> * Why PKCS5 aliases PKCS7.
> * Why we're adding this algorithm or a reference to a ticket.
> * What testing you're doing and why.
> * The refactoring you did around AES GCM and this new CBC class.
> 
> The README.md file needs to be updated to include the new algorithm.
> 
> Curious on your thoughts why we pulled in the NIST test vectors for this algorithm. ACCP doesn't consistently pull-in NIST vectors and tests for all algorithms. This PR accounts for half the rsp.gz files that we've imported. It's unclear to me what decision criteria is used for when we do vector testing in ACCP in addition to AWS-LC's extensive vector testing.

I think I've answered all your comments.

tst/com/amazon/corretto/crypto/provider/test/AesCbcNistTest.java
geedo0: I understand that we have them for other algorithms, but not for //all// algorithms. I'm trying to understand what our decision criteria is for including or not including KATs. It seems pretty random. Why are we exhaustively testing every single KAT style/variant for CBC but not for GCM? It just seems off to me for us to be KAT testing AES-CBC much more thoroughly than literally every other algorithm in ACCP.

geedo0:
> I think I've answered all your comments.

Yes, I see it all in the code. To clarify though, I was really hoping to see some of these design decisions documented in the PR itself. That helps me a lot when trying to reason out why certain code looks the way it does.

geedo0 (dismissed):

WillChilds-Klein (commented):
still reviewing, about 1/2 way through.

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
```diff
@@ -0,0 +1,704 @@
...
+                        outputArray,
+                        outputOffset));
+        // No need to update the cipherState since it's already in UPDATED state.
+      }
+      final long all = inputLen + unprocessedInput;
+      unprocessedInput = (int) (all - result);
+      return result;
+    } catch (final Exception e) {
+      cipherState = CipherState.NEEDS_INITIALIZATION;
+      saveNativeContextIfNeeded(ctxContainer[0]);
```
WillChilds-Klein, line 386: if we're resetting the cipher state, why save the native context? shouldn't we reset all state?

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
```diff
@@ -0,0 +1,704 @@
...
+      }
+
+      cipherState = CipherState.INITIALIZED;
+      unprocessedInput = 0;
+
+      return result;
+
+    } catch (final Exception e) {
+      cipherState = CipherState.NEEDS_INITIALIZATION;
+      saveNativeContextIfNeeded(ctxContainer[0]);
```
WillChilds-Klein, line 600: same question as earlier about resetting ctx on state reset

tst/com/amazon/corretto/crypto/provider/test/TestUtil.java
```diff
@@ -448,4 +461,330 @@ public static synchronized void restoreProviders(final Provider[] providers) {
...
+      if (chunk == null) continue;
+      result.put(chunk);
+    }
+
+    result.flip();
+
+    return result;
+  }
+
+  public static List<Integer> fixPattern(final int inputLen, final int c) {
```
WillChilds-Klein, line 559: nit: `fixedPattern` or `constantPattern`. "fix pattern" sounds like an action more than a description.

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
amirhosv: Here, we should either clean up the EVP_CIPHER_CTX by making a JNI call, or register it with the Janitor for clean up. I'm open to changing the behavior to clean up.

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
```diff
@@ -0,0 +1,704 @@
...
+    // All checks passes, so we update the state:
+    this.cipherState = CipherState.INITIALIZED;
+    this.opMode = opMode;
+    this.iv = iv;
+    this.key = keyBytes;
+    this.unprocessedInput = 0;
+  }
+
+  private static int checkOperation(final int opMode) throws InvalidParameterException {
+    return ((opMode == Cipher.ENCRYPT_MODE) || (opMode == Cipher.WRAP_MODE)) ? ENC_MODE : DEC_MODE;
```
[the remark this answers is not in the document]
amirhosv: The Cipher class does these checks and I've added a unit test `whenInitializingWithInvalidMode_expectException` to confirm this. However, the `PUBLIC_KEY` is equal to `1`, which is the same as `ENCRYPT_MODE`.

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
```diff
@@ -0,0 +1,704 @@
...
+    AES_CBC_PKCS7_PADDING_NAMES.add("AES_256/CBC/PKCS5Padding".toLowerCase());
+  }
+
+  private static final byte[] EMPTY_ARRAY = new byte[0];
+  private static final int NO_PADDING = 0;
+  private static final int PKCS7_PADDING = 1;
+  private static final int BLOCK_SIZE_IN_BYTES = 128 / 8;
+  private static final int MODE_NOT_SET = -1;
+  private static final int ENC_MODE = 1;
+  private static final int DEC_MODE = 0;
```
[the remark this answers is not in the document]
amirhosv: The values come from AWS-LC. I added comments.

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
```diff
@@ -0,0 +1,704 @@
...
+  @Override
+  protected void engineSetMode(final String mode) throws NoSuchAlgorithmException {
+    // no op. One only needs to provide an implementation if the same Spi class instance can be used
+    // for different modes.
+  }
+
+  @Override
+  protected void engineSetPadding(final String padding) throws NoSuchPaddingException {
+    // no op. One only needs to provide an implementation if the same Spi class instance is used for
+    // different paddings.
```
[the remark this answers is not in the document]
amirhosv: We do, but this method is not exposed to the user. The Cipher class uses it and since we only instantiate the Spi with the correct padding, the check here would be redundant. There are providers that don't set the padding in their Spi classes when it is instantiated. For those, it would be a good idea to override this method.

csrc/buffer.cpp
```diff
@@ -49,4 +50,53 @@ JByteArrayCritical::~JByteArrayCritical() { env_->ReleasePrimitiveArrayCritical(
 
 unsigned char* JByteArrayCritical::get() { return (unsigned char*)ptr_; }
 
+SimpleBuffer::SimpleBuffer(int size)
+    : buffer_(nullptr)
+{
+    buffer_ = (uint8_t*)malloc(size);
```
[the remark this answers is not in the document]
amirhosv: I think we only have a SecureBuffer that allocates on stack. The scratch buffer size used when output clobbers input depends on the input/output buffers at runtime, so don't know the size before hand.
I don't think there is any issue in using malloc/free. The only concern could be that we do not zero memory after deallocation. The buffer is used to hold cipherText/plainText. In that sense it's as secure as how Java treats buffers by the garbage collector.
I'm open to suggestion to see if zeroing memory is needed before freeing up.

csrc/buffer.h
```diff
@@ -603,5 +603,30 @@ class JByteArrayCritical {
     jbyteArray jarray_;
 };
```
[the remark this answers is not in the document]
amirhosv: Sure.

csrc/aes_cbc.cpp
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        return result;
+    }
+
+    int do_final(uint8_t* output)
+    {
+        int result = 0;
+        if (EVP_CipherFinal_ex(ctx_, output, &result) != 1) {
+            if (ERR_get_error() == AWS_LC_BAD_PADDING_ERROR_CODE) {
```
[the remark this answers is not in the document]
amirhosv: Done.

csrc/aes_cbc.cpp
```diff
@@ -0,0 +1,308 @@
...
+#define KEY_LEN_AES256              32
+
+namespace AmazonCorrettoCryptoProvider {
+
+class AesCbcCipher {
+    JNIEnv* jenv_;
+    EVP_CIPHER_CTX* ctx_;
+    bool own_ctx_;
+
+    static bool output_clobbers_input(uint8_t const* input, int input_len, uint8_t const* output, int unprocessed_input)
```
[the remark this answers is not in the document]
amirhosv: There is a method called `check_unprocessed_input` that ensures unprocessed_input is in the range of [0, 16]. The reason for using `int` is that all the types used for representing length in Java are signed ints.

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
amirhosv: The more I think about it, doing the JNI call and cleaning it up is better. By registering it with Janitor, we will only save IF the same Cipher object is re used.

tst/com/amazon/corretto/crypto/provider/test/AesCbcNistTest.java
amirhosv: I guess we never formally specified when to add KATs. My rule of thumb is to search online for KATs and if found, then add.

csrc/buffer.cpp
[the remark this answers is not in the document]
amirhosv: Two things:
1. The code is updated so that before freeing the memory it is zeroed.
2. I see that we have [SecureAlloc](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/csrc/env.h#L261) that could potentially be used. The main issue with this class is that it throws `std::bad_alloc`. In our JNI code, we only catch `java_ex`, which means any other type of exception could potentially result in undefined behavior in JNI calls.

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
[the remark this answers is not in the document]
WillChilds-Klein: I see, makes sense.

csrc/buffer.h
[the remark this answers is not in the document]
WillChilds-Klein: sorry, i'd meant doc comments for the new ShimBuffer, but it's not super important.

csrc/buffer.cpp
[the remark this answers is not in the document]
WillChilds-Klein: Replacing SecureAlloc with `OPENSSL_cleanse` makes sense. I wonder if we should extend that idea a little further, using `OPENSSL_malloc` and `OPENSSL_free` in SimpleBuffer. This would provide two benefits:

1. Automatically [call](https://github.com/aws/aws-lc/blob/e3c74c00a5303757e122087dfb6e9722e5dcb246/crypto/mem.c#L246) `OPENSSL_cleanse` on free
2. Automatically use the same underlying allocator as AWS-LC in case they ever decide to switch to e.g. [jemalloc](https://jemalloc.net/jemalloc.3.html) or something. due to portability concerns, IMO this is unlikely.

not a blocker at all, just perhaps something to consider.

WillChilds-Klein (dismissed):
LGTM, just two small open comment threads to consider here or in future PRs.

csrc/buffer.h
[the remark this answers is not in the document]
amirhosv: No worries. I added comments to ShimByteBuffer as well.

csrc/buffer.cpp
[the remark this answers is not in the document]
amirhosv: That's great suggestion. I changed it so that it will use OPENSS_{malloc,free}.

WillChilds-Klein (approved):

PR #382, opened 2024-05-14
Make ACCP compatible with SunJCE when decrypting empty array with CBC with padding

amirhosv:
*Description of changes:*

For empty cipher text, SunJCE returns empty plain text when decrypting with padding enabled. This is despite the fact that cipher text with padding is always at least 16 bytes.  AWS-LC has a different behavior: `EVP_CipherFinal` fails when no input is passed during decryption with PKCS7Padding. For compatibility reasons, we follow SunJCE's approach.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
@@ -250,6 +250,7 @@ class AesCbcSpi extends CipherSpi {
     this.iv = iv;
     this.key = keyBytes;
     this.unprocessedInput = 0;
+    this.inputIsEmpty = true;
     initLastBlock();
   }
 
@@ -377,7 +378,9 @@ class AesCbcSpi extends CipherSpi {
       final byte[] outputArray,
       final int outputOffset) {
 
-    inputIsEmpty = inputIsEmpty && (inputLen == 0);
+    if (inputLen > 0) {
+      inputIsEmpty = false;
+    }
 
     // Unlike, doFinal (which needs to decide if a context should be released or not), update always
     // has to save the context.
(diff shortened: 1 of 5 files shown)


src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
```diff
@@ -372,6 +377,8 @@ private int update(
       final byte[] outputArray,
       final int outputOffset) {
 
+    inputIsEmpty = inputIsEmpty && (inputLen == 0);
```
geedo0, line 380: nit: I think it'd clearer if the logic looked like this. No need to reason about the state of inputIsEmpty between invocations.

```
if (inputLen > 0) {
  inputIsEmpty = false;
}
```

geedo0 (dismissed):

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
amirhosv: If I don't get any other comments, I'll have it done in the PR that is going to refactor the scratch pad length log, which I promised.

WillChilds-Klein: is the idea here to track `inputIsEmpty` state across **all** update/doFinal calls, and if _any_ of them have non-empty inputs, ensure we go through full processing logic in doFinal?

WillChilds-Klein: if so, do we also need to reset `inputIsEmpty` in [engineInit](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/AesCbcSpi.java#L232)?

WillChilds-Klein (approved):

geedo0 (approved):

PR #384, opened 2024-06-06
Address -Xcheck:jni warnings for AES/CBC

amirhosv:
*Description of changes:*

+ If one of the input/output buffers is a direct byte buffer and the other one is a byte[], there is a chance that we call GetDirectBufferAddress after calling GetPrimitiveArrayCritical. This triggers a warning and this PR addresses that.
+ JIOBlobs is introduced that handles both input and output buffers. This class helps with addressing the warnings caused by -Xcheck:jni and it also avoids calling GetPrimitiveArrayCritical twice if the same byte[] is used for input and output.
+ When decrypting in AES/CBC/PKCS7Padding mode, AWS-LC touches the output buffer beyond its returned length. This means in ACCP we need to over allocate when calling update in multi-step decryption.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -687,11 +687,9 @@ add_custom_target(check-with-jni-flag
     COMMAND ${TEST_JAVA_EXECUTABLE}
         -Xcheck:jni
         ${TEST_RUNNER_ARGUMENTS}
-        --select-class=com.amazon.corretto.crypto.provider.test.AesTest
-        --select-class=com.amazon.corretto.crypto.provider.test.AesGcmKatTest
-        --select-class=com.amazon.corretto.crypto.provider.test.AesCbcTest
-        --select-class=com.amazon.corretto.crypto.provider.test.AesCbcNistTest
-        --select-class=com.amazon.corretto.crypto.provider.test.AesCbcIso10126Test
+        --select-package=com.amazon.corretto.crypto.provider.test
+        --exclude-package=com.amazon.corretto.crypto.provider.test.integration
+        --exclude-classname=com.amazon.corretto.crypto.provider.test.SecurityManagerTest
 
     DEPENDS accp-jar tests-jar)
 
@@ -840,7 +838,7 @@ else()
         check-junit-AesLazy
         check-junit-AesEager
         check-junit-DifferentTempDir
-        # check-with-jni-flag for macos, we disable this check due to failures in java.net.Inet6AddressImpl.lookupAllHostAddr
+        # check-with-jni-flag for macos, we disable this check due to failures in java.net.Inet6AddressImpl.lookupAllHostAddr: https://bugs.openjdk.org/browse/JDK-8205076
     )
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -683,6 +683,18 @@ add_custom_target(check-junit-SecurityManager
 
     DEPENDS accp-jar tests-jar)
 
+add_custom_target(check-with-jni-flag
+    COMMAND ${TEST_JAVA_EXECUTABLE}
+        -Xcheck:jni
+        ${TEST_RUNNER_ARGUMENTS}
+        --select-class=com.amazon.corretto.crypto.provider.test.AesTest
+        --select-class=com.amazon.corretto.crypto.provider.test.AesGcmKatTest
+        --select-class=com.amazon.corretto.crypto.provider.test.AesCbcTest
+        --select-class=com.amazon.corretto.crypto.provider.test.AesCbcNistTest
+        --select-class=com.amazon.corretto.crypto.provider.test.AesCbcIso10126Test
+
+    DEPENDS accp-jar tests-jar)
+
 add_custom_target(check-junit-AesLazy
     COMMAND ${TEST_JAVA_EXECUTABLE}
         -Dcom.amazon.corretto.crypto.provider.nativeContextReleaseStrategy=LAZY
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -683,6 +683,18 @@ add_custom_target(check-junit-SecurityManager
...
 
+add_custom_target(check-with-jni-flag
+    COMMAND ${TEST_JAVA_EXECUTABLE}
+        -Xcheck:jni
+        ${TEST_RUNNER_ARGUMENTS}
+        --select-class=com.amazon.corretto.crypto.provider.test.AesTest
+        --select-class=com.amazon.corretto.crypto.provider.test.AesGcmKatTest
+        --select-class=com.amazon.corretto.crypto.provider.test.AesCbcTest
+        --select-class=com.amazon.corretto.crypto.provider.test.AesCbcNistTest
+        --select-class=com.amazon.corretto.crypto.provider.test.AesCbcIso10126Test
```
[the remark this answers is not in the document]
geedo0: Yeah, this would be good to run across the entire test suite. Only reason not too would be if the code is not in good health today and fixing all the errors it flags would be prohibitively expensive.

CMakeLists.txt
```diff
@@ -804,17 +816,33 @@ add_custom_target(check-install-via-properties-with-debug
...
+        DEPENDS check-recursive-init
+        check-install-via-properties
+        check-install-via-properties-with-debug
+        check-junit
+        check-junit-SecurityManager
+        check-external-lib
+        check-junit-AesLazy
+        check-junit-AesEager
+        check-junit-DifferentTempDir
+        # check-with-jni-flag for macos, we disable this check due to failures in java.net.Inet6AddressImpl.lookupAllHostAddr
```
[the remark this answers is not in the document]
geedo0: You can refactor this to be more DRY by doing something like this.

```
set(check_targets check-recursive-init
    check-install-via-properties-with-debug
    check-install-via-properties
    check-junit
    check-install-via-properties-with-debug
    check-junit-SecurityManager
    check-junit
    check-external-lib
    check-junit-SecurityManager
    check-junit-AesLazy
    check-external-lib
    check-junit-AesEager
    check-junit-AesLazy
    check-junit-DifferentTempDir
    check-junit-AesEager
    check-junit-DifferentTempDir
)

if(${CMAKE_SYSTEM_NAME} MATCHES "Linux"))
  set(check_targets ${check_targets} check-with-jni-flag)
endif()
```

csrc/buffer.cpp
```diff
@@ -100,4 +100,74 @@ JBinaryBlob::~JBinaryBlob()
...
+    if (outputDirectByteBuffer != nullptr) {
+        // One should be null. In the Java layer, we ensure this.
+        if (outputArray != nullptr) {
+            throw java_ex(EX_ERROR,
+                "THIS SHOULD NOT BE REACHABLE. Both outputDirectByteBuffer and outputArray cannot be provided.");
+        }
+        output_ptr_ = (inputDirectByteBuffer == outputDirectByteBuffer)
+            ? input_ptr_
+            : ((uint8_t*)env->GetDirectBufferAddress(outputDirectByteBuffer));
+    }
```
geedo0, line 131: nit: you can extract this to a method for both input/output

geedo0 (dismissed):

CMakeLists.txt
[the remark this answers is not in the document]
amirhosv: Added all the unit tests.

CMakeLists.txt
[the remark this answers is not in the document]
amirhosv: Thanks. I added the link to the bug as a comment.

csrc/buffer.cpp
amirhosv: They are slightly different. For the second one (in this case `outputDirectByteBuffer`), we also check if it is the same as input: if yes, we avoid calling `GetDirectBufferAddress`.

CMakeLists.txt
```diff
@@ -804,17 +814,22 @@ add_custom_target(check-install-via-properties-with-debug
...
     check-junit-AesLazy
     check-junit-AesEager
-    check-junit-DifferentTempDir
-)
+    check-junit-DifferentTempDir)
+
+if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+  set(check_targets ${check_targets} check-with-jni-flag)
+endif()
+# For MacOS, check-with-jni-flag is not added due to failures in java.net.Inet6AddressImpl.lookupAllHostAddr: https://bugs.openjdk.org/browse/JDK-8205076
```
shipilev, line 830: Can you give a little more context why this exclusion? I thought I fixed JDK-8205076 for all platforms in 17.0.11. Do you need that fix in 11u?

geedo0 (approved):

CMakeLists.txt
shipilev: Yeah, it fails in current 11u. 11u backport is here: https://github.com/openjdk/jdk11u-dev/pull/2769 -- if approved, would be in October release. Does not block this PR, but maybe mention this check can be removed after 11u catches up?

shipilev: > The failure also happens for JDK 8

This is really weird, because this issue should not affect JDK 8. Looking into the raw logs, I see that tests look to be running with 11u still? Note the `java.base@11.0.23` below:

```
2024-06-13T15:07:34.5982930Z Complete job name: JDK8
...
2024-06-13T15:16:25.0335520Z FATAL ERROR in native method: Field type (static) mismatch in JNI get/set field operations
2024-06-13T15:16:25.0337710Z make[2]: *** [CMakeFiles/check-with-jni-flag.dir/all] Error 2
2024-06-13T15:16:25.0338320Z 	at java.net.Inet6AddressImpl.lookupAllHostAddr(java.base@11.0.23/Native Method)
2024-06-13T15:16:25.0339400Z make[1]: *** [CMakeFiles/coverage.dir/rule] Error 2
2024-06-13T15:16:25.0339970Z 	at java.net.InetAddress$PlatformNameService.lookupAllHostAddr(java.base@11.0.23/InetAddress.java:930)
2024-06-13T15:16:25.0342430Z make: *** [coverage] Error 2
2024-06-13T15:16:25.0345990Z 	at java.net.InetAddress.getAddressesFromNameService(java.base@11.0.23/InetAddress.java:1543)
2024-06-13T15:16:25.0349340Z 
```

amirhosv: You're right. It uses JDK11 still for coverage target.

alexw91 (approved):

PR #385, opened 2024-06-21
Register ACCP's SecureRandom by default

amirhosv:
*Description of changes:*

+ Register ACCP's SecureRandom by default.
+ This change also updates the AWS-LC's commits are used with FIPS and none-FIPS builds.
  + As part of updating AWS-LC, EC_METHOD_get_field_type, and RAND_poll are removed since they always return constant values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,10 +252,6 @@ The FIPS builds use a different version of AWS-LC along with `FIPS=1` build flag
 AWS-LC will have FIPS certification. As a result, ACCP in FIPS mode only uses a version of AWS-LC
 that has FIPS certification or it will have in future.
 
-For performance reasons, ACCP does not register a SecureRandom implementation in FIPS mode.
-Relevant operations within the FIPS module boundary (e.g. key generation, non-deterministic signing, etc.) will still use AWS-LC's internal DRBG.
-Users who require ACCP to provide FIPS-validated pseudo-randomness _outside_ the module boundary via SecureRandom should set `registerSecureRandom=true`.
-
 When changing between FIPS and non-FIPS builds, be sure to do a full `clean` of your build environment.
 
 ##### All targets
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -348,13 +344,10 @@ Thus, these should all be set on the JVM command line using `-D`.
...
   If `true`, then ACCP will register a SecureRandom implementation (`LibCryptoRng`) backed by AWS-LC
   Else, ACCP will not register a SecureRandom implementation, meaning that the JCA will source SecureRandom instances from another registered provider. AWS-LC will still use its internal DRBG for key generation and other operations requiring secure pseudo-randomness.
-  LibCryptoRng is very fast during steady state operation in all cases. In FIPS mode, however, AWS-LC-FIPS's CPU jitter-based entropy source incurs a ~10ms initialization cost for every new thread.
-  This means that there is a slight "pause" before ACCP FIPS's SecureRandom can produce pseudo-random bytes in highly threaded environments.
-  Because, in extreme cases this could present an availability risk, we do not register LibCryptoRng by default in configurations where this initialization cost is incurred (i.e. FIPS mode).
-  Non-FIPS AWS-LC does not use CPU jitter for its DRBG seed's entropy, and therefore does not incur this initialization cost, therefore we register LibCryptoRng by default when not in FIPS mode.
+  LibCryptoRng is very fast during steady state operation in all cases.
```
geedo0, line 350: I don't think we need to state this explicitly.

csrc/ec_utils.cpp
```diff
@@ -97,16 +97,6 @@ JNIEXPORT jint JNICALL Java_com_amazon_corretto_crypto_provider_EcUtils_curveNam
         BigNumObj orderBN;
 
         const EC_POINT* generator = NULL;
```
geedo0, line 99: All the changes to this file don't seem relevant to SecureRandom's default value. What's going on here?

geedo0 (changes requested):

csrc/ec_utils.cpp
amirhosv: [EC_METHOD_get_field_type always returns NID_X9_62_prime_field](https://github.com/aws/aws-lc/blob/AWS-LC-FIPS-2.0.13/include/openssl/ec.h#L369). In AWS-LC v1.30.1 it is marked as deprecated, so I've removed it and simplified the code according to the fact that it always returns `NID_X9_62_prime_field`.

README.md
amirhosv: Done.

csrc/ec_utils.cpp
geedo0: For the future, it would help to explain this kind of stuff in the PR description so there's less confusion. It would also be good hygiene to have separate PRs for the secureRandom change and bringing LC up to date.

geedo0 (approved):

csrc/loader.cpp
```diff
@@ -32,9 +32,6 @@ void initialize()
     CRYPTO_library_init();
     ERR_load_crypto_strings();
     OpenSSL_add_all_digests();
-
-    // seed the PRNG
-    RAND_poll();
```
WillChilds-Klein, line 37: we don't need to do this anymore?

WillChilds-Klein: ah, i see. it's a no-op in AWS-LC

https://github.com/aws/aws-lc/blob/4368aaa6975ba41bd76d3bb12fac54c4680247fb/crypto/rand_extra/rand_extra.c#L49

WillChilds-Klein (approved):

csrc/loader.cpp
amirhosv: Yes. [This is no op](https://github.com/aws/aws-lc/blob/AWS-LC-FIPS-2.0.13/include/openssl/rand.h#L90) and it's marked deprecated in v1.30.1.

csrc/ec_utils.cpp
amirhosv: I totally agree. I think we should update LC versions more often in separate PRs.

PR #389, opened 2024-06-29
Avoid using RSA_blinding_off_temp_for_accp_compatibility

amirhosv:
*Description of changes:*

+ RSA_blinding_off_temp_for_accp_compatibility is deprecated and for non-FIPS builds, RSA_new_private_key_no_e must be used to create private RSA keys that do not have public exponent:
https://github.com/aws/aws-lc/blob/v1.30.1/include/openssl/rsa.h#L648

### DIFF
--- a/csrc/bn.cpp
+++ b/csrc/bn.cpp
@@ -68,4 +68,13 @@ jbyteArray bn2jarr(raii_env& env, const BIGNUM* bn)
     return jarr;
 }
 
+void bn_dup_into(BIGNUM** dst, BIGNUM const* src)
+{
+    BN_free(*dst);
+    *dst = BN_dup(src);
+    if (*dst == nullptr) {
+        throw_openssl("BN_dup failed.");
+    }
+}
+
 } // namespace AmazonCorrettoCryptoProvider
--- a/csrc/bn.h
+++ b/csrc/bn.h
@@ -139,6 +139,9 @@ public:
 
 inline BigNumObj bn_zero() { return BigNumObj(); }
 
+// Frees the data referenced by *dst, if any, and copies src to *dst.
+void bn_dup_into(BIGNUM** dst, BIGNUM const* src);
+
 } // namespace AmazonCorrettoCryptoProvider
 
 #endif
--- a/csrc/java_evp_keys.cpp
+++ b/csrc/java_evp_keys.cpp
@@ -496,21 +496,19 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_r
         if (privateExponentArr) {
             BigNumObj privExp = BigNumObj::fromJavaArray(env, privateExponentArr);
 
-            int res = 1;
             if (BN_is_zero(pubExp)) {
                 // RSA blinding can't be performed without |e|.
-                rsa.set(RSA_new_private_key_no_e(modulus, privExp));
+                rsa.set(new_private_RSA_key_with_no_e(modulus, privExp));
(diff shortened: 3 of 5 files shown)


csrc/bn.h
```diff
@@ -139,6 +139,9 @@ class BigNumObj {
 
 inline BigNumObj bn_zero() { return BigNumObj(); }
 
+// Copies src to dst.
```
geedo0, line 142: This is a misleading comment. It does not simply copy from src to dst, it frees whatever is behind the dst pointer, allocates new memory for a copy of src, and returns that into the dst pointer.

csrc/java_evp_keys.cpp
```diff
@@ -496,21 +496,19 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_r
...
-                rsa.set(RSA_new_private_key_no_e(modulus, privExp));
+                rsa.set(new_private_RSA_key_with_no_e(modulus, privExp));
+                // new_private_RSA_key_with_no_e does not take the ownership of its arguments
             } else {
-                res = RSA_set0_key(rsa, modulus, pubExp, privExp);
-            }
-
-            if (res != 1) {
-                throw_openssl(EX_RUNTIME_CRYPTO, "Unable to set RSA values");
+                if (1 != RSA_set0_key(rsa, modulus, pubExp, privExp)) {
```
geedo0, line 504: nit: the convention is `function() != 1`

geedo0 (approved):

csrc/java_evp_keys.cpp
amirhosv: Done.

csrc/bn.h
amirhosv: Updated the comment to reflect this.

geedo0 (approved):

csrc/bn.cpp
```diff
@@ -68,4 +68,13 @@ jbyteArray bn2jarr(raii_env& env, const BIGNUM* bn)
     return jarr;
 }
 
+void bn_dup_into(BIGNUM** dst, BIGNUM const* src)
+{
+    BN_free(*dst);
```
WillChilds-Klein, line 73: should we null-check `dst` before de-referencing?

csrc/keyutils.cpp
```diff
@@ -170,35 +168,36 @@ const EVP_MD* digestFromJstring(raii_env& env, jstring digestName)
...
 
-    // RSA_blinding_off_temp_for_accp_compatibility is marked deprecated, so we need to disable the
-    // "deprecated-declarations" check for this invocation.
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-    // We need to change this API once AWS-LC provides a method similar to the following:
-    // https://github.com/google/boringssl/blob/master/include/openssl/rsa.h#L630
-    RSA_blinding_off_temp_for_accp_compatibility(rsa);
-#pragma GCC diagnostic pop
+    RSA* result = RSA_new_private_key_no_e(n, d);
```
WillChilds-Klein, line 192: why is using this method preferable to doing it the FIPS way with the `dup_into`s?

amirhosv: In non-FIPS, the RSA structure is opaque and we don't have a way to turn off blinding. This method creates a private RSA key and sets the blinding flag properly. I think in the next FIPS release, the RSA structure will become opaque and such a method will become available.

WillChilds-Klein (approved):

PR #390, opened 2024-07-01
Add AWS-LC src & ref flags to Gradle build

sp717:
*Issue #, if available:*

*Description of changes:*
In case a user needs to build ACCP with a specific version of AWS-LC, this change adds 2 flags that can be passed to Gradle to specify a source directory for AWS-LC or a specific Git reference. 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,8 +22,8 @@ if (ext.isFips) {
 }
 
 // Check for user inputted git version ID.
-if (System.properties["AWSLC_REF"]) {
-    ext.awsLcGitVersionId = System.properties["AWSLC_REF"]
+if (System.env.AWSLC_GITVERSION) {
+    ext.awsLcGitVersionId = System.env.AWSLC_GITVERSION
 }
 
 ext.isLegacyBuild = Boolean.getBoolean('LEGACY_BUILD')
@@ -97,8 +97,8 @@ def getClangFormatVersion() {
 def awslcSrcPath = "${projectDir}/aws-lc/"
 
 // Check for user inputted AWS-LC source directory. 
-if (System.properties["AWSLC_SRC_DIR"]) {
-    awslcSrcPath = System.properties["AWSLC_SRC_DIR"]
+if (System.env.AWSLC_SRC_DIR) {
+    awslcSrcPath = System.env.AWSLC_SRC_DIR
 }
 
 // Execute cmake3 command to see if it exists. Mainly to support AL2.
--- a/build.gradle
+++ b/build.gradle
@@ -22,8 +22,8 @@ if (ext.isFips) {
 }
 
 // Check for user inputted git version ID.
-if (System.env.AWSLC_GITVERSION) {
-    ext.awsLcGitVersionId = System.env.AWSLC_GITVERSION
+if (System.properties["AWSLC_GITVERSION"]) {
+    ext.awsLcGitVersionId = System.properties["AWSLC_GITVERSION"]
 }
 
 ext.isLegacyBuild = Boolean.getBoolean('LEGACY_BUILD')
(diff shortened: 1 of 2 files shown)


WillChilds-Klein (approved):

build.gradle
```diff
@@ -20,6 +20,12 @@ if (ext.isFips) {
 } else {
     ext.awsLcGitVersionId = 'v1.30.1'
 }
+
+// Check for user inputted git version ID.
+if (System.properties["AWSLC_REF"]) {
```
geedo0, line 25: `REF` is a somewhat ambiguous qualifier and doesn't align with the nomenclature used by the variable you're assigning to. I'd prefer something like `AWSLC_GITVERSION` which is totally unambiguous.

geedo0:
What's the impetus for this change? Improving developer experience or is this feeding into new CI scripts?

build.gradle
```diff
@@ -20,6 +20,12 @@ if (ext.isFips) {
...
+    ext.awsLcGitVersionId = System.properties["AWSLC_REF"]
```
geedo0, line 26: Also, I'm pretty sure you meant to use `System.env.AWSLC_REF`. What you have here fetches from the JVM System Properties which is a totally different thing.

https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_environment_variables

geedo0 (changes requested):

build.gradle
sp717: Updated!

sp717:
> What's the impetus for this change? Improving developer experience or is this feeding into new CI scripts?

This change helps with updating AWS-LC's CI to include ACCP. The current ACCP build defaults to using v1.30.1 of AWS-LC, but we'd like to test against any new version of AWS-LC, which this update should solve.

build.gradle
sp717: Would that still be able to access input from the gradle build command using -D (example below)? I tested the change and it didn't seem to work the same way, but I may have misunderstood. 

ex:
./gradlew build DAWSLC_REF="version"

geedo0: I reviewed the conventions in our Gradle script, it looks like we've developed a preferences for using Java system properties like this. What you have here is consistent and that's good. Bigger picture, the nice thing about using environmental variables to manipulate a build is that they are managed external to Gradle and are part of the POSIX standard.

```
# Set them within the scope of your shell
export AWSLC_REF="version"
./gradlew build

# Set them within the scope of your gradle process
AWSLC_REF="version" ./gradlew build

# Set them as part of a python script executing a process...
env = {'AWSLC_REF' : 'version'}
subprocess.run(['gradlew', 'build'], env=env)

```

geedo0 (dismissed):

build.gradle
```diff
@@ -90,6 +96,11 @@ def getClangFormatVersion() {
 
 def awslcSrcPath = "${projectDir}/aws-lc/"
 
+// Check for user inputted AWS-LC source directory. 
+if (System.properties["AWSLC_SRC_DIR"]) {
+    awslcSrcPath = System.properties["AWSLC_SRC_DIR"]
```
amirhosv, line 101: The error message [here](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/build.gradle#L177) needs to be updated. In case the user sets this property to a custom path and they have not checked out aws-lc's repo, then running git submodule is not going to help.

build.gradle
amirhosv: In retrospect, we should have used [Gradles properties](https://docs.gradle.org/current/userguide/project_properties.html) instead of JVM properties or even environment variables.

I'm not suggesting changing that in this PR since the use of JVM properties has been preferred.

build.gradle
sp717: Updated! Would we have to also add error cases for non-empty directories that don't contain the required source files, or should we let the subsequent build commands throw those errors?

amirhosv: I don't see it needed.

amirhosv (approved):

WillChilds-Klein (approved):

PR #394, opened 2024-07-29
Add Ed25519 DSA 

sp717:
*Issue #, if available:*

*Description of changes:*

### Description:

This PR adds Ed25519 digital signatures. The relevant classes and functions are added as part of ACCP’s EVP Key, Signature, and Keyfactory classes. Ed25519 pre hash mode has not been implemented, as it is dependent on the JDK 15+ interfaces that we cannot currently access.

We are unable to use the NamedParameterSpec, EdECKey, EdECPrivateKey, and EdECPublicKey classes for our implementation, as ACCP must be compiled on JDK 8, and the EdDSA classes are only available on JDK versions 15+. As a result, the implementation utilizes the PrivateKey and PublicKey parent classes available in older java versions for Ed keys.  

SignatureRaw from the EVP API was used for Ed25519 signatures, as this form of digital signatures does not support digest updates. The signing process can only take place after the entire message has been received (one-shot signing). EvpSignatureRaw matches this functionality (only updating the message via a java buffer and passing the complete message to native methods for signing at the end), whereas the standard EvpSignature allows for partial signing and verifying. 

The pkcs82evp native method was updated to allow the ACCP key factory to import BouncyCastle (BC) private keys. BC encodes the EdDSA public key as an additional field in the PKCS8 encoding of the private key. This caused issues when importing the private key to ACCP, as the method used for the translation (d2i_PKCS8_PRIV_KEY_INFO) was unable to process the additional information. We switched the implementation to use d2i_PrivateKey instead, which successfully pulls the private key from the PKCS8 encoding, regardless of the additional field. 

To fix versioning issues for JDK < 15, JUnit flags were used for tests, and a boolean checking the Java version was used to conditionally register Ed25519 in AmazonCorrettoCryptoProvider.java. 

### Testing:

New unit tests are included in EdDSATest.java. These include tests for invalid inputs, key generation, sign/verify functionality, and interoperability with the BC and SunEC providers. 

### To Do:

- [x] - Update Readme
- [ ] - Update Changelog
- [x] - Add any missing unit tests
- [x] - Update Threadstorm test - https://github.com/corretto/amazon-corretto-crypto-provider/blob/d71028c67c2e06cadf44c5eaa138e11d26192cb9/tst/com/amazon/corretto/crypto/provider/test/KeyReuseThreadStormTest.java#L38-L45

### DIFF
--- a/csrc/ed_gen.cpp
+++ b/csrc/ed_gen.cpp
@@ -11,7 +11,7 @@ void generateEdKey(raii_env* env, EVP_PKEY_auto& key)
 {
     EVP_PKEY_CTX_auto ctx = EVP_PKEY_CTX_auto::from(EVP_PKEY_CTX_new_id(EVP_PKEY_ED25519, nullptr));
     CHECK_OPENSSL(ctx.isInitialized());
-    CHECK_OPENSSL(EVP_PKEY_keygen_init(ctx) > 0);
+    CHECK_OPENSSL(EVP_PKEY_keygen_init(ctx) == 1);
     CHECK_OPENSSL(EVP_PKEY_keygen(ctx, key.getAddressOfPtr()));
 }
 
--- a/csrc/ed_gen.cpp
+++ b/csrc/ed_gen.cpp
@@ -1,6 +1,9 @@
 // Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+#include <openssl/evp.h>
 #include "auto_free.h"
+#include "env.h"
+#include "generated-headers.h"
 
 using namespace AmazonCorrettoCryptoProvider;
 
--- a/csrc/ed_gen.cpp
+++ b/csrc/ed_gen.cpp
@@ -1,9 +1,9 @@
 // Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-#include <openssl/evp.h>
 #include "auto_free.h"
 #include "env.h"
 #include "generated-headers.h"
+#include <openssl/evp.h>
 
 using namespace AmazonCorrettoCryptoProvider;
 
--- a/csrc/ed_gen.cpp
+++ b/csrc/ed_gen.cpp
@@ -1,0 +1,35 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+#include "auto_free.h"
+#include "bn.h"
+#include "buffer.h"
+#include "env.h"
+#include "generated-headers.h"
+#include "keyutils.h"
+#include "util.h"
+#include <openssl/bn.h>
+#include <openssl/evp.h>
+#include <memory>
+
+using namespace AmazonCorrettoCryptoProvider;
+
+void generateEdEcKey(raii_env* env, EVP_PKEY_auto& key)
+{
+    EVP_PKEY_CTX_auto ctx = EVP_PKEY_CTX_auto::from(EVP_PKEY_CTX_new_id(EVP_PKEY_ED25519, nullptr));
+    CHECK_OPENSSL(ctx.isInitialized());
+    CHECK_OPENSSL(EVP_PKEY_keygen_init(ctx) > 0);
+    CHECK_OPENSSL(EVP_PKEY_keygen(ctx, key.getAddressOfPtr()));
+}
+
+JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EdGen_generateEvpEdEcKey(JNIEnv* pEnv, jclass)
+{
+    try {
+        raii_env env(pEnv);
+        EVP_PKEY_auto key;
+        generateEdEcKey(&env, key);
+        return reinterpret_cast<jlong>(key.take());
+    } catch (java_ex& ex) {
+        ex.throw_to_java(pEnv);
+    }
+    return 0;
+}
--- a/csrc/ed_gen.cpp
+++ b/csrc/ed_gen.cpp
@@ -7,7 +7,7 @@
 
 using namespace AmazonCorrettoCryptoProvider;
 
-void generateEdKey(raii_env* env, EVP_PKEY_auto& key)
+void generateEdKey(EVP_PKEY_auto& key)
 {
     EVP_PKEY_CTX_auto ctx = EVP_PKEY_CTX_auto::from(EVP_PKEY_CTX_new_id(EVP_PKEY_ED25519, nullptr));
     CHECK_OPENSSL(ctx.isInitialized());
@@ -20,7 +20,7 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EdGen_generateE
     try {
         raii_env env(pEnv);
         EVP_PKEY_auto key;
-        generateEdKey(&env, key);
+        generateEdKey(key);
         return reinterpret_cast<jlong>(key.take());
     } catch (java_ex& ex) {
         ex.throw_to_java(pEnv);
--- a/csrc/ed_gen.cpp
+++ b/csrc/ed_gen.cpp
@@ -1,15 +1,6 @@
 // Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 #include "auto_free.h"
-#include "bn.h"
-#include "buffer.h"
-#include "env.h"
-#include "generated-headers.h"
-#include "keyutils.h"
-#include "util.h"
-#include <openssl/bn.h>
-#include <openssl/evp.h>
-#include <memory>
 
 using namespace AmazonCorrettoCryptoProvider;
 
--- a/csrc/ed_gen.cpp
+++ b/csrc/ed_gen.cpp
@@ -13,7 +13,7 @@
 
 using namespace AmazonCorrettoCryptoProvider;
 
-void generateEdEcKey(raii_env* env, EVP_PKEY_auto& key)
+void generateEdKey(raii_env* env, EVP_PKEY_auto& key)
 {
     EVP_PKEY_CTX_auto ctx = EVP_PKEY_CTX_auto::from(EVP_PKEY_CTX_new_id(EVP_PKEY_ED25519, nullptr));
     CHECK_OPENSSL(ctx.isInitialized());
@@ -21,12 +21,12 @@ void generateEdEcKey(raii_env* env, EVP_PKEY_auto& key)
     CHECK_OPENSSL(EVP_PKEY_keygen(ctx, key.getAddressOfPtr()));
 }
 
-JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EdGen_generateEvpEdEcKey(JNIEnv* pEnv, jclass)
+JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EdGen_generateEvpEdKey(JNIEnv* pEnv, jclass)
 {
     try {
         raii_env env(pEnv);
         EVP_PKEY_auto key;
-        generateEdEcKey(&env, key);
+        generateEdKey(&env, key);
         return reinterpret_cast<jlong>(key.take());
     } catch (java_ex& ex) {
         ex.throw_to_java(pEnv);
--- a/csrc/java_evp_keys.cpp
+++ b/csrc/java_evp_keys.cpp
@@ -342,19 +342,12 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpEdPriva
 
         EVP_PKEY* key = reinterpret_cast<EVP_PKEY*>(keyHandle);
 
-        EVP_PKEY_CTX_auto ctx = EVP_PKEY_CTX_auto::from(EVP_PKEY_CTX_new(key, nullptr));
-        CHECK_OPENSSL(ctx.isInitialized());
+        size_t bufSize;
 
+        CHECK_OPENSSL(EVP_PKEY_get_raw_private_key(key, NULL, &bufSize) == 1);
         OPENSSL_buffer_auto privateKeyBuffer;
-        PKCS8_PRIV_KEY_INFO_auto pkcs8 = PKCS8_PRIV_KEY_INFO_auto::from(EVP_PKEY2PKCS8(key));
-        CHECK_OPENSSL(pkcs8.isInitialized());
-
-        // This next line allocates memory
-        int bufInt = i2d_PKCS8_PRIV_KEY_INFO(pkcs8, &privateKeyBuffer);
-        CHECK_OPENSSL(bufInt > 0);
-
-        size_t bufSize = (size_t)bufInt;
-        CHECK_OPENSSL(EVP_PKEY_get_raw_private_key(key, privateKeyBuffer, &bufSize) > 0);
+        privateKeyBuffer.buf = (uint8_t*)OPENSSL_malloc(bufSize);
+        CHECK_OPENSSL(EVP_PKEY_get_raw_private_key(key, privateKeyBuffer, &bufSize) == 1);
 
         result = env->NewByteArray(bufSize);
         env->SetByteArrayRegion(result, 0, bufSize, privateKeyBuffer);
--- a/csrc/java_evp_keys.cpp
+++ b/csrc/java_evp_keys.cpp
@@ -350,14 +350,15 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpEdPriva
         CHECK_OPENSSL(pkcs8.isInitialized());
 
         // This next line allocates memory
-        int bufLen = i2d_PKCS8_PRIV_KEY_INFO(pkcs8, &privateKeyBuffer);
-        CHECK_OPENSSL(bufLen > 0);
+        int bufInt = i2d_PKCS8_PRIV_KEY_INFO(pkcs8, &privateKeyBuffer);
+        CHECK_OPENSSL(bufInt > 0);
 
-        size_t bufLen2 = (size_t)bufLen;
-        CHECK_OPENSSL(EVP_PKEY_get_raw_private_key(key, privateKeyBuffer, &bufLen2) > 0);
+        size_t bufSize = (size_t)bufInt;
+        CHECK_OPENSSL(bufSize == (size_t)bufInt);
+        CHECK_OPENSSL(EVP_PKEY_get_raw_private_key(key, privateKeyBuffer, &bufSize) > 0);
 
-        result = env->NewByteArray(bufLen2);
-        env->SetByteArrayRegion(result, 0, bufLen2, privateKeyBuffer);
+        result = env->NewByteArray(bufSize);
+        env->SetByteArrayRegion(result, 0, bufSize, privateKeyBuffer);
     } catch (java_ex& ex) {
         ex.throw_to_java(pEnv);
     }
--- a/csrc/java_evp_keys.cpp
+++ b/csrc/java_evp_keys.cpp
@@ -345,15 +345,14 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpEdPriva
         size_t bufSize;
 
         CHECK_OPENSSL(EVP_PKEY_get_raw_private_key(key, NULL, &bufSize) == 1);
-        OPENSSL_buffer_auto privateKeyBuffer;
-        privateKeyBuffer.buf = (uint8_t*)OPENSSL_malloc(bufSize);
-        CHECK_OPENSSL(EVP_PKEY_get_raw_private_key(key, privateKeyBuffer, &bufSize) == 1);
+        SimpleBuffer privateKeyBuffer(bufSize);
+        CHECK_OPENSSL(EVP_PKEY_get_raw_private_key(key, privateKeyBuffer.get_buffer(), &bufSize) == 1);
 
         result = env->NewByteArray(bufSize);
         if (!result) {
             throw_java_ex(EX_OOM, "Unable to allocate private key array");
         }
-        env->SetByteArrayRegion(result, 0, bufSize, privateKeyBuffer);
+        env->SetByteArrayRegion(result, 0, bufSize, (jbyte*)privateKeyBuffer.get_buffer());
     } catch (java_ex& ex) {
         ex.throw_to_java(pEnv);
     }
--- a/csrc/java_evp_keys.cpp
+++ b/csrc/java_evp_keys.cpp
@@ -104,7 +104,7 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_p
 
         {
             jni_borrow borrow = jni_borrow(env, pkcs8Buff, "pkcs8Buff");
-            result.set(der2EvpPrivateKey(borrow, derLen, shouldCheckPrivate, EX_INVALID_KEY_SPEC));
+            result.set(der2EvpPrivateKey(borrow, derLen, nativeValue, shouldCheckPrivate, EX_INVALID_KEY_SPEC));
             if (EVP_PKEY_base_id(result) != nativeValue) {
                 throw_java_ex(EX_INVALID_KEY_SPEC, "Incorrect key type");
             }
@@ -326,6 +326,42 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpEcPriva
         ex.throw_to_java(pEnv);
         return NULL;
     }
+}
+
+/*
+ * Class:     com_amazon_corretto_crypto_provider_EvpEdEcPrivateKey
+ * Method:    getPrivateKey
+ */
+JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpEdEcPrivateKey_getPrivateKey(
+    JNIEnv* pEnv, jclass, jlong keyHandle)
+{
+    jbyteArray result = NULL;
+
+    try {
+        raii_env env(pEnv);
+
+        EVP_PKEY* key = reinterpret_cast<EVP_PKEY*>(keyHandle);
+
+        EVP_PKEY_CTX_auto ctx = EVP_PKEY_CTX_auto::from(EVP_PKEY_CTX_new(key, nullptr));
+        CHECK_OPENSSL(ctx.isInitialized());
+
+        OPENSSL_buffer_auto privateKeyBuffer;
+        PKCS8_PRIV_KEY_INFO_auto pkcs8 = PKCS8_PRIV_KEY_INFO_auto::from(EVP_PKEY2PKCS8(key));
+        CHECK_OPENSSL(pkcs8.isInitialized());
+
+        // This next line allocates memory
+        int bufLen = i2d_PKCS8_PRIV_KEY_INFO(pkcs8, &privateKeyBuffer);
+        CHECK_OPENSSL(bufLen > 0);
+
+        size_t bufLen2 = (size_t)bufLen;
+        CHECK_OPENSSL(EVP_PKEY_get_raw_private_key(key, privateKeyBuffer, &bufLen2) > 0);
+
+        result = env->NewByteArray(bufLen2);
+        env->SetByteArrayRegion(result, 0, bufLen2, privateKeyBuffer);
+    } catch (java_ex& ex) {
+        ex.throw_to_java(pEnv);
+    }
+    return result;
 }
 
 JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpRsaKey_getModulus(
--- a/csrc/java_evp_keys.cpp
+++ b/csrc/java_evp_keys.cpp
@@ -329,10 +329,10 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpEcPriva
 }
 
 /*
- * Class:     com_amazon_corretto_crypto_provider_EvpEdEcPrivateKey
+ * Class:     com_amazon_corretto_crypto_provider_EvpEdPrivateKey
  * Method:    getPrivateKey
  */
-JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpEdEcPrivateKey_getPrivateKey(
+JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpEdPrivateKey_getPrivateKey(
     JNIEnv* pEnv, jclass, jlong keyHandle)
 {
     jbyteArray result = NULL;
--- a/csrc/java_evp_keys.cpp
+++ b/csrc/java_evp_keys.cpp
@@ -93,7 +93,7 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpKey_enc
  * Signature: ([BI)J
  */
 JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_pkcs82Evp(
-    JNIEnv* pEnv, jclass, jbyteArray pkcs8der, jint nativeValue, jboolean shouldCheckPrivate)
+    JNIEnv* pEnv, jclass, jbyteArray pkcs8der, jint evpType, jboolean shouldCheckPrivate)
 {
     try {
         raii_env env(pEnv);
@@ -104,8 +104,8 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_p
 
         {
             jni_borrow borrow = jni_borrow(env, pkcs8Buff, "pkcs8Buff");
-            result.set(der2EvpPrivateKey(borrow, derLen, nativeValue, shouldCheckPrivate, EX_INVALID_KEY_SPEC));
-            if (EVP_PKEY_base_id(result) != nativeValue) {
+            result.set(der2EvpPrivateKey(borrow, derLen, evpType, shouldCheckPrivate, EX_INVALID_KEY_SPEC));
+            if (EVP_PKEY_base_id(result) != evpType) {
                 throw_java_ex(EX_INVALID_KEY_SPEC, "Incorrect key type");
             }
         }
@@ -122,7 +122,7 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_p
  * Signature: ([BI)J
  */
 JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_x5092Evp(
-    JNIEnv* pEnv, jclass, jbyteArray x509der, jint nativeValue)
+    JNIEnv* pEnv, jclass, jbyteArray x509der, jint evpType)
 {
     try {
         raii_env env(pEnv);
@@ -134,7 +134,7 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_x
         {
             jni_borrow borrow = jni_borrow(env, x509Buff, "x509Buff");
             result.set(der2EvpPublicKey(borrow, derLen, EX_INVALID_KEY_SPEC));
-            if (EVP_PKEY_base_id(result) != nativeValue) {
+            if (EVP_PKEY_base_id(result) != evpType) {
                 throw_java_ex(EX_INVALID_KEY_SPEC, "Incorrect key type");
             }
         }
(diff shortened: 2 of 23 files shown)


src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -441,6 +453,10 @@ public AmazonCorrettoCryptoProvider() {
     this.shouldRegisterSecureRandom =
         Utils.getBooleanProperty(PROPERTY_REGISTER_SECURE_RANDOM, true);
 
+    // Register EdDSA if JDK version >= 15
+    String[] version = System.getProperty("java.version").split("\\.");
```
WillChilds-Klein, line 457: we have a [util](https://github.com/corretto/amazon-corretto-crypto-provider/blob/cbe344117f5a2bcdd6777100857c506cbc2a4904/src/com/amazon/corretto/crypto/provider/Utils.java#L535) for this

geedo0:
Git soapbox incoming - It's good practice to squash your commits prior to submitting your code a review (especially when you've got a messy history like this). That way, you can hide all of the intermediate states of the code from the reviewers and starts the whole thing off on a cleaner slate. Subsequent iterations once the review has started should not be squashed to allow for inter-revision diffs. Once the PR is approved we'll do a squash and rebase to cleanly apply it to the trunk.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -441,6 +453,9 @@ public AmazonCorrettoCryptoProvider() {
     this.shouldRegisterSecureRandom =
         Utils.getBooleanProperty(PROPERTY_REGISTER_SECURE_RANDOM, true);
 
+    // Register EdDSA if JDK version >= 15
```
alexw91, line 456: This comment should be updated to explain *why* JDK >= 15 is needed

geedo0:
> We are unable to use the NamedParameterSpec, EdECKey, EdECPrivateKey, and EdECPublicKey classes for our implementation, as ACCP must be compiled on JDK 8, and the EdDSA classes are only available on JDK versions 15+. As a result, the implementation utilizes the PrivateKey and PublicKey parent classes available in older java versions for Ed keys.

This sounds like a short term one-way door decision that we'd be stuck with for a long time. This breaks forwards compatibility with Java code that implements against the expected JDK classes. e.g. If the JSSE were to use EdDSA it would not be able to delegate to ACCP for the implementation without an upstream code change. Do we have a story around how we may add compatibility for the standard Java classes? How much investigation was done around working around the JDK 8 limitation and solving that? Do we have hard requirements to support Ed25519 for JDK8 code bases?

src/com/amazon/corretto/crypto/provider/EdGen.java
```diff
@@ -0,0 +1,38 @@
...
+import java.security.KeyPair;
+import java.security.KeyPairGeneratorSpi;
+import java.security.SecureRandom;
+
+class EdGen extends KeyPairGeneratorSpi {
+  /**
+   * Generates a new Ed25519 key and returns a pointer to it.
+   *
+   * @param params a native pointer created by {@link #buildEcParams(int)}
+   * @param checkConsistency Run additional consistency checks on the generated keypair
```
geedo0, line 14: Java doc does not line up, this function accepts no parameters.

src/com/amazon/corretto/crypto/provider/EdGen.java
```diff
@@ -0,0 +1,38 @@
...
+  private final AmazonCorrettoCryptoProvider provider_;
+
+  EdGen(AmazonCorrettoCryptoProvider provider) {
+    Loader.checkNativeLibraryAvailability();
+    provider_ = provider;
+  }
+
+  public void initialize(int keysize, SecureRandom random) {
+    // Has some behavior in Java, but throws error as placeholder for now.
+    throw new UnsupportedOperationException();
```
geedo0, line 27: This needs to be taken care of before merging. Minimally, you'll want to delegate to the other `initialize` method with some sensible defaults.

src/com/amazon/corretto/crypto/provider/EvpEdEcKey.java
```diff
@@ -0,0 +1,11 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider;
+
+abstract class EvpEdEcKey extends EvpKey {
```
geedo0, line 5: Drop the EC, this naming convention further increases the confusion around all things related to elliptic curves.

csrc/ed_gen.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "auto_free.h"
+#include "bn.h"
+#include "buffer.h"
+#include "env.h"
+#include "generated-headers.h"
+#include "keyutils.h"
+#include "util.h"
+#include <openssl/bn.h>
+#include <openssl/evp.h>
+#include <memory>
```
geedo0, line 12: Clean up your includes. You seem to have more includes that actual symbols here which suggests you're including more than you use. Also, follow the convention of system libraries up top and all project headers below.

The Google [C-style guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) is excellent if you want to improve your style.

csrc/java_evp_keys.cpp
```diff
@@ -328,6 +328,42 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpEcPriva
...
+
+        OPENSSL_buffer_auto privateKeyBuffer;
+        PKCS8_PRIV_KEY_INFO_auto pkcs8 = PKCS8_PRIV_KEY_INFO_auto::from(EVP_PKEY2PKCS8(key));
+        CHECK_OPENSSL(pkcs8.isInitialized());
+
+        // This next line allocates memory
+        int bufLen = i2d_PKCS8_PRIV_KEY_INFO(pkcs8, &privateKeyBuffer);
+        CHECK_OPENSSL(bufLen > 0);
+
+        size_t bufLen2 = (size_t)bufLen;
```
geedo0, line 356: Never a fan of just adding 2 to a variable name when you can choose something more expressive. Maybe you can go `bufInt` and `bufSize`? Since you have both here, might as well assert equality on the two as well.

csrc/keyutils.cpp
```diff
@@ -7,26 +7,27 @@
...
 #include <openssl/rsa.h>
+#include <iostream>
 
 namespace AmazonCorrettoCryptoProvider {
 
-EVP_PKEY* der2EvpPrivateKey(
-    const unsigned char* der, const int derLen, bool shouldCheckPrivate, const char* javaExceptionClass)
+EVP_PKEY* der2EvpPrivateKey(const unsigned char* der,
+    const int derLen,
+    const int nativeValue,
```
geedo0, line 16: `nativeValue` means nothing to me. Can you rename this to something like `evpType`?

geedo0 (changes requested):
Covered key generation. Pending KeyFactory, Signature, and Tests.

geedo0 (approved):

geedo0 (changes requested):
clicked the wrong button

src/com/amazon/corretto/crypto/provider/EvpEdEcKey.java
sp717: Updated!

src/com/amazon/corretto/crypto/provider/EdGen.java
sp717: Updated

csrc/sign.cpp
```diff
@@ -66,7 +66,7 @@ bool initializeContext(raii_env& env,
     EVP_PKEY_up_ref(pKey);
     ctx->setKey(pKey);
 
-    if (md != nullptr) {
+    if (md != nullptr || EVP_PKEY_id(pKey) == EVP_PKEY_ED25519) {
```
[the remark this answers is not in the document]
sp717: `md` is implicitly null, but we need to follow that path for Ed25519 signatures, since we are using the EVP_DigestSign method that relies on DigestCtx being set (as opposed to the KeyCtx being used, as in L84).

csrc/keyutils.cpp
sp717: Updated. `nativeValue` was used in the [pkcs82evp](https://github.com/corretto/amazon-corretto-crypto-provider/blob/d71028c67c2e06cadf44c5eaa138e11d26192cb9/csrc/java_evp_keys.cpp#L96) function, so I copied the naming over to `der2EvpPrivateKey`. Would it be worthwhile to also change the variable name there (pkcs82evp)?

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
```diff
@@ -0,0 +1,289 @@
...
+    jceGen = null;
+    bcGen = null;
+  }
+
+  @Test
+  public void uniqueKeyGen() {
+    final KeyPair kp1 = nativeGen.generateKeyPair();
+    final KeyPair kp2 = nativeGen.generateKeyPair();
+
+    assertTrue(!kp1.equals(kp2));
```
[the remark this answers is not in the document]
sp717: Fixed

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
```diff
@@ -0,0 +1,289 @@
...
+
+    assertTrue(privateKeyACCP.length == privateKeyJCE.length);
+    for (int i = 0; i < privateKeyACCP.length; i++) {
+      assertTrue(privateKeyACCP[i] == privateKeyJCE[i]);
+    }
+
+    assertTrue(publicKeyACCP.length == publicKeyJCE.length);
+    for (int i = 0; i < publicKeyACCP.length; i++) {
+      assertTrue(publicKeyACCP[i] == publicKeyJCE[i]);
+    }
```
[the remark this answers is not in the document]
sp717: Updated

csrc/java_evp_keys.cpp
```diff
@@ -350,14 +350,15 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpEdEcPri
...
         // This next line allocates memory
-        int bufLen = i2d_PKCS8_PRIV_KEY_INFO(pkcs8, &privateKeyBuffer);
-        CHECK_OPENSSL(bufLen > 0);
+        int bufInt = i2d_PKCS8_PRIV_KEY_INFO(pkcs8, &privateKeyBuffer);
+        CHECK_OPENSSL(bufInt > 0);
 
-        size_t bufLen2 = (size_t)bufLen;
-        CHECK_OPENSSL(EVP_PKEY_get_raw_private_key(key, privateKeyBuffer, &bufLen2) > 0);
+        size_t bufSize = (size_t)bufInt;
+        CHECK_OPENSSL(bufSize == (size_t)bufInt);
```
geedo0, line 357: Remove this, I misunderstood the code. I thought passing in the reference to `EVP_PKEY_get_raw_private_key` would return the bytes written and that would be useful to assert.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
```diff
@@ -55,7 +56,14 @@ public void uniqueKeyGen() {
     final KeyPair kp1 = nativeGen.generateKeyPair();
     final KeyPair kp2 = nativeGen.generateKeyPair();
 
-    assertTrue(!kp1.equals(kp2));
+    final byte [] pk1 = kp1.getPrivate().getEncoded();
```
geedo0, line 59: I'm 100% sure the style checker is going to ding you on this. should be `final byte[] foo`

csrc/keyutils.cpp
geedo0: That'd be nice, but if it looks like you'd end up changing a lot of unrelated code don't bother.

sp717:
> > We are unable to use the NamedParameterSpec, EdECKey, EdECPrivateKey, and EdECPublicKey classes for our implementation, as ACCP must be compiled on JDK 8, and the EdDSA classes are only available on JDK versions 15+. As a result, the implementation utilizes the PrivateKey and PublicKey parent classes available in older java versions for Ed keys.
> 
> This sounds like a short term one-way door decision that we'd be stuck with for a long time. This breaks forwards compatibility with Java code that implements against the expected JDK classes. e.g. If the JSSE were to use EdDSA it would not be able to delegate to ACCP for the implementation without an upstream code change. Do we have a story around how we may add compatibility for the standard Java classes? How much investigation was done around working around the JDK 8 limitation and solving that? Do we have hard requirements to support Ed25519 for JDK8 code bases?

I don't think it would be too much work to add compatibility for the standard Java classes if they became available. The EvpEdPrivateKey class would just need to implement EdEcPrivateKey (the java class name) instead of PrivateKey, and an override tag to be added to the getPrivateKey method. The EvpEdPublicKey class would need the parent class to be changed, and also the getPoint method to be implemented (the EdEcPoint class doesn't exist until JDK 15, so the method can't be implemented yet). The EvpEdKey parent class would have to implement the EdEcKey class from Java, and a getParams method would need to be added (returns relevant NamedParameterSpec). 

[NamedParameterSpec](https://docs.oracle.com/en/java/javase/17/docs//api/java.base/java/security/spec/NamedParameterSpec.html) is used to specify default parameters for Ed25519, Ed448, X25519, and X448. This isn't applicable at the moment, since we just have Ed25519 support, but can be added in the future if the rest are added. I can change the current Ed25519 class to a parent EdDSA class and make a child class for Ed25519 right now, so if/when Ed448 is added, Ed25519 is already separated into its own class.

The reason we must support JDK8 is that to build ACCP, everything must compile on JDK 8 [(See this file)](https://github.com/corretto/amazon-corretto-crypto-provider/blob/a71a758393570b12a54deb9439c22d55cb698839/CMakeLists.txt#L185). We tried removing that line (and the --release 9 line above it), but it caused errors with JNI.

csrc/keyutils.cpp
sp717: Sounds good, just updated the name to evpType in the native methods and corresponding java methods.

src/com/amazon/corretto/crypto/provider/EvpKeyType.java
```diff
@@ -17,7 +17,8 @@
 /** Corresponds to native constants in OpenSSL which represent keytypes. */
 enum EvpKeyType {
   RSA("RSA", 6, RSAPublicKey.class, RSAPrivateKey.class),
-  EC("EC", 408, ECPublicKey.class, ECPrivateKey.class);
+  EC("EC", 408, ECPublicKey.class, ECPrivateKey.class),
+  Ed25519("EdDSA", 949, PublicKey.class, PrivateKey.class);
```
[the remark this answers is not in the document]
sp717: Updated to EdDSA instead of Ed25519

csrc/ed_gen.cpp
sp717: Updated

geedo0:
> The reason we must support JDK8 is that to build ACCP, everything must compile on JDK 8 [(See this file)](https://github.com/corretto/amazon-corretto-crypto-provider/blob/a71a758393570b12a54deb9439c22d55cb698839/CMakeLists.txt#L185). We tried removing that line (and the --release 9 line above it), but it caused errors with JNI.

This is what I was most curious about. How far along this line of inquiry you went... It would be interesting to know if there's some way for us to conditionally compile source code that imports JDK15 classes if the correct JDK is available to make this body of work "more complete" and compatible with JDK15+ code bases. I also understand that the build as it stands requires JDK8, but I wanted to know if we had a specific requirement to support ed25519 on JDK8. Lacking that requirement opens the door for us to avoid these kinds of private namespace workarounds.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
```diff
@@ -0,0 +1,284 @@
...
+  public void jceInteropValidation() throws GeneralSecurityException {
+    // Generate keys with ACCP and use JCE KeyFactory to get equivalent JCE Keys
+    final KeyPair keyPair = nativeGen.generateKeyPair();
+
+    final PKCS8EncodedKeySpec privateKeyPkcs8 =
+        new PKCS8EncodedKeySpec(keyPair.getPrivate().getEncoded());
+    final X509EncodedKeySpec publicKeyX509 =
+        new X509EncodedKeySpec(keyPair.getPublic().getEncoded());
+
+    final KeyFactory kf = KeyFactory.getInstance("Ed25519", "SunEC");
```
amirhosv, line 125: Without passing the keys to a keyfactory, one should be able to use `KeyPair` for signing and verifying.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
```diff
@@ -0,0 +1,284 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+import org.junit.jupiter.api.parallel.ResourceAccessMode;
+import org.junit.jupiter.api.parallel.ResourceLock;
+
+@ExtendWith(TestResultLogger.class)
+@Execution(ExecutionMode.CONCURRENT)
+@ResourceLock(value = TestUtil.RESOURCE_GLOBAL, mode = ResourceAccessMode.READ)
+@EnabledForJreRange(min = JRE.JAVA_15)
```
amirhosv, line 29: We need tests for JDKs before 15. For example, we should be able to use the keys generated by ACCP to sign and verify with ACCP on JDKs prior to 15.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
```diff
@@ -0,0 +1,284 @@
...
+
+    final KeyFactory kf = KeyFactory.getInstance("Ed25519", "SunEC");
+    final PrivateKey privateKey = kf.generatePrivate(privateKeyPkcs8);
+    final PublicKey publicKey = kf.generatePublic(publicKeyX509);
+
+    // Set up ACCP and JCE Signature Instances
+    final Signature nativeSig = Signature.getInstance("Ed25519", NATIVE_PROVIDER);
+    final Signature jceSig = Signature.getInstance("Ed25519", "SunEC");
+
+    // Sign with ACCP and verify with SunEC
```
amirhosv, line 133: If I'm not mistaken, EdDSA is deterministic. In that case, the signature produced by ACCP should be the same as other providers. Please add the corresponding assertions.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
```diff
@@ -0,0 +1,284 @@
...
+    eddsa.update(message, 0, message.length);
+    final byte[] signature = eddsa.sign();
+
+    eddsa.initVerify(keyPair.getPublic());
+    eddsa.update(message, 0, message.length);
+    assertTrue(eddsa.verify(signature));
+  }
+
+  @Test
+  public void mismatchSignature() throws GeneralSecurityException {
```
amirhosv, line 222: Please add some compatibility tests to ensure that ACCP behaves the same as SunJCE when the signature verification fails. For example, they both throw exceptions or they both return false.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
sp717: Since the ACCP keys don't use the specific Ed classes from Java, we can't directly use them for sign/verify. Using KeyFactory is the current fix, but being able to import the Ed classes from JDK 15+ would allow full compatibility. That wasn't possible currently, since ACCP needs to compile on jdk 8.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
sp717: Resolved in latest commit.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
sp717: Updated

csrc/ed_gen.cpp
```diff
@@ -0,0 +1,26 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+#include "auto_free.h"
```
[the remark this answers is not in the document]
sp717: I see, I misunderstood before. Just updated to reflect this change.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
sp717: SunEC's EdDSA KeyFactory was added in JDK 15. I don't think I could include it in our key generation, since it would cause the same compilation issues.

src/com/amazon/corretto/crypto/provider/EdGen.java
sp717: I've added an implementation in latest commit that matches SunEC's behavior, but doesn't use the same parameter spec classes. Is that a sufficient change or should we try to match the implementation more closely?

csrc/java_evp_keys.cpp
```diff
@@ -328,6 +328,42 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpEcPriva
...
+
+        EVP_PKEY_CTX_auto ctx = EVP_PKEY_CTX_auto::from(EVP_PKEY_CTX_new(key, nullptr));
+        CHECK_OPENSSL(ctx.isInitialized());
+
+        OPENSSL_buffer_auto privateKeyBuffer;
+        PKCS8_PRIV_KEY_INFO_auto pkcs8 = PKCS8_PRIV_KEY_INFO_auto::from(EVP_PKEY2PKCS8(key));
+        CHECK_OPENSSL(pkcs8.isInitialized());
+
+        // This next line allocates memory
+        int bufInt = i2d_PKCS8_PRIV_KEY_INFO(pkcs8, &privateKeyBuffer);
```
torben-hansen, line 353: Check out `EVP_PKEY_get_raw_private_key()`. Simpler way to get the length.

torben-hansen: Another method: This function is type specific. So, you can just make the assumption you are taking an Ed25519 private key which has a fixed length.

I guess getting the size at runtime is slightly more memory safe (aws-lc does check the encoded size in the third argument though).

csrc/sign.cpp
```diff
@@ -452,8 +452,25 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpSignatu
...
-        {
-            size_t sigLength;
+        size_t sigLength;
+
+        int keyType = EVP_PKEY_id(reinterpret_cast<EVP_PKEY*>(pKey));
+
+        if (keyType == EVP_PKEY_ED25519) {
+            jni_borrow message(env, messageBuf, "message");
+
+            if (!EVP_DigestSign(ctx.getDigestCtx(), NULL, &sigLength, message.data(), message.len())) {
```
torben-hansen, line 462: type-specific branch - so you know the output length.

If you think it's more memory safe to get it at runtime. Can just use `EVP_DigestSignFinal(ctx.getDigestCtx(), NULL, &sigLength)` and save some logic.

sp717: EVP_DigestSignFinal does a [check in the beginning](https://github.com/aws/aws-lc/blob/0c52b12898c96f7de61933db31d7f1f681b10691/crypto/fipsmodule/evp/digestsign.c#L193-L196) that I believe would throw an error for Ed25519.

torben-hansen: you are quite right about that! doesn't work

src/com/amazon/corretto/crypto/provider/EdGen.java
```diff
@@ -18,8 +19,10 @@ class EdGen extends KeyPairGeneratorSpi {
   }
 
   public void initialize(int keysize, SecureRandom random) {
-    // Has some behavior in Java, but throws error as placeholder for now.
-    throw new UnsupportedOperationException();
+    if (keysize != 255) {
+      throw new InvalidParameterException("Params must be Ed25519.");
+    }
+    return;
```
geedo0, line 25: This needs to call the `initialize(AlgorithmParameterSpec params, SecureRandom random)` interface. The idea is that some folks don't necessarily care/know what to put in AlgorithmParameterSpec. This interface is a simpler one which allows them to use the algorithm with sensible parameters selected by the maintainer. The way you have it right now, a client can call this interface and end up with an uninitialized KeyPairGenerator.

See the javadoc [here](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/security/KeyPairGeneratorSpi.html#initialize(int,java.security.SecureRandom)

csrc/ed_gen.cpp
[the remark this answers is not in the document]
geedo0: Ahh interesting, didn't realize that Spotless enforced some rules around this. Looks good now

geedo0 (changes requested):

src/com/amazon/corretto/crypto/provider/EdGen.java
geedo0: Spoke offline to clear this up. This implementation does not need to be initialized and we do not support customizing the RNG. The JDK implementation of edwards curves does accept an AlgorithmParameterSpec which allows the user to select between ed448/ed25519 but we don't support that either. We considered using reflections to provide compatibility for this, but there's a lot more work involved for us to support drop-in capability with JDK.

csrc/ed_gen.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "env.h"
+#include "generated-headers.h"
+
+using namespace AmazonCorrettoCryptoProvider;
+
+void generateEdKey(raii_env* env, EVP_PKEY_auto& key)
+{
+    EVP_PKEY_CTX_auto ctx = EVP_PKEY_CTX_auto::from(EVP_PKEY_CTX_new_id(EVP_PKEY_ED25519, nullptr));
+    CHECK_OPENSSL(ctx.isInitialized());
+    CHECK_OPENSSL(EVP_PKEY_keygen_init(ctx) > 0);
```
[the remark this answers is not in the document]
sp717: Updated ` > 0` to `== 1`, will be included in next commit

csrc/java_evp_keys.cpp
sp717: Updated to use `EVP_PKEY_get_raw_private_key` to get bufSize

src/com/amazon/corretto/crypto/provider/EdGen.java
geedo0: Can you remove the if statement here and simply throw an UnsupportedException like you had originally? After that, I'd be happy to approve the PR.

sp717: Just updated.

geedo0 (dismissed):
Looking good, thanks for getting this all to work!

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
amirhosv: I don't think you even need to check JDK version. All you need is to see if `KeyFactory.getInstance("Ed25519")` doesn't throw an exception; if not, then it's a JDK version that supports Ed25519 keys.

sp717: Updated, but when running the benchmark with the new changes, noticed ~20% dropoff in performance for key pair generation. (160,000 ops/second --> 126,000 ops/second). Do we think that tradeoff is worth it for easier interoperability?

src/com/amazon/corretto/crypto/provider/EdGen.java
```diff
@@ -2,19 +2,31 @@
...
   private static native long generateEvpEdKey();
 
   private final AmazonCorrettoCryptoProvider provider_;
+  private final KeyFactory kf;
 
   EdGen(AmazonCorrettoCryptoProvider provider) {
     Loader.checkNativeLibraryAvailability();
     provider_ = provider;
+    try {
+      kf = KeyFactory.getInstance("EdDSA", "SunEC");
```
geedo0, line 26: I'm a bit concerned about the FIPS implications. It looks like you're trying to export the KeyPair using JDK based interfaces by tumbling our Key classes into the SunEC KeyFactory. This helps with the interop story, but like we covered earlier there's a whole lot more work to get full interop working and doing it piecemeal can be counterproductive. Since this is along the lines of ASN.1 encode/decode it's probably okay, but it does break down our "all crypto comes from AWS-LC" story. Would like to get some more consensus from the team on how we should handle this.

amirhosv: * There are places that ACCP relies on other providers for ASN.1 encode/decode, like [GCM parameters](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java#L281).
* Without this, one cannot use keys generated by ACCP to perform sign/verify with another provider.
* It is expected that the private (public) key generated by ACCP can be type cast to [EdECPrivateKey](https://docs.oracle.com/en/java/javase/17/docs//api/java.base/java/security/interfaces/EdECPrivateKey.html) ([EdECPublicKey](https://docs.oracle.com/en/java/javase/17/docs//api/java.base/java/security/interfaces/EdECPublicKey.html)).

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
```diff
@@ -0,0 +1,281 @@
...
+    final Signature nativeSig = Signature.getInstance("Ed25519", NATIVE_PROVIDER);
+    final Signature jceSig = Signature.getInstance("Ed25519", "SunEC");
+
+    nativeSig.initSign(kp.getPrivate());
+    nativeSig.update(message1, 0, message1.length);
+    final byte[] signature = nativeSig.sign();
+
+    nativeSig.initVerify(kp.getPublic());
+    nativeSig.update(message2, 0, message2.length);
+    assertTrue(!nativeSig.verify(signature));
```
amirhosv, line 228: Use assertFalse.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
```diff
@@ -0,0 +1,281 @@
...
+
+    // Sign with SunEC and verify with ACCP
+    jceSig.initSign(privateKey);
+    jceSig.update(message, 0, message.length);
+    final byte[] signatureJCE = jceSig.sign();
+    nativeSig.initVerify(publicKey);
+    nativeSig.update(message);
+    assertTrue(nativeSig.verify(signatureJCE), "JCE->Native: Ed25519");
+
+    assertTrue(Arrays.equals(signatureJCE, signatureACCP));
```
amirhosv, line 143: Use assertArrayEquals.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
```diff
@@ -0,0 +1,281 @@
...
+
+    final PrivateKey privateKey = keyPair.getPrivate();
+    final PublicKey publicKey = keyPair.getPublic();
+
+    // Set up ACCP and JCE Signature Instances
+    final Signature nativeSig = Signature.getInstance("Ed25519", NATIVE_PROVIDER);
+    final Signature jceSig = Signature.getInstance("Ed25519", "SunEC");
+
+    // Sign with ACCP and verify with SunEC
+    final byte[] message = new byte[] {0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
```
amirhosv, line 127: Can we have more tests here? Maybe more messages and of different lengths: from 1 byte to 1024?

amirhosv (commented):
Some minor comments.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
sp717: Updated

geedo0 (approved):

src/com/amazon/corretto/crypto/provider/EvpKeyFactory.java
```diff
@@ -302,4 +302,27 @@ protected <T extends KeySpec> T engineGetKeySpec(Key key, Class<T> keySpec)
...
   }
+
+  static class EdDSA extends EvpKeyFactory {
+
+    EdDSA(AmazonCorrettoCryptoProvider provider) {
+      super(EvpKeyType.EdDSA, provider);
+    }
+
+    @Override
+    protected PrivateKey engineGeneratePrivate(KeySpec keySpec) throws InvalidKeySpecException {
```
amirhosv, line 313: `final KeySpec`

src/com/amazon/corretto/crypto/provider/EvpKeyFactory.java
```diff
@@ -302,4 +302,27 @@ protected <T extends KeySpec> T engineGetKeySpec(Key key, Class<T> keySpec)
...
+      return super.engineGeneratePrivate(keySpec);
+    }
+
+    @Override
+    protected PublicKey engineGeneratePublic(KeySpec keySpec) throws InvalidKeySpecException {
```
amirhosv, line 318: `final KeySpec`

src/com/amazon/corretto/crypto/provider/EvpKeyFactory.java
```diff
@@ -302,4 +302,27 @@ protected <T extends KeySpec> T engineGetKeySpec(Key key, Class<T> keySpec)
...
+      return super.engineGeneratePublic(keySpec);
+    }
+
+    @Override
+    protected <T extends KeySpec> T engineGetKeySpec(Key key, Class<T> keySpec)
```
amirhosv, line 323: Arguments can be marked `final`.

amirhosv (commented):
Minor comments.

csrc/java_evp_keys.cpp
```diff
@@ -328,6 +328,35 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpEcPriva
...
+    jbyteArray result = NULL;
+
+    try {
+        raii_env env(pEnv);
+
+        EVP_PKEY* key = reinterpret_cast<EVP_PKEY*>(keyHandle);
+
+        size_t bufSize;
+
+        CHECK_OPENSSL(EVP_PKEY_get_raw_private_key(key, NULL, &bufSize) == 1);
```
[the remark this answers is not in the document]
sp717: It sets bufSize to the size and returns 1 on success.

csrc/ed_gen.cpp
```diff
@@ -0,0 +1,29 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+#include "auto_free.h"
+#include "env.h"
+#include "generated-headers.h"
+#include <openssl/evp.h>
+
+using namespace AmazonCorrettoCryptoProvider;
+
+void generateEdKey(raii_env* env, EVP_PKEY_auto& key)
```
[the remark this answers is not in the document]
sp717: removed

csrc/java_evp_keys.cpp
```diff
@@ -328,6 +328,35 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpEcPriva
...
+        OPENSSL_buffer_auto privateKeyBuffer;
+        privateKeyBuffer.buf = (uint8_t*)OPENSSL_malloc(bufSize);
```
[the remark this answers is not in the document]
sp717: I tried using SimpleBuffer originally, and I believe the issue was that [SimpleBuffer cannot be cast](https://github.com/corretto/amazon-corretto-crypto-provider/blob/10e8edd67a2a965ec72da8c2e41646b67ece8fae/csrc/buffer.h#L615-L631) to `unsigned char*` or `uint8_t*` but [OPENSSL_buffer_auto can](https://github.com/corretto/amazon-corretto-crypto-provider/blob/10e8edd67a2a965ec72da8c2e41646b67ece8fae/csrc/auto_free.h#L122), and that functionality is needed for the [EVP_PKEY_get_raw_private_key method](https://github.com/aws/aws-lc/blob/519c1c5b6fa9772c5fda81d9896295d92f9c370c/include/openssl/evp.h#L281-L282).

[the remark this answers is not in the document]
amirhosv: You need to use `SimpleBuffer::get_buffer`.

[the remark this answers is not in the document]
sp717: Updated

src/com/amazon/corretto/crypto/provider/EdGen.java
alexw91: I'm in favor of keeping this change to increase compatibility. This issue was accidentally caught by one of our own  basic cross-compatibility unit tests between SunEC and ACCP. Any existing Ed25519 users coming from SunJCE who are thinking about migrating to ACCP will likely add their own similar cross-compatibility unit test, and could easily run into this exact same issue themselves. It's better for ACCP to solve it once, than have many customers re-discover and re-solve this issue themselves.

WillChilds-Klein: Given likely use-cases for Ed25519 signature key generation, I agree that compatibility/user experience outweigh performance here. We can optimize the latter in the future dependent on customer need.

geedo0 (dismissed):

alexw91 (dismissed):

geedo0 (approved):

alexw91 (approved):

PR #395, opened 2024-08-05
Apply the rule of 3/5/0 to JByteArrayCritical et al.

fabrice102:
*Issue #, if available:* N/A

*Description of changes:*

Apply the [rule of 3/5/0](https://en.cppreference.com/w/cpp/language/rule_of_three) to `JByteArrayCritical`, `SimpleBuffer`, `JBinaryBlob`, `JIOBlobs`. This ensures that these objects cannot be copied.

Before, if they were copied (e.g., due to being passed as argument to a function), the destructor would be called twice, which would release the critical region twice or which may release the critical region while another part of the code is still using it.

### DIFF
--- a/csrc/buffer.h
+++ b/csrc/buffer.h
@@ -599,9 +599,11 @@ public:
     unsigned char* get();
 
 #ifdef HAVE_CPP11
-    // deleting copy constructor and copy assignment to satisfy rule of three
+    // deleting copy & move operations to satisfy rule of five
     JByteArrayCritical(const JByteArrayCritical&) = delete;
     JByteArrayCritical& operator=(const JByteArrayCritical&) = delete;
+    JByteArrayCritical(JByteArrayCritical&&) = delete;
+    JByteArrayCritical& operator=(JByteArrayCritical&&) = delete;
 #endif
 
 private:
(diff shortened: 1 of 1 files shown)


csrc/buffer.h
```diff
@@ -598,6 +598,12 @@ class JByteArrayCritical {
     ~JByteArrayCritical();
     unsigned char* get();
 
+#ifdef HAVE_CPP11
+    // deleting copy constructor and copy assignment to satisfy rule of three
```
geedo0, line 602: Should we delete the move constructor and move assignment functions as well to comply with the rule of five?

csrc/buffer.h
```diff
@@ -598,6 +598,12 @@ class JByteArrayCritical {
...
+    JByteArrayCritical(const JByteArrayCritical&) = delete;
+    JByteArrayCritical& operator=(const JByteArrayCritical&) = delete;
+#endif
```
geedo0, line 605: nit: We could add an #else here to handle the RHEL5 case by using empty/private overrides. I'd prefer fewer LOC than handling EOL/Obsolete OS's though.

csrc/buffer.h
fabrice102: You're right. Technically, removing just the copy operations is sufficient. But it's much cleaner to remove all of them. I'm fixing that:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all

csrc/buffer.h
fabrice102: What do you prefer? Removing them just on C++11 compilers should catch most of the potential mis-uses of the classes. But there may be a corner case where a pre-C++11 compiler does a copy while a C++11 compiler does not (e.g., because some `#ifdef` somewhere else makes the pre-C++11 code slightly different).

geedo0: I'd avoid adding code to handle RHEL5 here. Forget I ever brought it up.

geedo0 (approved):

amirhosv (approved):

PR #398, opened 2024-08-19
[PQ] Add experimental support for HPKE

sgmenda-aws:
### Description of changes:

Add experimental support for using AWS-LC HPKE as a `Cipher` using AWS-LC's `hpke.h`:
- update to building with an experimental branch of aws-lc
- add new `AlgorithmParameterSpec` `HpkeParameterSpec`
- add new `EvpKey` `EvpHpkeKey`
- add new `PublicKey` `EvpHpkePublicKey` and new `PrivateKey` `EvpHpkePrivateKey`
- add new `KeyPairGenerator` `HpkeGen` which generates a HPKE key, backed by the native `hpke_gen.cpp`
- add new `Cipher``HpkeCipher`, which performs single-shot HPKE encryption and decryption, backed by the native`hpke_cipher.cpp`.

### Call-outs:

- lines 40-209 in `HpkeCipher.java` contain the core java implementation
- native files `hpke_gen.cpp` and `hpke_cipher.cpp` contain the main cpp wrappers
- general code organization into classes and methods
- there are some outstanding TODO that will be addressed in a later revision

### Testing:

Added `HpkeGenTest` and `HpkeCipherTest` to test basic correctness.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "aws-lc"]
 	path = aws-lc
-	url = https://github.com/sgmenda-aws/aws-lc
-	branch = experimental-pq-hybrid-with-hpke
+	url = https://github.com/aws/aws-lc
+	branch = experimental-pq-hybrid
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "aws-lc"]
 	path = aws-lc
-	url = https://github.com/awslabs/aws-lc
+	url = https://github.com/sgmenda-aws/aws-lc
+	branch = experimental-pq-hybrid-with-hpke
--- a/build.gradle
+++ b/build.gradle
@@ -194,31 +194,11 @@ task buildAwsLc {
         }
 
         if (!isLegacyBuild) {
-            // FIXME: Change to tag once stabilized.
-//            exec {
-//                workingDir projectDir
-//                commandLine "git", "submodule", "sync", "--recursive"
-//            }
-//            exec {
-//                workingDir projectDir
-//                commandLine "git", "submodule", "foreach", "'git fetch --all'"
-//            }
-//            exec {
-//                workingDir projectDir
-//                commandLine "git", "submodule", "update", "--init", "--recursive", "--remote"
-//            }
-//            exec {
-//                workingDir awslcSrcPath
-//                commandLine "git", "fetch", "--all"
-//            }
-//            exec {
-//                workingDir awslcSrcPath
-//                commandLine "git", "checkout", awsLcGitVersionId
-//            }
-//            exec {
-//                workingDir awslcSrcPath
-//                commandLine "git", "pull"
-//            }
+            // FIXME: Commented because it was breaking Github CI
+            // exec {
+            //     workingDir awslcSrcPath
+            //     commandLine "git", "checkout", awsLcGitVersionId
+            // }
         }
         mkdir "${buildDir}/awslc"
         mkdir sharedObjectOutDir
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,8 @@ group = 'software.amazon.cryptools'
 version = '2.4.1'
 ext.isFips = Boolean.getBoolean('FIPS')
 if (ext.isFips) {
-    ext.awsLcGitVersionId = 'AWS-LC-FIPS-2.0.13'
+    // TODO: replace with tags once stable
+    ext.awsLcGitVersionId = '72c276e9c709a2d9b94e41b06da6abf2b3805a4a'
 } else {
     ext.awsLcGitVersionId = '72c276e9c709a2d9b94e41b06da6abf2b3805a4a'
 }
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,8 @@ ext.isFips = Boolean.getBoolean('FIPS')
 if (ext.isFips) {
     ext.awsLcGitVersionId = 'AWS-LC-FIPS-2.0.13'
 } else {
-    ext.awsLcGitVersionId = 'v1.30.1'
+    // FIXME: Change to tag once stabilized.
+    ext.awsLcGitVersionId = 'experimental-pq-hybrid-with-hpke'
 }
 
 // Check for user inputted git version ID.
@@ -195,7 +196,7 @@ task buildAwsLc {
         if (!isLegacyBuild) {
             exec {
                 workingDir awslcSrcPath
-                commandLine "git", "fetch", "--tags"
+                commandLine "git", "fetch", "--all"
             }
             exec {
                 workingDir awslcSrcPath
--- a/build.gradle
+++ b/build.gradle
@@ -201,6 +201,10 @@ task buildAwsLc {
             }
             exec {
                 workingDir projectDir
+                commandLine "git", "submodule", "foreach", "'git fetch --all'"
+            }
+            exec {
+                workingDir projectDir
                 commandLine "git", "submodule", "update", "--init", "--recursive", "--remote"
             }
             exec {
--- a/build.gradle
+++ b/build.gradle
@@ -196,6 +196,10 @@ task buildAwsLc {
         if (!isLegacyBuild) {
             // FIXME: Change to tag once stabilized.
             exec {
+                workingDir projectDir
+                commandLine "git", "submodule", "sync"
+            }
+            exec {
                 workingDir awslcSrcPath
                 commandLine "git", "fetch", "--all"
             }
--- a/build.gradle
+++ b/build.gradle
@@ -194,6 +194,7 @@ task buildAwsLc {
         }
 
         if (!isLegacyBuild) {
+            // FIXME: Change to tag once stabilized.
             exec {
                 workingDir awslcSrcPath
                 commandLine "git", "fetch", "--all"
@@ -201,6 +202,10 @@ task buildAwsLc {
             exec {
                 workingDir awslcSrcPath
                 commandLine "git", "checkout", awsLcGitVersionId
+            }
+            exec {
+                workingDir awslcSrcPath
+                commandLine "git", "pull"
             }
         }
         mkdir "${buildDir}/awslc"
--- a/build.gradle
+++ b/build.gradle
@@ -200,6 +200,10 @@ task buildAwsLc {
                 commandLine "git", "submodule", "sync"
             }
             exec {
+                workingDir projectDir
+                commandLine "git", "submodule", "update", "--init", "--remote"
+            }
+            exec {
                 workingDir awslcSrcPath
                 commandLine "git", "fetch", "--all"
             }
--- a/build.gradle
+++ b/build.gradle
@@ -12,15 +12,13 @@ plugins {
     id "io.github.gradle-nexus.publish-plugin" version "1.1.0"
 }
 
-ext.experimentalPqHybrid = true
-
 group = 'software.amazon.cryptools'
 version = '2.4.1'
 ext.isFips = Boolean.getBoolean('FIPS')
 if (ext.isFips) {
     ext.awsLcGitVersionId = 'AWS-LC-FIPS-2.0.13'
 } else {
-    ext.awsLcGitVersionId = 'v1.30.1'
+    ext.awsLcGitVersionId = '72c276e9c709a2d9b94e41b06da6abf2b3805a4a'
 }
 
 // Check for user inputted git version ID.
@@ -194,7 +192,7 @@ task buildAwsLc {
             }
         }
 
-        if (!isLegacyBuild && !experimentalPqHybrid) {
+        if (!isLegacyBuild) {
             exec {
                 workingDir awslcSrcPath
                 commandLine "git", "fetch", "--tags"
--- a/build.gradle
+++ b/build.gradle
@@ -195,30 +195,30 @@ task buildAwsLc {
 
         if (!isLegacyBuild) {
             // FIXME: Change to tag once stabilized.
-            exec {
-                workingDir projectDir
-                commandLine "git", "submodule", "sync", "--recursive"
-            }
-            exec {
-                workingDir projectDir
-                commandLine "git", "submodule", "foreach", "'git fetch --all'"
-            }
-            exec {
-                workingDir projectDir
-                commandLine "git", "submodule", "update", "--init", "--recursive", "--remote"
-            }
-            exec {
-                workingDir awslcSrcPath
-                commandLine "git", "fetch", "--all"
-            }
-            exec {
-                workingDir awslcSrcPath
-                commandLine "git", "checkout", awsLcGitVersionId
-            }
-            exec {
-                workingDir awslcSrcPath
-                commandLine "git", "pull"
-            }
+//            exec {
+//                workingDir projectDir
+//                commandLine "git", "submodule", "sync", "--recursive"
+//            }
+//            exec {
+//                workingDir projectDir
+//                commandLine "git", "submodule", "foreach", "'git fetch --all'"
+//            }
+//            exec {
+//                workingDir projectDir
+//                commandLine "git", "submodule", "update", "--init", "--recursive", "--remote"
+//            }
+//            exec {
+//                workingDir awslcSrcPath
+//                commandLine "git", "fetch", "--all"
+//            }
+//            exec {
+//                workingDir awslcSrcPath
+//                commandLine "git", "checkout", awsLcGitVersionId
+//            }
+//            exec {
+//                workingDir awslcSrcPath
+//                commandLine "git", "pull"
+//            }
         }
         mkdir "${buildDir}/awslc"
         mkdir sharedObjectOutDir
--- a/build.gradle
+++ b/build.gradle
@@ -12,14 +12,15 @@ plugins {
     id "io.github.gradle-nexus.publish-plugin" version "1.1.0"
 }
 
+ext.experimentalPqHybrid = true
+
 group = 'software.amazon.cryptools'
 version = '2.4.1'
 ext.isFips = Boolean.getBoolean('FIPS')
 if (ext.isFips) {
     ext.awsLcGitVersionId = 'AWS-LC-FIPS-2.0.13'
 } else {
-    // FIXME: Change to tag once stabilized.
-    ext.awsLcGitVersionId = 'experimental-pq-hybrid-with-hpke'
+    ext.awsLcGitVersionId = 'v1.30.1'
 }
 
 // Check for user inputted git version ID.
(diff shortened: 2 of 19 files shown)


csrc/buffer.h
```diff
@@ -402,6 +402,7 @@ class jni_borrow {
     }
 
     size_t len() const { return m_length; }
+    size_t size() const { return len(); }
```
geedo0, line 405: nit: I'd avoid adding aliases like this unless there's a compatibility reason. Otherwise, we end up proliferating different yet identical methods in different parts of the code base for no reason.

build.gradle
```diff
@@ -193,14 +194,11 @@ task buildAwsLc {
...
         if (!isLegacyBuild) {
-            exec {
-                workingDir awslcSrcPath
-                commandLine "git", "fetch", "--tags"
-            }
-            exec {
-                workingDir awslcSrcPath
-                commandLine "git", "checkout", awsLcGitVersionId
-            }
+            // FIXME: Commented because it was breaking Github CI
```
geedo0, line 197: What's going on here?

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -152,6 +152,13 @@ private void buildServiceMap() {
     }
 
     addSignatures();
+    addHPKE();
+  }
+
+  private void addHPKE() {
+    // Support EVP_HPKE as a Cipher
```
geedo0, line 159: nit: obvious comment

src/com/amazon/corretto/crypto/provider/HpkeGen.java
```diff
@@ -0,0 +1,49 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider;
+
+import java.security.*;
```
geedo0, line 5: Don't use wildcard imports. Ideally, just let the IDE import everything for you.

src/com/amazon/corretto/crypto/provider/HpkeParameterSpec.java
```diff
@@ -0,0 +1,70 @@
...
+  // Base mode, HPKE-PQT384, HKDF-SHA384, AES256-GCM
+  public static final HpkeParameterSpec Pqt384Sha384Aes256gcm =
+      new HpkeParameterSpec(0x00, 0xff05, 0x0002, 0x0002);
+
+  private final int mode;
+  private final int kemId;
+  private final int kdfId;
+  private final int aeadId;
+
+  public HpkeParameterSpec(int mode, int kemId, int kdfId, int aeadId) {
```
geedo0, line 48: If you declare this constructor private, that prevents others from foot-gunning themselves with custom parameter sets.

src/com/amazon/corretto/crypto/provider/HpkeGen.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  /** Generates a new HPKE key and returns a pointer to it. */
+  private static native long generateEvpHpkeKemKeyFromSpec(int hpke_kem_id);
+
+  @Override
+  public void initialize(final AlgorithmParameterSpec params, final SecureRandom rnd)
+      throws InvalidAlgorithmParameterException {
+    if (params instanceof HpkeParameterSpec) {
+      // TODO: do validation
```
geedo0, line 24: I don't think this is necessary if you declare the HpkeParameterSpec constructor as private.

src/com/amazon/corretto/crypto/provider/HpkeGen.java
```diff
@@ -0,0 +1,49 @@
...
+  public void initialize(final int keysize, final SecureRandom rnd)
+      throws InvalidParameterException {
+    throw new InvalidParameterException(
+        "Cannot initialize a KEM key with keysize, must use AlgorithmParameterSpec.");
+  }
+
+  @Override
+  public KeyPair generateKeyPair() {
+    if (spec == null) {
+      // TODO: support default spec?
```
geedo0, line 41: Interesting question. On the one hand, that would make this easier to use. On the other hand, there are consequences to getting the default wrong and having to live with it. I'd lean towards not setting a default unless we find a good reason to support it.

csrc/hpke_gen.cpp
```diff
@@ -0,0 +1,35 @@
...
+ * Signature: (I)J
+ */
+JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_HpkeGen_generateEvpHpkeKemKeyFromSpec(
+    JNIEnv* pEnv, jclass, jint hpke_kem_id)
+{
+    EVP_HPKE_KEY_auto key;
+    try {
+        raii_env env(pEnv);
+        key.set(EVP_HPKE_KEY_new());
+        const EVP_HPKE_KEM* kem = EVP_HPKE_KEM_find_by_id(hpke_kem_id);
```
geedo0, line 28: Assert that this was successful.

src/com/amazon/corretto/crypto/provider/EvpHpkePrivateKey.java
```diff
@@ -0,0 +1,30 @@
...
+
+  EvpHpkePrivateKey(InternalKey key, HpkeParameterSpec spec) {
+    super(key, false, spec);
+  }
+
+  EvpHpkePrivateKey(final long ptr, HpkeParameterSpec spec) {
+    this(new InternalKey(ptr), spec);
+  }
+
+  // Copied from EvpEcPrivateKey
```
geedo0, line 20: You can drop this comment, the provenance isn't necessary for something this small.

src/com/amazon/corretto/crypto/provider/HpkeParameterSpec.java
```diff
@@ -0,0 +1,70 @@
...
+  public static final HpkeParameterSpec Pqt256Sha256Aes128gcm =
+      new HpkeParameterSpec(0x00, 0xff04, 0x0001, 0x0001);
+  // Base mode, HPKE-PQT384, HKDF-SHA384, AES256-GCM
+  public static final HpkeParameterSpec Pqt384Sha384Aes256gcm =
+      new HpkeParameterSpec(0x00, 0xff05, 0x0002, 0x0002);
+
+  private final int mode;
+  private final int kemId;
+  private final int kdfId;
+  private final int aeadId;
```
geedo0, line 46: It'd be helpful to have some documentation around what these fields are. i.e. What they are exactly and how a future developer may find values for new ParameterSpec's.

src/com/amazon/corretto/crypto/provider/HpkeCipher.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+  }
+
+  // Unsupported Methods
+  // -------------------
+
+  @Override
+  protected void engineUpdateAAD(byte[] src, int offset, int len) {
+    // TODO: implement AAD support
+    throw new IllegalStateException("HpkeCipher currently does not support AAD");
```
geedo0, line 275: This limitation is a bit surprising since all the parameter sets we're supporting have an AEAD cipher. Is this because we ran out of time or is there a another reason for this limitation?

geedo0 (commented):
Still have to cover cipher implementation and testing.

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "aws-lc"]
 	path = aws-lc
-	url = https://github.com/awslabs/aws-lc
+	url = https://github.com/sgmenda-aws/aws-lc
+	branch = experimental-pq-hybrid-with-hpke
```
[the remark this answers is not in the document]
sgmenda-aws: Yes, this is temporary till https://github.com/aws/aws-lc/pull/1777 is merged.

csrc/buffer.h
sgmenda-aws: Oops, I agree, I meant to remove it before submitting the PR.

build.gradle
sgmenda-aws: Github does not support `git checkout <branch>` in submodules. This is temporary till aws-lc merges the HPKE PR, once that's done, I'll add this back and replace the VersionId with a commit hash.

Sorry for not explicitly marking these temporary changes.

src/com/amazon/corretto/crypto/provider/HpkeCipher.java
sgmenda-aws: I ran out of time and wanted to open a PR to get feedback on the general structure. I will fix this in a subsequent commit (before the PR is merged.)

src/com/amazon/corretto/crypto/provider/HpkeGen.java
sgmenda-aws: Oops, this was IntelliJ overriding me, updated my settings to prevent it from doing this.

src/com/amazon/corretto/crypto/provider/HpkeGen.java
sgmenda-aws: Good catch.

src/com/amazon/corretto/crypto/provider/HpkeGen.java
sgmenda-aws: That's fair, will update.

src/com/amazon/corretto/crypto/provider/HpkeCipher.java
```diff
@@ -0,0 +1,309 @@
...
+
+  // Core Java Methods
+  // -----------------
+
+  @Override
+  protected void engineInit(int opmode, Key key, AlgorithmParameterSpec params, SecureRandom random)
+      throws InvalidKeyException, InvalidAlgorithmParameterException {
+    if ((opmode != Cipher.WRAP_MODE) && (opmode != Cipher.UNWRAP_MODE)) {
+      throw new IllegalStateException("HpkeCipher only supports WRAP_MODE and UNWRAP_MODE");
+    }
```
geedo0, line 80: This limitation is not made clear anywhere other than the source code. Should be apparent in the PR, Javadoc, and README.

src/com/amazon/corretto/crypto/provider/HpkeCipher.java
```diff
@@ -0,0 +1,309 @@
...
+      params_ = (HpkeParameterSpec) params;
+      javaCipherMode_ = opmode;
+      key_ = opmode == Cipher.WRAP_MODE ? (EvpHpkePublicKey) key : (EvpHpkePrivateKey) key;
+    }
+  }
+
+  /**
+   * Performs HPKE single-shot encryption, as specified in Section 6.1 of RFC 9180.
+   *
+   * @return concatenation of KEM encapsulated key and encrypted ciphertext
```
geedo0, line 111: Is this concatenation format standardized or is it unique to your implementation? AFAICT, RFC 9180 simply refers to the encapsulated key and ciphertext as separate things. If it's unique to this implementation this should be documented more clearly somewhere and minimally the mirror logic should be described during unwrap.

src/com/amazon/corretto/crypto/provider/HpkeCipher.java
```diff
@@ -0,0 +1,309 @@
...
+    if (input == output) {
+      // TODO: is it okay if they are non-overlapping, support that case if necessary.
+      throw new IllegalStateException("input and output must be separate arrays");
+    }
+
+    synchronized (lock_) {
+      if (key_ == null) {
+        throw new IllegalStateException("key should be set before finalizing");
+      }
+      if ((javaCipherMode_ != Cipher.WRAP_MODE) && (javaCipherMode_ != Cipher.UNWRAP_MODE)) {
```
geedo0, line 170: You're able to encrypt/decrypt arbitrary data by setting the cipher to wrap mode and calling engineDoFinal directly. This seems like unintentional behavior. It can probably be avoided by using a boolean flag to indicate that the corresponding engineWrap method was invoked.

src/com/amazon/corretto/crypto/provider/HpkeCipher.java
```diff
@@ -0,0 +1,309 @@
...
+   *
+   * @return number of bytes written to output
+   */
+  @Override
+  protected int engineDoFinal(
+      byte[] input, int inputOffset, int inputLen, byte[] output, int outputOffset)
+      throws ShortBufferException {
+
+    if (input == output) {
+      // TODO: is it okay if they are non-overlapping, support that case if necessary.
```
geedo0, line 162: Check out the AES-GCM implementation, we go through a lot of trouble to support this. I'd rank this below all the other stuff like encrypt/decrypt and AAD though.

src/com/amazon/corretto/crypto/provider/HpkeParameterSpec.java
```diff
@@ -0,0 +1,70 @@
...
+  // Base mode, DHKEM-X25519, HKDF-SHA256, AES128-GCM
+  public static final HpkeParameterSpec X25519Sha256Aes128gcm =
+      new HpkeParameterSpec(0x00, 0x0020, 0x0001, 0x0001);
+  // Base mode, DHKEM-X25519, HKDF-SHA256, ChaCha20/Poly1305
+  public static final HpkeParameterSpec X25519Sha256Chapoly =
+      new HpkeParameterSpec(0x00, 0x0020, 0x0001, 0x0003);
+
+  // Selected PQ and PQ/T ciphersuites (experimental)
+
+  // Base mode, HPKE-MLKEM768, HKDF-SHA256, AES128-GCM
```
geedo0, line 20: Our PQ guidance suggests using AES256 minimum for PQ.

tst/com/amazon/corretto/crypto/provider/test/HpkeCipherTest.java
```diff
@@ -0,0 +1,104 @@
...
+    return KeyPairGenerator.getInstance("HPKE", TestUtil.NATIVE_PROVIDER);
+  }
+
+  private static Cipher getCipher() throws GeneralSecurityException {
+    return Cipher.getInstance("HPKE", TestUtil.NATIVE_PROVIDER);
+  }
+
+  @Test
+  public void testWrapUnwrap() throws GeneralSecurityException {
+    final HpkeParameterSpec[] namedSpecs = {
```
geedo0, line 39: Check out JUnit's built-in support for parameterized testing.

https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests

src/com/amazon/corretto/crypto/provider/HpkeCipher.java
sgmenda-aws: Good call out, I addressed this by:
1. moving the implementation from DoFinal to an internal method
2. having wrap and unwrap call that internal method
3. disallowing wrap and unwrap in DoFinal.

src/com/amazon/corretto/crypto/provider/HpkeCipher.java
sgmenda-aws: I ended up removing this limitation.

src/com/amazon/corretto/crypto/provider/HpkeCipher.java
sgmenda-aws: I added AAD support

src/com/amazon/corretto/crypto/provider/HpkeCipher.java
sgmenda-aws: This pattern of concatenating the encapsulated key and the ciphertext is a  well-know, see for instance: https://libsodium.gitbook.io/doc/public-key_cryptography/sealed_boxes#algorithm-details

I updated the javadocs to be more clear

src/com/amazon/corretto/crypto/provider/HpkeCipher.java
sgmenda-aws: Added support for overlapping buffers.

src/com/amazon/corretto/crypto/provider/HpkeCipher.java
```diff
@@ -92,56 +105,184 @@ protected void engineInit(int opmode, Key key, AlgorithmParameterSpec params, Se
...
+
+  /**
+   * Initializes a HPKE instance.
+   *
+   * @param opmode cipher mode, one of ENCRYPT, DECRYPT, WRAP, and UNWRAP
+   * @param key must be an instance of EvpHpkePublicKey for ENCRYPT and WRAP, and EvpHpkePrivateKey
+   *     for DECRYPT and UNWRAP
+   * @param params the algorithm parameters, must be an instance of HpkeParameterSpec, and may not
+   *     be null.
+   * @param random a source of randomness
```
geedo0, line 131: ...which we discard

geedo0 (approved):
This is looking pretty good. Only pending items I see are the placeholder AWS-LC references and failing CI. If we can get the CI passing, I'll happily approve.

sgmenda:
Turns out `-Xcheck:jni` does not like splitting one java array into multiple buffers, like splitting `output` into `enc` and `ct`. And, it was silently preventing writes to `ct`, making encrypt return an all-zero buffer. This was causing the failures.

Since `-Xcheck:jni` is not enabled on MacOS, I didn't notice it initially. And since the truncated logs don't make it explicit that this test was failing with `-Xcheck:jni` (and passing without.) I was extremely confused.

The following commit fixes this issue, and the test now passes with `-Xcheck:jni`: https://github.com/sgmenda/amazon-corretto-crypto-provider/commit/32eacb911d797e5352b3f8eb08271b543831f68d

geedo0:
Alright, just so I fully understand what's going on here. You understandably no longer have access to the original Github account so you pushed this fix to a separate fork/account?

Since you're no longer with us and this work is largely complete. The sensible next step is to cherry-pick this last commit to the branch and see how it runs against the tests. If we like the results enough, I'm inclined to merge the work into the experimental branch as-is since it's already largely complete. We can come back in later to do the necessary work to integrate it into the main trunk.

initsecret:
@geedo0 Precisely. I figured out the bug after so pushed a fix to my personal account. Yes, if this passes the tests, I was hoping you'd merge into the experimental branch. Alternatively, if you prefer, I could open a new PR, but it seems easier to apply this fix on top with `git am` like
```
wget -O fix.patch https://github.com/initsecret/amazon-corretto-crypto-provider/commit/32eacb911d797e5352b3f8eb08271b543831f68d.patch
git am fix.patch
git push
```

geedo0:
I've brought this to the team's attention and we are working on getting this PR integrated while being respectful of your personal time. The fact that we've locked you out of this account is problematic though. The idea we landed on is to have Amir fork your work and submit a new PR that includes your last change. We'll then test/review/iterate from there, but I've recommended that we simply bias towards merging it into the experimental branch as any further cleanup can be pulled into the greater effort of re-integrating this when the time comes.

PR #401, opened 2024-08-29
Add Benchmarks for Ed25519

sp717:
*Issue #, if available:*

*Description of changes:*
Adds new benchmarks for Ed25519 key pair generation, sign, and verify 

### DIFF
--- a/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/SignatureBase.java
+++ b/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/SignatureBase.java
@@ -16,15 +16,17 @@ public class SignatureBase {
   protected byte[] signature;
 
   protected void setup(
-      String provider,
-      String keyAlg,
-      AlgorithmParameterSpec keyParams,
-      String sigAlg,
-      AlgorithmParameterSpec sigParams)
-      throws Exception {
+          String provider,
+          String keyAlg,
+          AlgorithmParameterSpec keyParams,
+          String sigAlg,
+          AlgorithmParameterSpec sigParams)
+          throws Exception {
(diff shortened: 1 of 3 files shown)


benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/SignatureBase.java
```diff
@@ -38,7 +40,9 @@ protected void setup(
     signer.update(message);
     signature = signer.sign();
     verifier.update(message);
-    assert verifier.verify(signature);
```
WillChilds-Klein, line 41: note to other reviewers -- we found that this `assert` not only wasn't enforced in the benchmark build, but it also swallowed up checked exceptions from this call! so, we shouldn't use the `assert` keyword in the  benchmark.

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/SignatureBase.java
```diff
@@ -16,15 +16,17 @@ public class SignatureBase {
...
+          String provider,
+          String keyAlg,
+          AlgorithmParameterSpec keyParams,
+          String sigAlg,
+          AlgorithmParameterSpec sigParams)
+          throws Exception {
     BenchmarkUtils.setupProvider(provider);
     final KeyPairGenerator kpg = KeyPairGenerator.getInstance(keyAlg, provider);
-    kpg.initialize(keyParams);
+    if (!keyAlg.equals("Ed25519")) {
```
WillChilds-Klein, line 27: suggest adding a comment on the motivation for this guard

WillChilds-Klein (approved):

geedo0 (approved):

PR #402, opened 2024-09-04
Use environment variable for key store's password and minor updates

amirhosv:
*Description of changes:*

+ Use an environment variable for key store's password
+ Fix unit test's names
+ Update CHANGELOG

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -375,7 +375,7 @@ task build(overwrite:true) {
                 alias: jcecertAlias,
                 jar: "${buildDir}/cmake/AmazonCorrettoCryptoProvider.jar",
                 destDir: "${buildDir}/lib",
-                storepass: jcecertPassword,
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -708,7 +708,7 @@ task fakePublish {
 
 }
 
-if (project.hasProperty('mavenUser') && project.hasProperty('jcecertAlias')) {
```
WillChilds-Klein, line 711: don't we still need the mavenUser to publish?

amirhosv: We're using [environment variables](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/build.gradle#L770-L771) that specify a user token and password for publishing. The `gradle.properties` file that we'll use for releases will not have `mavenUser` defined in it anymore.

WillChilds-Klein (approved):

alexw91 (approved):

PR #403, opened 2024-09-09
Experimental FIPS build

amirhosv:
*Description of changes:*

Experimental FIPS builds allows one to experiment with non-FIPS branches/releases of AWS-LC and build it in FIPS mode.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,8 +41,14 @@ set(ENABLE_NATIVE_TEST_HOOKS NO CACHE BOOL "Enable debugging hooks in the RNG. D
 set(TEST_DATA_DIR ${PROJECT_SOURCE_DIR}/test-data/ CACHE STRING "Path to directory containing test data")
 set(ORIG_SRCROOT ${PROJECT_SOURCE_DIR} CACHE STRING "Path to root of original package")
 set(PROVIDER_VERSION_STRING "" CACHE STRING "X.Y.Z formatted version of the provider")
+set(EXPERIMENTAL_FIPS NO CACHE BOOL "Determines if this build is for FIPS mode with extra features from a non-FIPS branch of AWS-LC.")
 set(FIPS NO CACHE BOOL "Determine if this build is for FIPS mode")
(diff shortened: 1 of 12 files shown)


alexw91 (approved):

CMakeLists.txt
```diff
@@ -303,10 +312,8 @@ add_custom_command(
 add_custom_target(accp-jar-source DEPENDS ${ACCP_JAR_SOURCE})
 
 if(ENABLE_NATIVE_TEST_HOOKS)
-  add_executable(test_keyutils EXCLUDE_FROM_ALL
-        csrc/test_keyutils.cpp
-    )
-    link_with_openssl(test_keyutils)
```
WillChilds-Klein, line 309: why don't we need to link the test utils with AWS-LC anymore?

CMakeLists.txt
```diff
@@ -284,6 +290,9 @@ if(FIPS)
     set(TEST_FIPS_PROPERTY "-DFIPS=true")
 else()
     set(TEST_FIPS_PROPERTY "-DFIPS=false")
+endif()
+
+if(EXPERIMENTAL_FIPS OR (NOT FIPS))
```
WillChilds-Klein, line 295: should we drop a comment here that this guard can be relaxed once CKDF changes make it onto the FIPS branch version that we track?

amirhosv: [There is a similar comment on the Java side.](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java#L106). I'll add one comment here as well.

CMakeLists.txt
amirhosv: AWS-LC is statically linked into libaccp. AWS-LC has made some [changes](https://github.com/aws/aws-lc/pull/1332) to how the FIPS integrity checks works, and as a result, when AWS-LC is linked statically twice, on some platforms, like aarch64, it fails.

WillChilds-Klein (approved):

alexw91 (approved):

PR #408, opened 2024-10-17
Avoid unnecessary array copy in singlePass

pengxiaolong:
*Issue #, if available:* 407

*Description of changes:*

Both java_buffer::from_array and GetByteArrayRegion support consuming a slice of the array by passing offset and length, the change adds a `bufOffset` arguments to `fastDigest` to avoid unnecessary array copy in singlePass which needs to invoke `fastDigest` method.

We have tested the fix with JMH:

```
    private static ThreadLocal<MessageDigest> messageDigest = new ThreadLocal<MessageDigest>() {
        protected synchronized MessageDigest initialValue() {
            MessageDigest md = null;
            try {
                md = MessageDigest.getInstance("md5");
            } catch (NoSuchAlgorithmException e) {
                throw new RuntimeException("cannot find md5 algorithm", e);
            }
            return md;
        }
    };

    public static byte[][] inputs = new byte[][]{
            "1234567890123456789012345678901212345678901234567890123456789012".getBytes(), // 64 bytes
            "12345678901234567890123456789012".getBytes(),
            "12345678901234567890123456789012".getBytes(),
            "12345678901234567890123456789012".getBytes(),
            "12345678901234567890123456789012".getBytes(),
            "12345678901234567890123456789012".getBytes(),
            "12345678901234567890123456789012".getBytes(),
            "12345678901234567890123456789012".getBytes()
    };

    @Benchmark
    public static void test() {
        MessageDigest md = messageDigest.get();
        for(byte[] input : inputs) {
            md.digest(input);
        }
    }
```

JMH result:
W/o fix:
```
Benchmark                                  Mode  Cnt     Score    Error   Units
ACCPMD5Benchmark.test                      avgt    5  1630.452 ± 97.567   ns/op
ACCPMD5Benchmark.test:·gc.alloc.rate       avgt    5   346.316 ± 20.570  MB/sec
ACCPMD5Benchmark.test:·gc.alloc.rate.norm  avgt    5   592.000 ±  0.001    B/op
ACCPMD5Benchmark.test:·gc.count            avgt    5    11.000           counts
ACCPMD5Benchmark.test:·gc.time             avgt    5    11.000               ms

```

W/ fix:
```
Benchmark                                  Mode  Cnt     Score    Error   Units
ACCPMD5Benchmark.test                      avgt    5  1583.675 ± 94.346   ns/op
ACCPMD5Benchmark.test:·gc.alloc.rate       avgt    5   154.179 ±  9.225  MB/sec
ACCPMD5Benchmark.test:·gc.alloc.rate.norm  avgt    5   256.000 ±  0.001    B/op
ACCPMD5Benchmark.test:·gc.count            avgt    5     4.000           counts
ACCPMD5Benchmark.test:·gc.time             avgt    5     4.000               ms
```

### DIFF
--- a/template-src/com/amazon/corretto/crypto/provider/TemplateHashSpi.java
+++ b/template-src/com/amazon/corretto/crypto/provider/TemplateHashSpi.java
@@ -45,7 +45,7 @@ public final class TemplateHashSpi extends MessageDigestSpi implements Cloneable
      * @param buf Input buffer
      */
     // NOTE: This method trusts that all of the array lengths and bufLen are sane.
-    static native void fastDigest(byte[] digest, byte[] buf, int bufLen);
+    static native void fastDigest(byte[] digest, byte[] buf, int bufOffset, int bufLen);
 
     /**
      * @return The size of result hashes for this hash function
@@ -123,12 +123,8 @@ public final class TemplateHashSpi extends MessageDigestSpi implements Cloneable
     }
 
     private static byte[] singlePass(byte[] src, int offset, int length) {
-        if (offset != 0 || length != src.length) {
-            src = Arrays.copyOf(src, length);
-            offset = 0;
-        }
         final byte[] result = new byte[HASH_SIZE];
-        fastDigest(result, src, src.length);
+        fastDigest(result, src, offset, src.length);
         return result;
     }
 
--- a/template-src/com/amazon/corretto/crypto/provider/TemplateHashSpi.java
+++ b/template-src/com/amazon/corretto/crypto/provider/TemplateHashSpi.java
@@ -124,7 +124,7 @@ public final class TemplateHashSpi extends MessageDigestSpi implements Cloneable
 
     private static byte[] singlePass(byte[] src, int offset, int length) {
         final byte[] result = new byte[HASH_SIZE];
-        fastDigest(result, src, offset, src.length);
+        fastDigest(result, src, offset, length);
         return result;
     }
 
--- a/csrc/hash_template.cpp.template
+++ b/csrc/hash_template.cpp.template
@@ -130,7 +130,7 @@ JNIEXPORT void JNICALL JNI_NAME(updateNativeByteBuffer)(
 }
 
 JNIEXPORT void JNICALL JNI_NAME(fastDigest)(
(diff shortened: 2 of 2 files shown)


amirhosv (approved):
Thanks for catching this.

template-src/com/amazon/corretto/crypto/provider/TemplateHashSpi.java
```diff
@@ -45,7 +45,7 @@ public final class TemplateHashSpi extends MessageDigestSpi implements Cloneable
      * @param buf Input buffer
      */
     // NOTE: This method trusts that all of the array lengths and bufLen are sane.
-    static native void fastDigest(byte[] digest, byte[] buf, int bufLen);
+    static native void fastDigest(byte[] digest, byte[] buf, int bufOffset, int bufLen);
```
geedo0, line 48: Can you explain why supporting arbitrary offsets was necessary? I don't see the code path where that ever becomes non-0. The garbage seems to be generated entirely by the length checking logic.

https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/InputBuffer.java#L299

pengxiaolong: It is not necessary, but the idea to keep it arbitrary is from impl of singlePass which is called by InputBuffer:
```
        if (offset != 0 || length != src.length) {
            src = Arrays.copyOf(src, length);
            offset = 0;
        }
```

Although the code above has bug, correct code should be `src = Arrays.copyOfRange(src, offset, offset + length);
`

geedo0: Ack, thanks for explaining. A reduction in the code footprint, removing buggy logic, and a %-increase in throughput sounds like a win-win-win to me.

geedo0 (approved):

template-src/com/amazon/corretto/crypto/provider/TemplateHashSpi.java
pengxiaolong: Thanks!

PR #410, opened 2024-11-08
Make Ed25519 KeyFactory registration Opt-in

amirhosv:
*Description of changes:*

* The keys generated by KeyFactory of ACCP for Ed25519 do not implement EdEcKey interface from JCA since this interface is not present in JDKs prior to 15.
* Once ACCP supports multi-jar release, then this option can be removed and the KeyFactory for Ed25519 can always be registered.


src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -91,8 +93,15 @@ private void buildServiceMap() {
     addService("KeyFactory", "EC", "EvpKeyFactory$EC");
 
     if (shouldRegisterEdDSA) {
-      addService("KeyFactory", "EdDSA", "EvpKeyFactory$EdDSA");
-      addService("KeyFactory", "Ed25519", "EvpKeyFactory$EdDSA");
+      // KeyFactories are used to convert key encodings to Java Key objects. ACCP's KeyFactory for
+      // Ed25519 returns keys of type EvpEdPublicKey and EvpEdPrivateKey that do not implement
+      // EdECKey interface. One should register the KeyFactories from ACCP if they only use the
+      // output of the factories with ACCP's services.
+      // Once ACCP supports multi-release jar, then this option can be removed.
```
geedo0, line 100: Can you elaborate on what you mean here?

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -745,4 +757,12 @@ EvpKey translateKey(Key key, EvpKeyType keyType) throws InvalidKeyException {
       return (EvpKey) getKeyFactory(keyType).translateKey(key);
     }
   }
+
+  // In case the user does not register Ed25519 KeyFactories by ACCP, we still need one to be used
+  // internally.
+  private static class EdKeyFactory extends KeyFactory {
```
geedo0, line 763: I don't think this belongs inside the AmazonCorrettoCryptoProvider class, way too specific to a specific algorithm to be here. Maybe make it a standalone class or move it to the EvpKeyFactory class instead?

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -91,8 +93,15 @@ private void buildServiceMap() {
...
 
     if (shouldRegisterEdDSA) {
-      addService("KeyFactory", "EdDSA", "EvpKeyFactory$EdDSA");
-      addService("KeyFactory", "Ed25519", "EvpKeyFactory$EdDSA");
+      // KeyFactories are used to convert key encodings to Java Key objects. ACCP's KeyFactory for
+      // Ed25519 returns keys of type EvpEdPublicKey and EvpEdPrivateKey that do not implement
+      // EdECKey interface. One should register the KeyFactories from ACCP if they only use the
+      // output of the factories with ACCP's services.
+      // Once ACCP supports multi-release jar, then this option can be removed.
+      if (shouldRegisterEdKeyFactory) {
```
WillChilds-Klein, line 101: instead of making the consumers think about configuration, can we check java version here and only register if > 15?

geedo0: Nah, that's not enough. If a customer has an application, that needs their key factory to produce Sun JCE keys, they can't use our implementation without changing their code.

amirhosv: Ideally, we only want to register Ed25519 related services for JDKs 15+. However, we don't have a systematic way to implement a class that depends on some types that are not available in JDK 8. The systematic solution is to use [multi-release jars](https://openjdk.org/jeps/238).

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
amirhosv: KeyFactory is not intended to be instantiated explicitly: one should get an instance by calling the static factory methods KeyFactory::getInstance.
This usage is not common and `EdKeyFactory` is only used by [translateKey](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java#L741). I think it's better to be kept in this class.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
amirhosv: As the comment says, the EvpEdPublicKey (and EvpEdPrivateKey) does not implement [EdECKey](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/security/interfaces/EdECKey.html) interfaces, but technically, they should. This interface is not available in JDKs prior to 15. We need change ACCP's build so that on JDK 15+, the EvpEdPublicKey would implement these interfaces. The systematic solution for that is multi-release jar.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
geedo0: I see, so we'd release a multi-release jar that exposes a SunJCE compatible implementation of Ed25519 for JDK15+. And on lower JDKs we'd remove support? Or would we still be able to provide the kind of support we have now?

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
geedo0: If we could do it all over again, I'd avoid having translateKey in this class as well. Way too much stuff inside here right now and it'd all benefit from a good refactor. Won't block the PR over it though.

geedo0 (approved):

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
amirhosv: We can provide the same support as of now. The idea behind multi-release jar is that the same Java class can be implemented in multiple files. Each file is an independent implementation and at runtime, depending on the JDK version, the correct one would be loaded.

alexw91: Are we sure that we need multi-release jars? Is there a way to do this with Java reflection? 

I suspect that there is a way to at runtime check if the class `java.security.interfaces.EdECPublicKey` exists (if it does then ACCP is on JDK >= 15), and instantiate a [Reflection Proxy](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/reflect/Proxy.html) and [InvocationHandler](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/reflect/InvocationHandler.html) that implements that interface at runtime. It's hacky, but I think doable.

WillChilds-Klein: makes sense.

WillChilds-Klein (approved):

alexw91 (approved):

PR #412, opened 2024-11-11
Fix ByteBuffer offset calculation for AES/CBC/NoPadding

bmathiske:
*411*

The shim buffer implementation was missing the array offset into its source byte buffer, which needs to be added to the source byte buffer's position to obtain the correct effective position in the underlying byte array.
The added tests checks that this works for ByteBuffer slices with aligned and unaligned offsets.

This contribution was made possible by Guillaume Rose and Aline Bousquet at Apple. They found the bug and created a reproducer program, on which the submitted unit tests are based.


geedo0 (approved):
Great find, thanks for the contribution!

amirhosv (approved):

PR #414, opened 2024-12-23
Remove CPP coverage target from default coverage target

geedo0:
*Issue #, if available:*
ACCP-120 

*Description of changes:*
Mainly, we want to remove this to unblock our CI. This task is currently
breaking all Mac CI dimensions (and possibly others). The root cause
seems related to some mismatches in the related software dependencies
used to create the coverage report.

More broadly, this coverage report is not in a state of good health and
should not bere-introduced to the CI until we take the time addressing
these issues: Redundant unit test execution due to dependency hierarchy,
ignored errors in the coverage report resulting in poor coverage of the
coverage data, surfacing the coverage reporting in a useful manner.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -601,7 +601,7 @@ task coverage_cpp_report {
 }
 
 task coverage {
-    dependsOn coverage_java_report, coverage_cpp_report
+    dependsOn coverage_java_report
 }
 
 task release {


build.gradle
```diff
@@ -601,7 +601,7 @@ task coverage_cpp_report {
 }
 
 task coverage {
-    dependsOn coverage_java_report, coverage_cpp_report
+    dependsOn coverage_java_report
```
WillChilds-Klein, line 604: should we just remove the `coverage` target from `release`? do we actually use coverage data in any non-local-dev workflow (CI, imports, internal pipelines, etc.)?

geedo0: FWIW, the Java coverage report does work and is reported correctly into CodeBuild. Basically, it doesn't have any of the issues related to that cpp coverage target. I don't see a reason to drop it.

Also, removing from release means you have to make changes in at least one other build-script that doesn't use the release target.

WillChilds-Klein (approved):

brian-jarvis-aws (approved):

PR #416, opened 2024-12-26
JMH Benchmarking Fixups

geedo0:
*Issue #, if available:*
ACCP-118

*Description of changes:*
These changes intend to improve the consistency and freshness of the JMH
benchmarks to support better reporting.

- Upgrade the BouncyCastle dependency to the latest `jdk18on` artifact.
  The artifact we were previously using has not seen an update in 3
  years.
- Remove the ms TimeUnit override for AES-XTS. This is the only
  benchmark that does this and breaks uniformity across the suite.
- Update Random's TimeUnit to microseconds. Nanosecond precision is not
  supported by CloudWatch and the ns precision is not strictly
  necessary.
- Fix some misnamed benchmark names.

### DIFF
--- a/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/Random.java
+++ b/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/Random.java
@@ -25,7 +25,7 @@ import org.openjdk.jmh.annotations.Threads;
  * mode (ops/s), because the throughput mode sums the number of operations over all the threads
  */
 @BenchmarkMode(Mode.AverageTime)
-@OutputTimeUnit(TimeUnit.NANOSECONDS)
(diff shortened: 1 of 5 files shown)


brian-jarvis-aws (approved):

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/Random.java
```diff
@@ -25,7 +25,7 @@
  * mode (ops/s), because the throughput mode sums the number of operations over all the threads
  */
 @BenchmarkMode(Mode.AverageTime)
-@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
```
WillChilds-Klein, line 28: can this be configured centrally so it's the same across all benchmarks?

WillChilds-Klein (approved):

PR #422, opened 2025-01-28
Support "pure" ML-DSA

WillChilds-Klein:
*Issue #, if available:* i/ACCP-100

*Description of changes:*

# Overview

This PR exposes ML-DSA functionality through the Signature, KeyPairGenerator, and KeyFactory JCA interfaces. The implementation closely follows [our Ed25519 implementation](https://github.com/corretto/amazon-corretto-crypto-provider/commit/0bfd4f278f07a7ee4f1b82070b650af5c0958279#diff-797ef5bd8c859f6e598eb60150c9cf5aae1b21f02b968288beb4be766f0b28aaR344), as both algorithms are currently used through AWS-LC's `EVP_DigestSign` API initialized with a `NULL` message digest (these newer algorithms use a fixed, non-parameterizable digest function per spec).

We also provide benchmarks for key generation, signing, and verifying.

# Interoperability

This change has been written to be interoperable and forwards-compatible with other JCA provider ML-DSA implementations, such as [BouncyCastle's](https://github.com/bcgit/bc-java/tree/8b813e11536eddea8a2efc370ee7db0ddadcd97c/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/mldsa) and OpenJDK's [JEP 497](https://openjdk.org/jeps/497).

This means that ACCP callers cannot use JDK's [NamedParameterSpec classes](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/security/spec/NamedParameterSpec.java#L72-L94) or BouncyCastle's [MLDSAParameterSpec](https://github.com/bcgit/bc-java/blob/main/prov/src/main/java/org/bouncycastle/jcajce/spec/MLDSAParameterSpec.java), but allows ACCP to provide ML-DSA on supported LTS JDK versions, independent of other providers. Instead of using ParameterSpec classes, callers can request a specific security level via the algorithm name passed into `KeyPairGenerator.getInstance` (if no level is specified, `"ML-DSA-44"` is the default):

```
KeyPairGenerator g = KeyPairGenerator.getInstance("ML-DSA-87");
KeyPair kp = g.generateKeyPair(); // an ML-DSA-87 key pair

g = KeyPairGenerator.getInstance("ML-DSA");
kp = g.generateKeyPair(); // an ML-DSA-44 key pair
```

We also update our BouncyCastle test dependency to 1.80, allowing us to pefrorm interoperability tests with BouncyCastle's ML-DSA implementation.

## Limitations

As mentioned above, ACCP does not currently support [ParameterSpec-based initialization](https://docs.oracle.com/javase/8/docs/api/java/security/KeyPairGenerator.html#initialize-java.security.spec.AlgorithmParameterSpec-java.security.SecureRandom-) on `KeyPairGenerator` instances. We also don't support `keysize`-based initialization as it is somewhat ambiguous for ML-DSA. These limitations may change in the future.

ACCP's ML-DSA keys cannot be used directly with BouncyCastle [KeyFactory](https://github.com/bcgit/bc-java/blob/8b813e11536eddea8a2efc370ee7db0ddadcd97c/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/mldsa/MLDSAKeyFactorySpi.java#L112) or [Signature](https://github.com/bcgit/bc-java/blob/8b813e11536eddea8a2efc370ee7db0ddadcd97c/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/mldsa/SignatureSpi.java#L83) instances, as BouncyCastle only supports its own keys today. To use ACCP keys with BouncyCastle, callers will need to serialize them into `X509EncodedKeySpec` or `PKCS8EncodedKeySpec` instances and pass those to BouncyCastle's `KeyFactory` for conversion.

Finally, ACCP cannot yet deserialize ML-DSA private keys serialized by BouncyCastle. BouncyCastle currently encodes the private key's 32 byte seed upon serialization, while ACCP expects the fully expanded private key material.

# Future Work

- Migrate to streaming APIs in AWS-LC (we currently buffer entire message in memory)
- Support deserializing private key seeds (e.g. from BouncyCastle) and expanding them (this will require AWS-LC supporting ibid., we're actively working on this)
- Add ML-DSA to EvpSignatureTest suite (unblocked once we can parse BouncyCastle keys per above)
- Interoperability tests with JDK 24 (once we support that version)

### DIFF
--- a/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/KeyGenMlDSA.java
+++ b/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/KeyGenMlDSA.java
@@ -1,0 +1,36 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider.benchmarks;
+
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+
+import com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+
+@State(Scope.Benchmark)
+public class KeyGenMlDSA {
+
+    @Param({AmazonCorrettoCryptoProvider.PROVIDER_NAME, "BC"})
+    public String provider;
+
+    @Param({"ML-DSA-44", "ML-DSA-65", "ML-DSA-87"})
+    public String algo;
+
+    private KeyPairGenerator kpg;
+
+    @Setup
+    public void setup() throws Exception {
+        BenchmarkUtils.setupProvider(provider);
+        kpg = KeyPairGenerator.getInstance(algo, provider);
+    }
+
+    @Benchmark
+    public KeyPair generate() {
+        return kpg.generateKeyPair();
+    }
+}
--- a/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/SignatureMlDSA.java
+++ b/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/SignatureMlDSA.java
@@ -1,0 +1,34 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider.benchmarks;
+
+import com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+
+@State(Scope.Benchmark)
+public class SignatureMlDSA extends SignatureBase {
+    @Param({AmazonCorrettoCryptoProvider.PROVIDER_NAME, "BC"})
+    public String provider;
+
+    @Param({"ML-DSA-44", "ML-DSA-65", "ML-DSA-87"})
+    public String algo;
+
+    @Setup
+    public void setup() throws Exception {
+        super.setup(provider, algo, null, "ML-DSA", null);
+    }
+
+    @Benchmark
+    public byte[] sign() throws Exception {
+        return super.sign();
+    }
+
+    @Benchmark
+    public boolean verify() throws Exception {
+        return super.verify();
+    }
+}
--- a/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
+++ b/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
@@ -93,7 +93,6 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
     addService("KeyFactory", "RSA", "EvpKeyFactory$RSA");
     addService("KeyFactory", "EC", "EvpKeyFactory$EC");
 
-
     if (shouldRegisterMLDSA) {
       addService("KeyFactory", "ML-DSA", "EvpKeyFactory$MLDSA");
       addService("KeyFactory", "ML-DSA-44", "EvpKeyFactory$MLDSA");
--- a/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
+++ b/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
@@ -91,10 +91,13 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
 
     addService("KeyFactory", "RSA", "EvpKeyFactory$RSA");
     addService("KeyFactory", "EC", "EvpKeyFactory$EC");
-    addService("KeyFactory", "ML-DSA", "EvpKeyFactory$MlDSA");
-    addService("KeyFactory", "ML-DSA-44", "EvpKeyFactory$MlDSA");
-    addService("KeyFactory", "ML-DSA-65", "EvpKeyFactory$MlDSA");
-    addService("KeyFactory", "ML-DSA-87", "EvpKeyFactory$MlDSA");
+
+    if (!isFips() || isExperimentalFips()) {
+      addService("KeyFactory", "ML-DSA", "EvpKeyFactory$MlDSA");
+      addService("KeyFactory", "ML-DSA-44", "EvpKeyFactory$MlDSA");
+      addService("KeyFactory", "ML-DSA-65", "EvpKeyFactory$MlDSA");
+      addService("KeyFactory", "ML-DSA-87", "EvpKeyFactory$MlDSA");
+    }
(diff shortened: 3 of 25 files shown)


src/com/amazon/corretto/crypto/provider/MlDsaGen.java
```diff
@@ -0,0 +1,63 @@
...
+import java.security.KeyPairGeneratorSpi;
+import java.security.SecureRandom;
+import java.security.spec.AlgorithmParameterSpec;
+
+class MlDsaGen extends KeyPairGeneratorSpi {
+  /** Generates a new ML-DSA key and returns a pointer to it. */
+  private static native long generateEvpMlDsaKey(int nid);
+
+  private final AmazonCorrettoCryptoProvider provider_;
+  private int type_ = -1;
```
geedo0, line 15: The nomenclature around this value is all over the place. I see the names `type`, `nid`, and `level` to refer to this integer. And I can't really work out why we're using the values 2, 3, and 5 for it.

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/SignatureMlDSA.java
```diff
@@ -0,0 +1,34 @@
...
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+
+@State(Scope.Benchmark)
+public class SignatureMlDSA extends SignatureBase {
+    @Param({AmazonCorrettoCryptoProvider.PROVIDER_NAME, "BC"})
+    public String provider;
+
+    @Param({"ML-DSA-44", "ML-DSA-65", "ML-DSA-87"})
+    public String algo;
```
geedo0, line 18: It's important for us to be more consistent with the parameter names because they're used directly as CloudWatch metric dimensions. Can you expand this to `algorithm` to align with other benchmarks?

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/KeyGenMlDSA.java
```diff
@@ -0,0 +1,36 @@
...
+import org.openjdk.jmh.annotations.State;
+
+@State(Scope.Benchmark)
+public class KeyGenMlDSA {
+
+    @Param({AmazonCorrettoCryptoProvider.PROVIDER_NAME, "BC"})
+    public String provider;
+
+    @Param({"ML-DSA-44", "ML-DSA-65", "ML-DSA-87"})
+    public String algo;
```
geedo0, line 22: `algorithm`

tst/com/amazon/corretto/crypto/provider/test/KeyReuseThreadStormTest.java
```diff
@@ -235,6 +241,35 @@ public void edThreadStorm() throws Throwable {
     executeThreads(threads);
   }
 
+  @Test
+  public void mlDsaThreadStorm() throws Throwable {
+    final byte[] rngSeed = TestUtil.getRandomBytes(20);
+    System.out.println("RNG Seed: " + Arrays.toString(rngSeed));
```
geedo0, line 247: Aside from consistency with other tests, how is this useful?

geedo0 (commented):
Minor comments, but would like to get those cleaned up before a merge.

Not that I necessarily agree with the stance, but lately we've been including complete KAT suites for every new algorithm to the library.

src/com/amazon/corretto/crypto/provider/EvpKeyType.java
```diff
@@ -18,7 +18,8 @@
 enum EvpKeyType {
   RSA("RSA", 6, RSAPublicKey.class, RSAPrivateKey.class),
   EC("EC", 408, ECPublicKey.class, ECPrivateKey.class),
-  EdDSA("EdDSA", 949, PublicKey.class, PrivateKey.class);
+  EdDSA("EdDSA", 949, PublicKey.class, PrivateKey.class),
+  MlDSA("ML-DSA", 993, PublicKey.class, PrivateKey.class);
```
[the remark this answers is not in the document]
WillChilds-Klein: ah, yeah, i guess "**M**odule **L**attice" is two words where "**Ed**wards" is one word. will change.

src/com/amazon/corretto/crypto/provider/MlDsaGen.java
WillChilds-Klein: fair. will standardize this on `level`, which refers to "claimed NIST level" [here](https://openquantumsafe.org/liboqs/algorithms/sig/ml-dsa.html).

tst/com/amazon/corretto/crypto/provider/test/KeyReuseThreadStormTest.java
WillChilds-Klein: This test, implemented for each PKEY type we wrap, provides some extra assurance that our usage of PKEYs is thread-safe.

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/SignatureMlDSA.java
WillChilds-Klein: will do.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -92,6 +92,13 @@ private void buildServiceMap() {
     addService("KeyFactory", "RSA", "EvpKeyFactory$RSA");
     addService("KeyFactory", "EC", "EvpKeyFactory$EC");
 
+    if (!isFips() || isExperimentalFips()) {
```
alexw91, line 95: Nit: store this in a variable `boolean shouldRegisterMLDSA = ( !isFips() || isExperimentalFips() );`, and then use the variable throughout this file.

WillChilds-Klein: sure, makes sense.

tst/com/amazon/corretto/crypto/provider/test/KeyReuseThreadStormTest.java
WillChilds-Klein: Oh, you mean printing the seed? That would allow us to deterministically reproduce any issues this test catches, which would be ~impossible without the seed.

src/com/amazon/corretto/crypto/provider/MlDsaGen.java
```diff
@@ -12,12 +12,16 @@ class MlDsaGen extends KeyPairGeneratorSpi {
   private static native long generateEvpMlDsaKey(int nid);
```
geedo0, line 12: nit: this should also be level

geedo0 (dismissed):

src/com/amazon/corretto/crypto/provider/MlDsaGen.java
WillChilds-Klein: yes.

geedo0 (approved):

geedo0 (approved):

WillChilds-Klein:
@geedo0 

> lately we've been including complete KAT suites for every new algorithm to the library.

I would've included some KATs if it were straightforward. However, because ML-DSA introduces distinct randomness into ever signature, signatures with identical parameters are unique. Since ACCP/JCA doesn't have a clean way of exposing AWS-LC's ML-DSA derandomization APIs, KATs for this PR are currently infeasible.

alexw91 (approved):

geedo0 (approved):

alexw91 (approved):

PR #423, opened 2025-01-31
Support "External Mu" ML-DSA

WillChilds-Klein:
*Issue #, if available:* i/ACCP-128

*Description of changes:*

# Overview

This PR adds support for the External Mu variant of ML-DSA described in [AWS-LC PR 2113](https://github.com/aws/aws-lc/pull/2113). To facilitate calculation of mu in testing and for consumers, we expose a `computeMLDSAMu` helper function in a new `PublicUtils` class. Notably, External Mu ML-DSA is interoperable with Pure ML-DSA across signing and verification, as shown in tests.

# Interoperability

We expose External Mu as a distinct `Signature` service to preserve forwards-compatibility with other providers' potential future implementations. Conventionally, this "prehash" mode would be configured with a call to [`setParameter`](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/security/Signature.html#setParameter(java.security.spec.AlgorithmParameterSpec)), passing an `AlgorithmParameterSpec` implementation [as in EdDSA](https://download.java.net/java/early_access/panama/docs/api/java.base/java/security/spec/EdDSAParameterSpec.html). Because no other providers have yet created an `AlgorithmParameterSpec` implementation that is aware of External Mu, this approach would have required us to expose an ACCP-specific `AlgorithmParameterSpec` that callers would need to use, hampering interoperability.

---


src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -231,6 +231,7 @@ private void addSignatures() {
 
     if (shouldRegisterMLDSA) {
       addService("Signature", "ML-DSA", "EvpSignatureRaw$MLDSA");
+      addService("Signature", "ML-DSA-Ext-Mu", "EvpSignatureRaw$MLDSAExtMu");
```
WillChilds-Klein, line 234: There's currently no JCA-standard name for this (yet?), so we have some latitude in choosing the algorithm name. I'm not sold on `ML-DSA-Ext-Mu`, open to suggestions from reviewers.

jakemas: In [draft-ietf-lamps-dilithium-certificates](https://datatracker.ietf.org/doc/draft-ietf-lamps-dilithium-certificates/) we use `ExternalMu-ML-DSA`. In [ACVP ML-DSA JSON Specification](https://pages.nist.gov/ACVP/draft-celi-acvp-ml-dsa.html) it is just referred to as `externalmu`.

Notably, theres no hyphen between `external` and `mu` in both.

We chose `ExternalMu-ML-DSA` in the IETF to hint at consistency with NISTs notion of prefixing the mode name distinction: i.e.,`HashML-DSA`.

As such `ML-DSA-ExtMu`, `ML-DSA-ExternalMu` are good options. If you like  `ML-DSA-Ext-Mu`, `ML-DSA-External-Mu` with the extra hyphen, that's also fine!

src/com/amazon/corretto/crypto/provider/PublicUtils.java
```diff
@@ -0,0 +1,28 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider;
+
+import java.security.PublicKey;
+
+/** Public utility methods */
+public class PublicUtils {
```
geedo0, line 8: Is the intent here to begin developing/exposing ACCP exclusive utility classes? The only use is within test code. Is there an intended customer for this interface (let's continue this convo off of Github)? I am a little concerned about how specific this is to ML-DSA and how we might pigeon-hole ourselves with this class/package hierarchy if we decide to add a lot more utility functions in the future. The potential for a 1000+ line monster class is here.

WillChilds-Klein: There has been some interest from internal teams to use this in testing, but we can hold off on exposing it until there's a clearly articulated need. I've moved the mu helper function to TestUtil.

geedo0 (dismissed):

geedo0 (approved):

alexw91 (approved):

Issue #377, opened 2024-04-10
Populate the security Provider "info" string


jeking3:
Currently getInfo() returns an empty string.  It could return info about the name, version, and FIPS compliance.

https://github.com/corretto/amazon-corretto-crypto-provider/blob/32fce7ab1077817dd4e36ba0fb6a9fe9cecee679/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java#L384C5-L384C65

For comparison, see Bouncy Castle:

https://github.com/bcgit/bc-java/blob/5b1360854d85fd27b75720015be68f9e172db013/prov/src/main/jdk1.1/org/bouncycastle/jce/provider/BouncyCastleProvider.java#L46

geedo0:
I'm open to the idea, but what is your use case? If we were to do this, it shouldn't be used for any kind of assertion. Perhaps just for logging the provider info?

[JDK Doc](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/security/Provider.html#getInfo())

jeking3:
The use case is to align it with other crypto providers and the Provider interface.  The constructor allows for an info string to be provided.  There is a getInfo() method.  I cannot do this reliably because ACCP passes an empty string in for the info:
```
   log.info("Preferred crypto provider: " + preferredProvider.getInfo());
```

Something like this would be great:
```
    AmazonCorrettoCryptoProvider v2.3.3 [FIPS]
```

Where FIPS is optional, and there may be other optional configuration things that could go in a list here in the future.

jeking3:
Here's what you get for the name | info from each provider today (OpenJDK8 + ACCP injected by security properties)
```
23 Apr 2024 14:33:48,579 Security providers at startup:
23 Apr 2024 14:33:48,579 Slot #1: AmazonCorrettoCryptoProvider | 
23 Apr 2024 14:33:48,579 Slot #2: SUN | SUN (DSA key/parameter generation; DSA signing; SHA-1, MD5 digests; SecureRandom; X.509 certificates; JKS & DKS keystores; PKIX CertPathValidator; PKIX CertPathBuilder; LDAP, Collection CertStores, JavaPolicy Policy; JavaLoginConfig Configuration)
23 Apr 2024 14:33:48,579 Slot #3: SunRsaSign | Sun RSA signature provider
[...]

PR #427, opened 2025-02-04
Add provider info string

chockalingamc:
*Issue #* [377](https://github.com/corretto/amazon-corretto-crypto-provider/issues/377)

*Description of changes:*
Adding a provider info string to AmazonCorrettoCryptoProvider using [Semantic Versioning](https://semver.org) format, with additional info on the underlying AWS-LC version.

Example output for the provider info string

* **AmazonCorrettoCryptoProvider 2.5.0 (AWS-LC v1.44.0)**
Normal build with a 'main' branch version of AWS-LC

* **AmazonCorrettoCryptoProvider 2.5.0+FIPS (AWS-LC-FIPS-3.0.0)**
Built in FIPS mode with `AWS-LC-FIPS-x.y.z` version of AWS-LC

* **AmazonCorrettoCryptoProvider 2.5.0+FIPS+EXP (AWS-LC v1.44.0)**
Built in FIPS mode with a 'main' branch version of AWS-LC (ie: with EXPERIMENTAL_FIPS flag)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,11 +15,7 @@ plugins {
 group = 'software.amazon.cryptools'
 version = '2.5.0'
 ext.isExperimentalFips = Boolean.getBoolean('EXPERIMENTAL_FIPS')
-if (ext.isExperimentalFips) {
-    ext.isFips = true
-} else {
-    ext.isFips = Boolean.getBoolean('FIPS')
-}
+ext.isFips = ext.isExperimentalFips || Boolean.getBoolean('FIPS')
 
 if (ext.isExperimentalFips || !ext.isFips) {
     // Experimental FIPS uses the same AWS-LC version as non-FIPS builds.
@@ -336,6 +332,7 @@ task executeCmake(type: Exec) {
     args "-DOPENSSL_ROOT_DIR=${buildDir}/awslc/bin", '-DCMAKE_BUILD_TYPE=Release', '-DPROVIDER_VERSION_STRING=' + version
     args "-DTEST_RUNNER_JAR=${configurations.testRunner.singleFile}"
     args "-DCMAKE_VERBOSE_MAKEFILE:BOOL=ON"
+    args "-DAWS_LC_VERSION_STRING=" + awsLcGitVersionId
     if (isFips) {
         args "-DFIPS=ON"
     }
@@ -486,6 +483,12 @@ task single_test(type: Exec) {
     commandLine 'make', 'check-junit-single'
 }
 
+task test_install_via_properties(type: Exec) {
+    dependsOn executeCmake
+    workingDir "${buildDir}/cmake"
+    commandLine 'make', 'check-install-via-properties'
+}
+
 task test_integration_exec(type: Exec) {
     dependsOn executeCmake
     workingDir "${buildDir}/cmake"
@@ -531,6 +534,7 @@ task coverage_cmake(type: Exec) {
     args '-DPROVIDER_VERSION_STRING=' + version, projectDir
     args "-DTEST_RUNNER_JAR=${configurations.testRunner.singleFile}"
     args "-DCMAKE_VERBOSE_MAKEFILE:BOOL=ON"
+    args "-DAWS_LC_VERSION_STRING=" + awsLcGitVersionId
     if (isLegacyBuild) {
         args "-DLEGACY_BUILD=ON"
     }
@@ -883,11 +887,6 @@ task generateEclipseClasspath {
                     }
                 }
                 classpathentry('kind': 'src', 'output': 'build/eclipse/src', 'path': 'build/cmake/generated-java') {
-                    attributes {
-                        attribute('name': 'optional', 'value': 'true')
-                    }
-                }
-                classpathentry('kind': 'src', 'output': 'build/eclipse/src', 'path': 'build/coverage-cmake/generated-java') {
                     attributes {
                         attribute('name': 'optional', 'value': 'true')
                     }
--- a/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
+++ b/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
@@ -17,6 +17,7 @@ import static com.amazon.corretto.crypto.provider.HkdfSecretKeyFactorySpi.HKDF_W
 import static com.amazon.corretto.crypto.provider.HkdfSecretKeyFactorySpi.HKDF_WITH_SHA256;
 import static com.amazon.corretto.crypto.provider.HkdfSecretKeyFactorySpi.HKDF_WITH_SHA384;
 import static com.amazon.corretto.crypto.provider.HkdfSecretKeyFactorySpi.HKDF_WITH_SHA512;
+import static com.amazon.corretto.crypto.provider.Loader.FIPS_BUILD;
 import static com.amazon.corretto.crypto.provider.Loader.PROVIDER_VERSION;
 import static com.amazon.corretto.crypto.provider.Loader.PROVIDER_VERSION_STR;
 import static java.lang.String.format;
@@ -70,8 +71,7 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
 
   static {
     if (!Loader.IS_AVAILABLE && DebugFlag.VERBOSELOGS.isEnabled()) {
-      getLogger("AmazonCorrettoCryptoProvider")
-          .fine("Native JCE libraries are unavailable - disabling");
+      getLogger(PROVIDER_NAME).fine("Native JCE libraries are unavailable - disabling");
     }
     INSTANCE = new AmazonCorrettoCryptoProvider();
   }
@@ -461,7 +461,7 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
         case FAILED:
           synchronized (this) {
             if (!failMessagePrinted) {
-              getLogger("AmazonCorrettoCryptoProvider")
+              getLogger(PROVIDER_NAME)
                   .severe(
                       "Self tests failed - disabling. "
                           + "Detailed results: "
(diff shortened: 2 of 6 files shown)


src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -500,13 +500,15 @@ private void resetAllSelfTests() {
...
+  // The super constructor taking a double version is deprecated in java 9. However, the alternate
+  // constructor taking a string version is unavailable in java 8. So to build on both with
+  // warnings on, our only choice is to suppress deprecation warnings.
   @SuppressWarnings({"deprecation"})
   public AmazonCorrettoCryptoProvider() {
-    super("AmazonCorrettoCryptoProvider", PROVIDER_VERSION, "");
+    super(
+        PROVIDER_NAME,
+        PROVIDER_VERSION,
+        PROVIDER_NAME + ' ' + PROVIDER_VERSION_STR + (FIPS_BUILD ? "-FIPS" : ""));
```
brian-jarvis-aws, line 511: This will produce something like `AmazonCorrettoCryptoProvider X.Y.Z-FIPS` in the FIPS case, right? It would seem more natural to me for this to be `AmazonCorrettoCryptoProvider-FIPS X.Y.Z` (see, for example, how the [Maven package](https://central.sonatype.com/artifact/software.amazon.cryptools/AmazonCorrettoCryptoProvider-FIPS) name and version number is constructed).

geedo0 (approved):

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
chockalingamc: Reason for that format was that FIPS is more of a version qualifier than provider qualifier. 
The [java.security.Provider](https://docs.oracle.com/javase/9/docs/api/java/security/Provider.html#Provider-java.lang.String-java.lang.String-java.lang.String-) class also supports that format of version strings (number followed by other optional information separated by a '+' or '-').

Similar version identifiers are seen in [semantic versioning standards](https://semver.org/#spec-item-9) as well.

geedo0: Disagree (but overall indifferent and approved the PR regardless), FIPS isn't along the same lines of [alpha, beta, rc1, rc2...] used as version qualifiers. It's more like a different flavor of ACCP itself rather than something used as the release version.

From the semver doc:

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version ... A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92, 1.0.0-x-y-z.--.

chockalingamc: In that case, we probably should update the provider name itself to suffix -FIPS ?

build.gradle
```diff
@@ -887,11 +891,6 @@ task generateEclipseClasspath {
                         attribute('name': 'optional', 'value': 'true')
                     }
                 }
-                classpathentry('kind': 'src', 'output': 'build/eclipse/src', 'path': 'build/coverage-cmake/generated-java') {
```
chockalingamc, line 890: `build/coverage-cmake/generated-java` should be actually `build/cmake-coverage/generated-java`. 
But removing this classpathentry because if the correct folder is used it will create duplicate classes in the classpath, as the previous classpath entry `build/cmake/generated-java` has the exact same artifacts.

build.gradle
```diff
@@ -887,11 +891,6 @@ task generateEclipseClasspath {
...
-                    attributes {
-                        attribute('name': 'optional', 'value': 'true')
-                    }
-                }
```
chockalingamc, line 894: `build/coverage-cmake/generated-java` should be actually `build/cmake-coverage/generated-java`. 
But removing this classpathentry because if the correct folder is used it will create duplicate classes in the classpath, as the previous classpath entry `build/cmake/generated-java` has the exact same artifacts.

build.gradle
```diff
@@ -486,6 +483,12 @@ task single_test(type: Exec) {
     commandLine 'make', 'check-junit-single'
 }
 
+task test_install_via_properties(type: Exec) {
+    dependsOn executeCmake
+    workingDir "${buildDir}/cmake"
+    commandLine 'make', 'check-install-via-properties'
+}
+
 task test_integration_exec(type: Exec) {
```
chockalingamc, line 492: Added a gradle task to run the standalone `SecurityPropertyTester` test. The `single_test` option does not work here as it is not a Junit test.

build.gradle
```diff
@@ -15,11 +15,7 @@ plugins {
 group = 'software.amazon.cryptools'
 version = '2.5.0'
 ext.isExperimentalFips = Boolean.getBoolean('EXPERIMENTAL_FIPS')
-if (ext.isExperimentalFips) {
-    ext.isFips = true
-} else {
-    ext.isFips = Boolean.getBoolean('FIPS')
-}
+ext.isFips = ext.isExperimentalFips || Boolean.getBoolean('FIPS')
```
chockalingamc, line 19: Just refactoring

build.gradle
```diff
@@ -486,6 +483,12 @@ task single_test(type: Exec) {
     commandLine 'make', 'check-junit-single'
 }
 
+task test_install_via_properties(type: Exec) {
```
chockalingamc, line 486: Added a gradle task to run the standalone SecurityPropertyTester test. The single_test option does not work here as it is not a Junit test.

build.gradle
```diff
@@ -486,6 +483,12 @@ task single_test(type: Exec) {
...
+    dependsOn executeCmake
+    workingDir "${buildDir}/cmake"
+    commandLine 'make', 'check-install-via-properties'
+}
+
```
chockalingamc, line 491: Added a gradle task to run the standalone SecurityPropertyTester test. The single_test option does not work here as it is not a Junit test.

build.gradle
```diff
@@ -887,11 +891,6 @@ task generateEclipseClasspath {
                         attribute('name': 'optional', 'value': 'true')
                     }
                 }
-                classpathentry('kind': 'src', 'output': 'build/eclipse/src', 'path': 'build/coverage-cmake/generated-java') {
-                    attributes {
-                        attribute('name': 'optional', 'value': 'true')
-                    }
-                }
```
chockalingamc, line 894: `build/coverage-cmake/generated-java` should be actually `build/cmake-coverage/generated-java`. 
But removing this classpathentry because if the correct folder is used it will create duplicate classes in the classpath, as the previous classpath entry `build/cmake/generated-java` has the exact same artifacts.

build.gradle
```diff
@@ -15,11 +15,7 @@ plugins {
 group = 'software.amazon.cryptools'
 version = '2.5.0'
 ext.isExperimentalFips = Boolean.getBoolean('EXPERIMENTAL_FIPS')
-if (ext.isExperimentalFips) {
-    ext.isFips = true
-} else {
-    ext.isFips = Boolean.getBoolean('FIPS')
-}
+ext.isFips = ext.isExperimentalFips || Boolean.getBoolean('FIPS')
```
chockalingamc, line 18: Just refactoring

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -508,7 +510,14 @@ public AmazonCorrettoCryptoProvider() {
...
         PROVIDER_VERSION,
-        PROVIDER_NAME + ' ' + PROVIDER_VERSION_STR + (FIPS_BUILD ? "-FIPS" : ""));
+        PROVIDER_NAME
+            + ' '
+            + PROVIDER_VERSION_STR
+            + (FIPS_BUILD ? "+FIPS" : "")
+            + (EXPERIMENTAL_FIPS_BUILD ? "+EXP" : "")
+            + " ("
+            + AWS_LC_VERSION_STR
+            + ")");
```
geedo0, line 520: Can you use a format string instead? This is getting hard to parse.

src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -153,6 +156,20 @@ static String getProperty(String propertyName, String def) {
...
+                  () -> {
+                    try (InputStream is = Loader.class.getResourceAsStream("version.properties")) {
+                      if (is == null) {
+                        return System.getProperty(PROPERTY_AWS_LC_VERSION_STR);
+                      }
+                      Properties p = new Properties();
+                      p.load(is);
+                      return p.getProperty(PROPERTY_AWS_LC_VERSION_STR);
+                    }
+                  });
```
geedo0, line 172: Would be good to extract a method here to avoid the repetition of this routine.

build.gradle
```diff
@@ -531,6 +534,7 @@ task coverage_cmake(type: Exec) {
     args '-DPROVIDER_VERSION_STRING=' + version, projectDir
     args "-DTEST_RUNNER_JAR=${configurations.testRunner.singleFile}"
     args "-DCMAKE_VERBOSE_MAKEFILE:BOOL=ON"
+    args "-DAWS_LC_VERSION_STRING=" + awsLcGitVersionId
```
geedo0, line 537: I applaud the effort of plumbing this through for the info string, but this is not the only build script you have to update to introduce this cmake flag. There's a lot of other paths through which ACCP gets built in a way that bypasses Gradle (ask offline). Since you added an assertion for it it will also break all of those builds.

tst/com/amazon/corretto/crypto/provider/test/SecurityPropertyTester.java
```diff
@@ -31,12 +31,22 @@ public static void main(String[] args) throws Exception {
 
     String infoStr = NATIVE_PROVIDER.getInfo();
     System.out.println("Security Provider : " + infoStr);
-    assertTrue(infoStr.startsWith(NATIVE_PROVIDER.getName()));
+    String[] tokens = infoStr.split("[()]");
+    String[] semver = tokens[0].split("[\\s+-]");
+    String[] addInfo = tokens[1].split(",");
```
geedo0, line 36: Suggest not being this strict with the testing. This is supposed to just be an informative string, and we don't need to rigorously validate the format and contents. Just makes it more of a hassle to adapt it later on.

tst/com/amazon/corretto/crypto/provider/test/SecurityPropertyTester.java
```diff
@@ -31,12 +31,22 @@ public static void main(String[] args) throws Exception {
...
     } else {
-      assertTrue(infoStr.endsWith(NATIVE_PROVIDER.getVersionStr()));
+      assertFalse(infoStr.contains("FIPS"));
     }
+    if (NATIVE_PROVIDER.isExperimentalFips()) {
+      assertEquals("EXP", semver[3]);
+    } else {
+      assertFalse(infoStr.contains("EXP"));
+    }
+    assertEquals(NATIVE_PROVIDER.getAwsLcVersionStr(), addInfo[0]);
```
geedo0, line 49: This will break some of our other builds which have not been updated to account for the new cmake variable.

geedo0 (changes requested):

build.gradle
chockalingamc: Ack. Will update other relevant paths.

geedo0 (approved):

alexw91 (approved):

PR #429, opened 2025-02-06
Refactor Cipher/AES/CBC Service Registration

geedo0:
*Issue #, if available:*
ACCP-122, V1600360624

*Description of changes:*
Previously, we registered the AES/CBC service by registering separate services for each combination of cipher_aliases x padding_aliases. This results in a lot more services than is necessary, is more verbose, and more importantly breaks JCK compatibility tests. This refactor properly registers our AES/CBC implementation as one service that has multiple cipher aliases and different supported padding schemes.

To fix the JCK test, we implement the `engineSetPadding` interface and throw a `NoSuchPaddingException` as needed.

We also take the opportunity to make a quality of life improvement to the `singleTest` gradle task to force a rebuild the Java source to avoid staleness issues during development.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -480,9 +480,17 @@ task unit_tests(type: Copy) {
 }
 test.dependsOn unit_tests
 
-task single_test(type: Exec) {
+task singleTest(type: Exec) {
+    group 'Verification'
+    description 'Pass in the test class using -DSINGLE_TEST=${fully_qualified_test_class}'
     dependsOn executeCmake
     workingDir "${buildDir}/cmake"
+    // Our cmake doesn't properly react java source changes, but it will rebuild them if the jars are missing
+    doFirst {
+        delete fileTree("${buildDir}/cmake").matching {
+            include '*.jar'
+        }
+    }
     commandLine 'make', 'check-junit-single'
 }
 
--- a/src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
+++ b/src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
@@ -18,6 +18,7 @@ import java.security.spec.InvalidParameterSpecException;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.function.Predicate;
 import javax.crypto.BadPaddingException;
 import javax.crypto.Cipher;
 import javax.crypto.CipherSpi;
@@ -96,7 +97,7 @@ class AesCbcSpi extends CipherSpi {
 
   // State
   private CipherState cipherState;
-  private final int padding;
+  private int padding;
   // CBC processes data one block at a time. There are two scenarios where not all the input passed
   // to engineUpdate is processed:
   //     1. Input length is not a multiple of the block size,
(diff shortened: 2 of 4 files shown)


src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
```diff
@@ -142,14 +143,29 @@ private boolean noPadding() {
...
   protected void engineSetMode(final String mode) throws NoSuchAlgorithmException {
-    // no op. One only needs to provide an implementation if the same Spi class instance can be used
-    // for different modes.
+    if (!"CBC".equalsIgnoreCase(mode)) {
+      throw new NoSuchAlgorithmException("Only CBC mode is supported.");
+    }
   }
 
   @Override
   protected void engineSetPadding(final String padding) throws NoSuchPaddingException {
```
alexw91, line 152: There are two variables named `padding`. `private int padding;` on line 100, and `final String padding` here on line 152. Can you rename one of them so that they don't have the same name?

geedo0: Took this a step further and replaced the field's type with the Padding enum to give us a bit more compile time guarantees around the possible padding values passed to AWS-LC.

alexw91 (approved):

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
```diff
@@ -137,19 +138,34 @@ private enum CipherState {
...
-    // no op. One only needs to provide an implementation if the same Spi class instance is used for
-    // different paddings.
+    if (padding == null) {
+      throw new NoSuchPaddingException("Padding cannot be null.");
+    }
+    if (!inputIsEmpty) {
+      throw new NoSuchPaddingException("Padding cannot be set during an operation.");
+    }
+    Predicate<String> paddingPredicate = n -> n.split("/")[2].equalsIgnoreCase(padding);
```
WillChilds-Klein, line 159: dereferencing the `split(...)` result without checking result seems like a recipe for NPE for malformed `padding` String inputs. can we add a check?

WillChilds-Klein (approved):
Great improvements @geedo0! I have one concern noted in comments, but benefits of this PR as written outweigh that concern.

build.gradle
```diff
@@ -480,9 +480,17 @@ task unit_tests(type: Copy) {
 }
 test.dependsOn unit_tests
 
-task single_test(type: Exec) {
+task singleTest(type: Exec) {
+    group 'Verification'
+    description 'Pass in the test class using -DSINGLE_TEST=${fully_qualified_test_class}'
     dependsOn executeCmake
     workingDir "${buildDir}/cmake"
+    // Our cmake doesn't properly react java source changes, but it will rebuild them if the jars are missing
```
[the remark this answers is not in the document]
geedo0: No problem! It was so annoying trying to figure out what was stale, how to clean it, how to avoid a full LC rebuild... This seemed to hit the 80/20 of faster builds and addressing most of the staleness in the ACCP dev-loop.

src/com/amazon/corretto/crypto/provider/AesCbcSpi.java
geedo0: I made the sets immutable. I also tried to refactor it to avoid expanding out all the name combinations, but it looks more annoying than expected (have to dig into all the ways the JDK passes in these names).

WillChilds-Klein (approved):

alexw91 (approved):

PR #432, opened 2025-02-12
Support Ed25519ph

WillChilds-Klein:
*Issue #, if available:* ACCP-123

*Description of changes:*

This PR implements [Ed25519ph](https://datatracker.ietf.org/doc/html/rfc8032#section-5.1) ("pre-hash") support in ACCP. We conform to SunEC's and BouncyCastle's implementations, using SHA-512 to digest the message as `Signature` is updated. Unlike regular Ed25519, we use `EvpSignature` at the java layer, which calls into AWS-LC's `EVP_DigestSign` instead of `EVP_PKEY_sign` as `EvpSignatureRaw` uses. To support this,  we perform a slight refector, pulling up the `preHash` constructor parameter to `EvpSignatureBase` and removing all constructor overloads to make the code more explicit. This `preHash` boolean is then plumbed down to native context initialization and used to determine whether EVP_PKEY conversion is required (see source comment in `sign.cpp`).

Unlike JDK, we do not support specifying "pre-hash" via [EdDSAParameterSpec](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/security/spec/EdDSAParameterSpec.html) due to complications in compiling references to it on JDK < 15. If we do end up requiring this particular point of compatibility, we can implement it in the future using reflection. Today, callers request "pre-hash" via the signature algorithm name, i.e. `Signature.getInstance("Ed25519ph");`.

We provide interoperability tests with SunEC and BouncyCastle as well as a simple Known Answer Test from RFC 8032. To support these tests, we implement a `Signature` instance for BouncyCastle using its lower-level `Signer` API, as BouncyCastle doesn't provide a JCA `Signature` implementation for Ed25519ph.

---


src/com/amazon/corretto/crypto/provider/EvpSignature.java
```diff
@@ -16,6 +16,7 @@ class EvpSignature extends EvpSignatureBase {
    * @param digestPtr the value from {@link Utils#getEvpMdFromName(String)} representing the digest
    *     to use with this signature
    * @param paddingType the integer defined by OpenSSL as the padding type to be used.
+   * @param preHash whether the signature is a modern "pre-hash" variant
```
geedo0, line 19: Are there going to be pre-hash variants for all signature algorithms in our modern times? I know we're anticipating this for ML-DSA, but are these available for ECDSA and any other algorithm?

csrc/sign.cpp
```diff
@@ -76,6 +77,32 @@ bool initializeContext(raii_env& env,
             throw_openssl("Unable to create MD_CTX");
         }
 
+#if !defined(FIPS_BUILD) || defined(EXPERIMENTAL_FIPS_BUILD)
+        if (preHash && EVP_PKEY_id(pKey) == EVP_PKEY_ED25519) {
```
geedo0, line 81: Since you plan on having other pre-hash algorithms. It makes sense to me for you to pull down this `pkey==ED25519` logic to nest underneath the `if(preHash)`.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
```diff
@@ -117,81 +201,130 @@ public void keyFactoryValidation() throws GeneralSecurityException {
...
+
+  @Test // sanity check to assert that JCE and BC are interoperable
+  public void bcJceInteropValidationPh() throws GeneralSecurityException {
+    assumeTrue(ed25519phIsEnabled());
+    final Signature jceSig = Signature.getInstance("Ed25519", "SunEC");
+    makeJceSignaturePh(jceSig);
+    testInteropValidation(jceSig, bcPrehashSig, true);
+  }
+
+  public void testInteropValidation(Signature signer, Signature verifier, boolean preHash)
```
geedo0, line 256: Nit: This test tests both directions so there's no distinction between signer/verifier in the interface.

geedo0 (dismissed):

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
WillChilds-Klein: that's a fair point. update.

csrc/sign.cpp
WillChilds-Klein: This block of logic is very much specific to Ed25519 and the NID conversion we do there. We do have other `preHash` (the local variable) above on L74 for ML-DSA ExternalMu, and it applies differently.

src/com/amazon/corretto/crypto/provider/EvpSignature.java
WillChilds-Klein: This is a good question, and I have no crystal ball, but I doubt "pre-hash" standards would be applied retrospectively to older (EC)DSA. NIST cryptographic standards over the last ~30 years seem pretty spiky -- bump in the early 2000's with AES, ECDSA, SHA2. Then the current bump with PQ and the new ECC stuff (edwards, x25519). With that, they seem to push newer standard algorithms and focus little on standards-level "backporting".

As I understand it, the reason for the new pre-hash variants has to do with algorithm internals (Ed25519 [incorporating key material](https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.6) into the hash, [ML-DSA's µ](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf#algorithm.7)). Whereas ECDSA [always signs a straight-up hash of the message](https://csrc.nist.gov/files/pubs/fips/186-3/final/docs/fips_186-3.pdf#%5B%7B%22num%22%3A95%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22Fit%22%7D%5D).

geedo0 (approved):

alexw91 (approved):

PR #436, opened 2025-02-27
Add some guidance for running Java debugger

geedo0:
*Issue #, if available:*
N/A

*Description of changes:*
This finishes wiring up the TEST_JAVA_ARGS CMake variable to allow
builders to override the JVM flags during unit test runs. We can then
use that to instantiate a remote debug server as a workaround for
debugging ACCP.


DEVELOPMENT-GUIDE.md
```diff
@@ -191,4 +191,27 @@ So, when you are in a critical region there are exceedingly few operations you c
...
+incorporate the natively compiled portion of the library. A simple workaround is to use the singleTest task along with
+JVM arguments to start a remote Java debugger session. This then allows you to connect and debug your test in the IDE.
+
+```
+# 1. Create a new Run/Debug Configuration for "Remote JVM Debug" in IntelliJ or the equivalent for your IDE.
+
+# 2. Launch your test target using the singlelTest target. n.b. Suspend is set to y to break immediately.
+TEST_CLASS=com.amazon.corretto.crypto.provider.test.KeyGeneratorTest
+./gradlew \
+-DTEST_JAVA_ARGS='-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005' \
```
WillChilds-Klein, line 208: nit: would `*:5005` bind to a non-localhost interface? if we can restrict this to localhost, might be best to not open up ports to the network unnecessarily during development

WillChilds-Klein (approved):
nice!

DEVELOPMENT-GUIDE.md
geedo0: This won't open any new ports without an iptables/firewall/security group intervention. That's on user to explicitly open it up.

WillChilds-Klein (approved):

justsmth (approved):

Issue #332, opened 2023-09-25
JDK21 LTS Builds


alexw91:
JDK 21 LTS was released on September 19th, we should add builds of ACCP that use this new JDK version.

PR #437, opened 2025-02-28
Upgrade gradle wrapper to latest version (8.13)

chockalingamc:
Upgrade gradle wrapper to latest version (8.13)

*Issue #:* [332](https://github.com/corretto/amazon-corretto-crypto-provider/issues/332)

*Description of changes:*
In prep for full Java 21 support, gradle needs to be updated to version 8.5 or later. 
https://docs.gradle.org/8.5/release-notes.html#java-21 
https://docs.gradle.org/8.5/userguide/compatibility.html#java
Updating gradle to latest version available now (8.13). 

Also upgrading plugin 'com.google.osdetector' to version '1.7.3' to fix gradle deprecation warnings.


build.gradle
```diff
@@ -480,8 +476,8 @@ task unit_tests(type: Copy) {
 test.dependsOn unit_tests
 
 task singleTest(type: Exec) {
-    group 'Verification'
-    description 'Pass in the test class using -DSINGLE_TEST=${fully_qualified_test_class}'
+    group = 'Verification'
```
geedo0, line 479: Did something change in the new version of gradle? This is inconsistent with the rest of our grade source. Omitting the = is just some groovy syntactic sugar.

gradlew
WillChilds-Klein, line 1: were these updates auto-generated in the upgrade? if so, by which command?

geedo0 (approved):
LGTM assuming the CI failures are false alarms.

gradlew
geedo0: Yeah, gradlew gets generated when you initialize a gradle project. The expectation is to generate it, check it into the repo, and not touch it

chockalingamc: yeah, if you already have gradle 8.1 (the 'latest' label was introduced since 8.1) 
    `gradle wrapper --gradle-wrapper latest` 
or you could specify the exact version if upgrading from lower versions
    `gradle wrapper --gradle-wrapper 8.13`. 

The command updates the gradle.properties, the wrapper JAR, gradlew, & gradlew.bat.

The changes related to exec commands are manual (to fix a deprecated warning for exec API and use the new Provider.exec API

build.gradle
chockalingamc: These were warnings that will be flagged as errors in future versions (10.0). 
```
Properties should be assigned using the 'propName = value' syntax. Setting a property via the Gradle-generated 'propName value' or 'propName(value)' syntax in Groovy DSL has been deprecated.
This is scheduled to be removed in Gradle 10.0.

Solutions
[enum]  Use assignment ('description = <value>') instead.
```

Reg other such assignments without a '=' I guess known parameters to a specific task type, like 'dependsOn', 'commandLine' etc to a 'Exec' type task are different. The 'group' and 'description' params are not part of the task type spec. They are additional metadata I presume.

geedo0: Thanks for explaining, that's good to know. I actually prefer it this way, less implicit behavior that you have to keep track of.

geedo0 (changes requested):
The legacy platform CI task is failing due to libc linker issues. We don't need the latest version of gradle to work on that, but we do need to ensure that we can compile ACCP's native code successfully on the version of GCC present on that machine.

```
Caused by: java.lang.UnsatisfiedLinkError: /root/.gradle/native/0.2.7/x86_64-linux-gnu/libgradle-fileevents.so: /lib/libc.so.6: version `GLIBC_2.14' not found (required by /root/.gradle/native/0.2.7/x86_64-linux-gnu/libgradle-fileevents.so)
```

chockalingamc:
Pushing out a new PR #442 from latest main. Closing this out.

chockalingamc:
> The legacy platform CI task is failing due to libc linker issues. We don't need the latest version of gradle to work on that, but we do need to ensure that we can compile ACCP's native code successfully on the version of GCC present on that machine.
> 
> ```
> Caused by: java.lang.UnsatisfiedLinkError: /root/.gradle/native/0.2.7/x86_64-linux-gnu/libgradle-fileevents.so: /lib/libc.so.6: version `GLIBC_2.14' not found (required by /root/.gradle/native/0.2.7/x86_64-linux-gnu/libgradle-fileevents.so)
> ```

Yes, Gradle's native integration is compiled with GLIBC 2.14 which is not available in the older ubuntu 10.04 platform.

Updated our codebuild CI task for legacy platform to include the system property `org.gradle.native=false` while running gradle. This disabled native integration and lets gradle run successfully.

PR #442, opened 2025-03-12
Upgrade gradle wrapper to version 8.5 for JDK21 support

chockalingamc:
Upgrade gradle wrapper to version 8.5 for JDK21 support

*Issue #:* [332](https://github.com/corretto/amazon-corretto-crypto-provider/issues/332)

*Description of changes:*
In prep for full Java 21 support, gradle needs to be updated to version 8.5 or later. 
https://docs.gradle.org/8.5/release-notes.html#java-21 
https://docs.gradle.org/8.5/userguide/compatibility.html#java

Note: Tried updating gradle to the latest version available now (8.13), but that failed CI tests with linker errors on legacy platforms (newer GLIBC_2.14 not available in older Ubuntu 10.04 platform). So staying with 8.5 until legacy builds are deprected.

Also upgrading plugin 'com.google.osdetector' to version '1.7.3' to fix gradle deprecation warnings.

(Closed out a old PR #437 related to this and starting clean from latest main branch)


chockalingamc:
Reg **AccpCiLegacyTests** failure:

Gradle's native integration in v 8.13 is compiled with GLIBC 2.14 which is not available in the older ubuntu 10.04 platform.

Updated our codebuild CI task for legacy platform to include the system property `org.gradle.native=false` while running gradle. This disabled native integration and let gradle run successfully.

build.gradle
```diff
@@ -647,11 +643,23 @@ task coverage {
     dependsOn coverage_java_report
 }
 
+// Since gradle upgrade to 8.13, Gradle macOS CI workflow fails with this error:
```
geedo0, line 646: Since we're using Gradle 8.5 now, do we still need this workaround? Want to avoid changing anything we don't need to.

tests/ci/run_accp_basic_tests.sh
```diff
@@ -64,10 +64,30 @@ fi
 export JAVA_HOME=$TEST_JAVA_HOME
 export PATH=$JAVA_HOME/bin:$PATH
 
+# Since gradle upgrade to 8.13, Gradle macOS CI workflow fails with this error:
```
geedo0, line 67: Same here

build.gradle
chockalingamc: yeah, I still intend to upgrade to 8.13 in future once I figure out a way to bypass the native stuff that causes the linker error.

chockalingamc: will remove for now and add back when upgrading to 8.13

WillChilds-Klein (approved):

gradle/wrapper/gradle-wrapper.jar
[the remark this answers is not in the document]
chockalingamc: Every new gradle version should have some update to the wrapper jar. The difference is not visible here, but shows up well in IDE diff. We should keep checking in the updated JAR to git.

PR #447, opened 2025-03-17
Fix warnings from JDK21 compiler

chockalingamc:
*Issue #332*

*Description of changes:*
Fix warnings from JDK21 compiler. If not, these will fail build due to cmake `-Werror` flag.

Warnings:

```
warning: [serial] non-transient instance field of a serializable class declared with a non-serializable type
```
```
warning: [options] source value 8 is obsolete and will be removed in a future release
warning: [options] target value 8 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options
```
Update: Also fixed a CI issue due to clang-format. Removed enforcing clang-format based spotless checks in CI runners.

### DIFF
--- a/.github/workflows/macos_ci_jobs.yml
+++ b/.github/workflows/macos_ci_jobs.yml
@@ -17,14 +17,14 @@ jobs:
   # JDK 8 job.
   macOS8:
     name: JDK8
-    runs-on: macos-latest
+    runs-on: macos-latest-large
     steps:
     - uses: actions/checkout@v3
       with:
         submodules: true
     - name: Install dependencies from brew
       run: |
-        env HOMEBREW_NO_AUTO_UPDATE=1 brew install lcov golang clang-format
+        env HOMEBREW_NO_AUTO_UPDATE=1 brew install lcov golang
     - name: Setup pip
       uses: actions/setup-python@v4
       with:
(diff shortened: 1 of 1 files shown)


chockalingamc:
The members are already non-serializable. JDK21 forces acknowledging it. 
The keys themselves are still serialized. These members are just not part of the ser/deser cycle. They get default values upon deserialization.

BTW `EvpKey` and the derived types `EvpEcKey` and `EvpEcPublicKey` are not directly serializable either. It overrides readResolve() and writeReplace() for custom ser/deser.

geedo0 (approved):

alexw91 (approved):

chockalingamc:
No control on the version of clang-format installed by brew in Github provided macOS CI runners has broken our build now.
JDK11 and JDK17 CI test runners installed clang-format 20.1.0, while others are still stuck on clang-forma 19.1.7.

chockalingamc:
Also, our CI jobs are all running on ARM ! 
Need to use either [macos-latest-large](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners#about-macos-larger-runners) or [macos-13](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) to be on `Intel` platform.

geedo0:
I think it's time we got rid of `clang-format` as a strict build requirement. It's just too inconsistent across multiple dimensions and is causing more trouble than it's worth for us. The simple way to do this is to simply remove the package from the workflow definition.

geedo0 (approved):

alexw91 (approved):

PR #456, opened 2025-04-22
Throw InvalidKeyException in sig init if Key alg is null

WillChilds-Klein:
*Issue #, if available:* t/V1749540670

*Description of changes:*

This is a small change to match [OpenJDK's behavior](https://github.com/corretto/corretto-17/blob/f11c13eeefe3d8795a7ab6fd78cb9479786fea5a/src/java.base/share/classes/sun/security/rsa/RSASignature.java#L124C21-L126).

We also modify `isFipsStatusOk()` to only consider AWS-LC's self-tests and disregard ACCP's own, which are irrelevant to FIPS. Using the full self test suite caused a deadlock on JDK8 because HMAC self-tests calling `Mac.newInstance` would check `isFipsStatusOk()`, which would in turn await the HMAC self-tests, etc.

---

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
+++ b/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
@@ -689,14 +689,15 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
       final long timeout = 3 * 1000;
       final long deadline = System.currentTimeMillis() + timeout;
       while (status == SelfTestStatus.NOT_RUN) {
+        if (System.currentTimeMillis() > deadline) {
+          throw new RuntimeCryptoException("FIPS self tests timed out");
+        }
         try {
           Thread.sleep(1);
         } catch (Exception e) {
           throw new RuntimeCryptoException(e);
         }
-        if (System.currentTimeMillis() > deadline) {
-          throw new RuntimeCryptoException("FIPS self tests timed out");
-        }
+        status = SelfTestSuite.AWS_LC_SELF_TESTS.getCachedResult().getStatus();
       }
     }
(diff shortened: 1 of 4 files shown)


geedo0 (approved):

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -682,19 +682,22 @@ public boolean isFipsStatusOk() {
...
-    if (getSelfTestStatus() == SelfTestStatus.NOT_RUN) {
-      // If FIPS self tests haven't completed, give them a 5s timeout to complete.
+    // For "FipsStatus", we only care about AWS-LC's self test status.
+    SelfTestStatus status = SelfTestSuite.AWS_LC_SELF_TESTS.getCachedResult().getStatus();
+    if (status == SelfTestStatus.NOT_RUN) {
+      // If FIPS self tests haven't completed, give them a 3s timeout to complete.
       final long timeout = 3 * 1000;
       final long deadline = System.currentTimeMillis() + timeout;
-      while (getSelfTestStatus() == SelfTestStatus.NOT_RUN) {
+      while (status == SelfTestStatus.NOT_RUN) {
```
alexw91, line 691: Nit: You are checking `status == SelfTestStatus.NOT_RUN` twice in a row, once in the if statement and then immediately again in the while statement. This should be able to be refactored and merged so that it's only checked once.

WillChilds-Klein: i'm intentionally guarding the call to `System.currentTimeMillis()` with the first check on L687, but i suppose i can change this to a do/while

alexw91 (approved):

PR #458, opened 2025-04-29
Support AES CFB

WillChilds-Klein:
# Description

This PR adds support for AES [CFB][1] to ACCP. Our CFB implementation is a bit simpler than other AES modes in ACCP. Like CBC, CFB does not have an authentication tag to manage. Like GCM, we do not support any padding schemes.

Because CFB [currently lives](https://github.com/aws/aws-lc/blob/main/crypto/decrepit/cfb/cfb.c) in AWS-LC's "libdecrepit" outside of the FIPS module boundary, we guard this feature under the "experimental FIPS" flag until we can consume a future version of AWS-LC-FIPS that moves CFB into the module boundary.

# Method

I took this feature as an opportunity to experiment with two prompt-based AI tools: [OpenHands](https://github.com/All-Hands-AI/OpenHands) (backed by Claude 3.7 sonnet) and [Q CLI](https://github.com/aws/amazon-q-developer-cli). This resulted in 3 trial PRs, each produced from a human-written `PROMPT.md` and guided by in-session prompting, but without any manual edits by me.

1. [Attempt 1][2] (OpenHands): The `PROMPT.md` for this attempt was woefully underspecified. It resulted in a long exchange and some shifty behavior from the agent. At one point it [changed a KAT][5] instead of fixing its implementation. The result was unusable.

2. [Attempt 2][3] (Q CLI): After refining the `PROMPT.md` to be more explicit, I was able to achieve useful results with Q CLI. It still required substantial in-session prompting, but the tool was able to make solid progress. The `PROMPT.md` file was handy in the face of context window exhaustion -- simply `/clear` the context and tell Q to reread the prompt. The resulting code appeared to work, but the implementation was very clunky. I think it over-indexed on the CBC implementation, adding a bunch of unnecessary padding schemes, despite [explicit instructions](https://github.com/WillChilds-Klein/amazon-corretto-crypto-provider/pull/7/files#diff-93f277324bed04c85a561677dccc3beb1afb2148c17c54aaa54393f9ecdb04cbR7) not to do so in `PROMPT.md`. It also induced a bunch of mysterious failures in unrelated tests that I didn't dig into.

3. [Attempt 3][4] (OpenHands): Using the same enhanced `PROMPT.md` as Attempt 2, I was able to achieve much better results with OpenHands than Attempt 1. The tool produced a cleaner implementation than Attempt 2 (although it produced a very convoluted benchmark). It also created an inexplicable test file in the project root and committed a copy of the ACCP jar into git for the benchmarks, which is obviously not ideal. It also added a redundant `decodeHex` in its test file when it should have used `TestUtil`. Tests all pass without affecting other tests.

I based this PR on Attempt 3, as you can see in the commit history below. You can read the final `PROMPT.md` and the generated `CHANGES.md` as removed [here](https://github.com/corretto/amazon-corretto-crypto-provider/pull/458/commits/6e86549461fb34dc04354f3552862fbc6376ecc9).

A commonality I found in all attempts' session prompting is that the AI tool(s) were ready to claim success prematurely. The lesson here is that operators should run tests manually to verify progress, and **be very careful in audits of generated test code**.

Notably, both Q CLI and OpenHands were able to take the NIST document linked in (attempt 2/3's) `PROMPT.md` and accurately extract the correct test vectors from NIST's PDF. After Attempt 1's nefarious corruption of said vectors, I made sure to verify that they were extracted correctly.

The AI tool(s) provided a solid base for this PR, but required: substantial clean up, some subtle behavioral fixes, and a lot of additional testing. The cleanup was straightforward, mostly consisting of finding unused parameters and using existing utilities instead of new implementations. Behavioral changes included [fixing][6] a potential resource leak in the case where a cipher is updated, but reset before being finalized. Additional testing is partially described below, and more fully available in the git history.

# Testing

KATs were pulled from [NIST SP 800-38A][7]. The test suite includes a (somewhat cursory) round-trip compatibility test with SunJCE. I added a number of unit tests to cover edge cases and in-place encrypt/decrypt. Coverage of `AesCfbSpi.java` code coverage is now at 90% with branch coverage at 96%.

Screenshot 2025-04-30 at 3 36 32 PM

# Performance

One-shot performance looks pretty solid:

Screenshot 2025-04-30 at 10 20 35 AM

Small updates, however, are much slower for both encrypt and decrypt (larger updates are fine). I suspect this is due to JNI overhead in a tight loop, and could be ameliorated with JDK19+'s [FFI](https://openjdk.org/jeps/424) on newer JDK runtimes. Results below:

Screenshot 2025-04-30 at 10 23 11 AM

We also see this performance deficiency for small encrypt updates [in AES GCM](https://dnabc10cx2yrf.cloudfront.net/c8gxlarge/index.html), but not for small decrypt updates. I suspect this discrepancy between CFB and GCM is due to ACCP's fast tag validation, which is not applicable to CFB. This is probably worth digging into further.

Benchmarks were run on an M3 Mac with this command executed from the `benchmarks/` directory:

```
./gradlew -PaccpLocalJar="../../build/cmake/AmazonCorrettoCryptoProvider.jar" -PincludeBenchmark="AesCfb*" lib:jmh
```

[1]: https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Cipher_feedback_(CFB)
[2]: https://github.com/WillChilds-Klein/amazon-corretto-crypto-provider/pull/6
[3]: https://github.com/WillChilds-Klein/amazon-corretto-crypto-provider/pull/7
[4]: https://github.com/WillChilds-Klein/amazon-corretto-crypto-provider/pull/8
[5]: https://github.com/WillChilds-Klein/amazon-corretto-crypto-provider/pull/6/commits/79c828a7654b6f10ff354c8a9181ef0c2101e299#diff-a8eb05f2e90823503a39e4025879e182f3f875ff655e68a823dfa1644663c91cR146
[6]: https://github.com/corretto/amazon-corretto-crypto-provider/pull/458/commits/7044c8663370a30429e9bc6dc220f3e74312deb1
[7]: https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://nvlpubs.nist.gov/nistpubs/legacy/sp/nistspecialpublication800-38a.pdf&ved=2ahUKEwihlem9-f-MAxWNkIkEHWX_PPQQFnoECAkQAQ&usg=AOvVaw3WVrP7fzeU7HWk2Y49aoOi

---


src/com/amazon/corretto/crypto/provider/AesCfbSpi.java
```diff
@@ -0,0 +1,322 @@
...
+  @Override
+  protected void engineInit(
+      final int opmode, final Key key, AlgorithmParameters params, final SecureRandom random)
+      throws InvalidKeyException, InvalidAlgorithmParameterException {
+    throw new UnsupportedOperationException(
+        "IV must be provided by passing an instance of IvParameterSpec.");
+  }
+
+  private void init(final int opmode, final byte[] keyBytes, final byte[] ivBytes) {
+    this.opMode = (opmode == Cipher.ENCRYPT_MODE) ? ENC_MODE : DEC_MODE;
```
geedo0, line 130: This allows the user to specify other modes like Wrap/Unwrap mode and end up with Decrypt. I'd rather we check for decrypt explicitly and throw an invalid parameter where the caller passes in junk.

tst/com/amazon/corretto/crypto/provider/test/AesCfbTest.java
```diff
@@ -0,0 +1,377 @@
...
+    assertArrayEquals(iv, cipher.getIV());
+    cipher.init(Cipher.ENCRYPT_MODE, key, SECURE_RANDOM);
+    assertFalse(
+        Arrays.equals(
+            iv, cipher.getIV())); // IV gen'd by SECURE_RANDOM should be different from |iv|
+    assertThrows(
+        UnsupportedOperationException.class,
+        () -> cipher.init(Cipher.ENCRYPT_MODE, key, (AlgorithmParameters) null, SECURE_RANDOM));
+
+    // We also don't support wrap/unsrap modes (yet?)
```
geedo0, line 360: I think this assertion is passing for the wrong reason (see above).

src/com/amazon/corretto/crypto/provider/AesCfbSpi.java
```diff
@@ -0,0 +1,322 @@
...
+    if (keyBytes.length != KEY_LEN_AES128 && keyBytes.length != KEY_LEN_AES256) {
+      throw new InvalidKeyException(
+          "Key length must be " + KEY_LEN_AES128 + " or " + KEY_LEN_AES256);
+    }
+
+    init(opmode, keyBytes, ivBytes);
+  }
+
+  @Override
+  protected void engineInit(
```
geedo0, line 122: Am I going crazy or is this the exact same function signature as above? i.e. This redefines engineInit?

src/com/amazon/corretto/crypto/provider/AesCfbSpi.java
```diff
@@ -0,0 +1,322 @@
...
+  }
+
+  @Override
+  protected byte[] engineGetIV() {
+    return iv.clone();
+  }
+
+  @Override
+  protected AlgorithmParameters engineGetParameters() {
+    return null;
```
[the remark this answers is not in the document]
WillChilds-Klein: I updated this to implement the behavior described [in JDK docs](https://docs.oracle.com/javase/8/docs/api/javax/crypto/CipherSpi.html#engineGetParameters--):

> The returned parameters may be the same that were used to initialize this cipher, or may contain a combination of default and random parameter values used by the underlying cipher implementation if this cipher requires algorithm parameters but was not initialized with any.
>
> Returns:
>    the parameters used with this cipher, or null if this cipher does not use any parameters.

geedo0 (dismissed):

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/AesCfbBase.java
```diff
@@ -0,0 +1,36 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider.benchmarks;
+
+import java.security.Key;
+import javax.crypto.Cipher;
+import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+
+public class AesCfbBase {
```
WillChilds-Klein, line 10: Some context here -- this and its deriving classes are basically identical to [the AES GCM benchmarks](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/AesGcmBase.java). I didn't want to overly crowd this PR by touching unrelated files, but I plan on following this up with another PR consolidating AES {GCM, CBC, CFB} into a unified set of base/derivative classes once this PR has been merged.

src/com/amazon/corretto/crypto/provider/AesCfbSpi.java
```diff
@@ -0,0 +1,333 @@
...
+        break;
+      case Cipher.DECRYPT_MODE:
+        this.opMode = DEC_MODE;
+        break;
+      default:
+        throw new InvalidAlgorithmParameterException("Invalid opmode: " + opmode);
+    }
+    this.keyLen = keyBytes.length;
+    this.key = keyBytes.clone();
+    this.ivParamSpec = new IvParameterSpec(ivBytes); // TODO [childw] enforce IV constraints
```
alexw91, line 167: Has this TODO been fixed?

alexw91:
LGTM. Just had one nit on what looks like a orphaned TODO.

src/com/amazon/corretto/crypto/provider/AesCfbSpi.java
WillChilds-Klein: ah, thanks for the catch. that's an artifact of [refactoring](https://github.com/corretto/amazon-corretto-crypto-provider/pull/458/commits/219889ff4c2e063d572d95dd185def1104a3ca92#diff-0fca1c846fc2c12ed2a4161a86bfaaa862e98e768f923cd80a74defd2a8853bdR169) the validation logic down into the common `private void init(...)`. removed.

alexw91 (approved):

geedo0 (approved):

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/AesCfbBase.java
WillChilds-Klein: https://github.com/corretto/amazon-corretto-crypto-provider/pull/460

PR #460, opened 2025-05-01
Refactor AES benchmarks, add AES CBC

WillChilds-Klein:
# Description

This commit cuts down on code duplication by refactoring common logic and parameters into `AesBase.java`. We also take the opportunity to add benchmarks for AES CBC. The consolidated benchmarks make it easy to parameterize by padding mode, but we defer this to avoid over-parameterizing the benchmark graphs until/if needed. We also considered marking cipher mode (i.e. GCM, CBC, or CFB) a benchmark parameter instead of distinct benchmark classes, but declined to do so for similar reasons.

# Testing

Ran the following command and confirmed that the suite subset executed correctly.

```
./gradlew -PaccpLocalJar="../../build/cmake/AmazonCorrettoCryptoProvider.jar" -PincludeBenchmark="Aes*" lib:jmh
```

### DIFF
--- a/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/AesBase.java
+++ b/benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/AesBase.java
@@ -1,0 +1,79 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider.benchmarks;
+
+import java.security.Key;
+import java.security.spec.AlgorithmParameterSpec;
+import javax.crypto.Cipher;
+import javax.crypto.spec.SecretKeySpec;
+
+public abstract class AesBase {
+  protected static final int PLAINTEXT_SIZE = 1024 * 1024;
+
+  protected Key key;
+  protected AlgorithmParameterSpec params1;
+  protected AlgorithmParameterSpec params2;
+  protected Cipher encryptor;
+  protected Cipher decryptor;
+  protected byte[] plaintext;
+  protected byte[] ciphertext;
+
+  protected abstract String getMode();
+
+  protected abstract AlgorithmParameterSpec createParameterSpec(byte[] iv);
+
+  protected abstract int getIvSize();
+
+  protected void setup(int keyBits, String provider, String padding) throws Exception {
+    BenchmarkUtils.setupProvider(provider);
+    key = new SecretKeySpec(BenchmarkUtils.getRandBytes(keyBits / 8), "AES");
+    params1 = createParameterSpec(BenchmarkUtils.getRandBytes(getIvSize()));
+    params2 = createParameterSpec(BenchmarkUtils.getRandBytes(getIvSize()));
+    final String algorithm = "AES/" + getMode() + "/" + padding;
+    encryptor = Cipher.getInstance(algorithm, provider);
+    decryptor = Cipher.getInstance(algorithm, provider);
+    encryptor.init(Cipher.ENCRYPT_MODE, key, params1);
+    decryptor.init(Cipher.DECRYPT_MODE, key, params1);
+    plaintext = BenchmarkUtils.getRandBytes(PLAINTEXT_SIZE);
+    ciphertext = encryptor.doFinal(plaintext);
+    encryptor.init(Cipher.ENCRYPT_MODE, key, params2);
+    decryptor.init(Cipher.DECRYPT_MODE, key, params2);
+  }
+
+  public byte[] oneShot1MiBEncrypt() throws Exception {
+    encryptor.init(Cipher.ENCRYPT_MODE, key, params1);
+    byte[] out = encryptor.doFinal(plaintext);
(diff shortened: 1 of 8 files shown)


benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/AesCbcUpdate.java
```diff
@@ -0,0 +1,58 @@
...
+
+@State(Scope.Benchmark)
+public class AesCbcUpdate extends AesBase {
+  @Param({"128", "256"})
+  public int keyBits;
+
+  @Param({AmazonCorrettoCryptoProvider.PROVIDER_NAME, "BC", "SunJCE"})
+  public String provider;
+
+  @Param({"NoPadding"}) // Other values: { "PKCS5Padding", "ISO10126Padding" }
```
geedo0, line 23: I think these would be interesting to add since the memory access patterns are really odd (especially for ISO10126Padding).

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/AesBase.java
```diff
@@ -0,0 +1,79 @@
...
+  protected abstract AlgorithmParameterSpec createParameterSpec(byte[] iv);
+
+  protected abstract int getIvSize();
+
+  protected void setup(int keyBits, String provider, String padding) throws Exception {
+    BenchmarkUtils.setupProvider(provider);
+    key = new SecretKeySpec(BenchmarkUtils.getRandBytes(keyBits / 8), "AES");
+    params1 = createParameterSpec(BenchmarkUtils.getRandBytes(getIvSize()));
+    params2 = createParameterSpec(BenchmarkUtils.getRandBytes(getIvSize()));
+    final String algorithm = "AES/" + getMode() + "/" + padding;
```
geedo0, line 32: nit: `padding` and `mode` should be wired up in the same way.

geedo0:
Nice refactor, this'll ensure consistency and let us parameterize/experiment things more easily in the future.

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/AesBase.java
WillChilds-Klein: Reason for this in description:

> We also considered marking cipher mode (i.e. GCM, CBC, or CFB) a benchmark parameter instead of distinct benchmark classes, but declined to do so for similar reasons.

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/AesCbcUpdate.java
WillChilds-Klein: I did that initially, but ran into problems with over-parameterization. JMH ran out of colors so just blacked-out half the bar graph.

geedo0: Agree that we don't strictly need this, but a 17% reduction relative to NoPadding in throughput is not negligible. If you're willing to humor me, I think it'd be useful to add a parameter for one of the padding schemes.

geedo0 (approved):

benchmarks/lib/src/jmh/java/com/amazon/corretto/crypto/provider/benchmarks/AesCbcUpdate.java
WillChilds-Klein: here's what the update benchmark looks like parameterized by padding. i agree that it's worth keeping some visibility on perf differences there. if we notice something, we can de-paramaterize the benchmarks a bit to dig in.

Screenshot 2025-05-05 at 12 12 38 PM

chockalingamc (approved):
Just an overall thought. The real common code seems to be just the setup part. The enc/dec methods are specific to AesCbcOneShot or AesCbcUpdate. May be they could live in those files instead of being grouped into AesBase and invoked via `super` calls. Anyways, I don't think its a blocker.

WillChilds-Klein:
> Just an overall thought. The real common code seems to be just the setup part. The enc/dec methods are specific to AesCbcOneShot or AesCbcUpdate. May be they could live in those files instead of being grouped into AesBase and invoked via `super` calls. Anyways, I don't think its a blocker.

Yeah, I hear that. I initially did the refactor with intermediate `AesOneShot` and `AesUpdate` classes, individual benchmark classes deriving those, and `AesBase` defining `abstract` encrypt/decrypt methods for the the intermediate classes to inherit. However, the leaf inheritors (e.g. `AesUpdateCbc`) still need to define methods themselves with JMH's `@Benchmark` decorator, so we can't fully get around leaf classes declaring methods then just calling e.g. `super.encrypt(...)`. Further, method signatures differ between Update and OneShot because Update takes a `chunkSize` parameter. This isn't a one-way door, and we can always look at more complex class hierarchies in the future, but I think the simplest architecture for now is flat -- single `AesBase` that the various individual benchmarks inherit from.

geedo0 (approved):

PR #461, opened 2025-05-01
Increase AWS-LC FIPS error message limit

WillChilds-Klein:
*Issue #, if available:*

*Description of changes:*

Additionally, we truncate oversized messages and append them to the queue to prevent information loss.

Finally, we add an adjacent gradle configuration clarification to `README.md`.

---

### DIFF
--- a/csrc/fips_status.cpp
+++ b/csrc/fips_status.cpp
@@ -17,7 +17,8 @@ extern "C" void AWS_LC_fips_failure_callback(char const* message);
 #if defined(FIPS_SELF_TEST_SKIP_ABORT)
 void AWS_LC_fips_failure_callback(char const* message)
 {
-    const size_t char_limit = 128;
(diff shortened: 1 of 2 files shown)


geedo0 (approved):

csrc/fips_status.cpp
```diff
@@ -17,13 +18,16 @@ extern "C" void AWS_LC_fips_failure_callback(char const* message);
 #if defined(FIPS_SELF_TEST_SKIP_ABORT)
 void AWS_LC_fips_failure_callback(char const* message)
 {
-    const size_t char_limit = 128;
+    fprintf(stderr, "AWS_LC_fips_failure_callback invoked with message: '%s'\n", message);
+    // Must track
+    // https://github.com/aws/aws-lc/blob/e4885d5e22f7dc482dd6bfa713b0e1b763b5c538/crypto/fipsmodule/self_check/self_check.c#L52
```
alexw91, line 23: Ideally, AWS-LC would expose this value in a header value somewhere and ACCP would use that value, but this is fine for now.

WillChilds-Klein: agreed.

PR #479, opened 2025-07-15
Remove usage of deprecated SecurityManager and AccessController

chockalingamc:
SecurityManager and AccessController have been deprecated since JDK17 and marked for removal in future.
[JEP-411](https://openjdk.org/jeps/411) has more background info on this.
In JDK21 usage of these classes will throw `UnsupportedOperationException`

Issue  #332

Description of changes: 
Removed usage of AccessController (which uses SecurityManager) and related tests.


csrc/auto_free.h
```diff
@@ -63,7 +63,7 @@
         void releaseOwnership() { PTR_NAME(name) = NULL; }                                                             \
         void clear()                                                                                                   \
         {                                                                                                              \
-            CONCAT2(name, _free)(PTR_NAME(name));                                                                      \
+            CONCAT2(name, _free)(PTR_NAME(name));                                                                       \
```
WillChilds-Klein, line 66: was this intentional?

WillChilds-Klein (approved):

csrc/auto_free.h
chockalingamc: Yeah :( for some reason clang-format + spotless insisted this, which doesn't make sense !

geedo0: I stand by my assertion that clang-format is more trouble than its worth

PR #480, opened 2025-07-15
Fix AccessibleByteArrayOutputStream buffer resize logic

alexw91:
*Issue #, if available:*

*Description of changes:*
Fixes a performance bug in AccessibleByteArrayOutputStream.java. 

Previously when resizing buffers between `INT_MAX/2` and `INT_MAX` we would only grow the capacity to `minimumNewCapacity` instead of doubling the capacity up to `INT_MAX`. This was due to values in that range being left-shifted by 1 bit being interpreted as a negative integer value.

This meant that for ACCP users that perform many small writes, performance would significantly degrade once the buffer capacity reached `INT_MAX/2` as a large portion of time would be spent repeatedly growing the buffer by only a small amount on each write instead of the correct behavior of immediately growing buffer capacity to INT_MAX and not paying the resize penalty on every write.


src/com/amazon/corretto/crypto/provider/AccessibleByteArrayOutputStream.java
```diff
@@ -103,21 +103,44 @@ private void growCapacity(final int newCapacity) {
...
       return;
     }
 
     final byte[] tmp =
         Arrays.copyOf(
-            buf,
-            doNotAllocateMoreThanNeeded
-                ? newCapacity
-                : Math.max(Math.min(limit, buf.length << 1), newCapacity));
+            buf, calculateNewCapacity(buf.length, minimumNewCapacity, doNotAllocateMoreThanNeeded));
```
chockalingamc, line 139: np: can skip passing `buf` since its instance variable.

src/com/amazon/corretto/crypto/provider/AccessibleByteArrayOutputStream.java
```diff
@@ -103,21 +103,44 @@ private void growCapacity(final int newCapacity) {
...
+  }
+
+  private int calculateNewCapacity(
+      final int currentCapacity,
+      final int minimumNewCapacity,
+      final boolean doNotAllocateMoreThanNeeded) {
+    int newCapacity =
+        doNotAllocateMoreThanNeeded
+            ? minimumNewCapacity
+            : Math.max(Math.min(limit, increaseCapacity(currentCapacity)), minimumNewCapacity);
```
chockalingamc: I guess we can achieve the same behavior with 
`Math.max(currentCapacity >= limit>>1 ? limit : currentCapacity<<1, minimumNewCapacity);` 
and just eliminate the `Math.min` that created the bug ?

Basically, we want to 
    `double up a value until a limit` 
and it can be done as 
    `if value >= half the limit, use limit, else use double the value`
without explicitly checking against Integer.MAX_VALUE/2 first and then doing another Math.min comparison with the limit.

src/com/amazon/corretto/crypto/provider/AccessibleByteArrayOutputStream.java
chockalingamc: never mind, i see it helps with the unit tests

tst/com/amazon/corretto/crypto/provider/test/AccessibleByteArrayOutputStreamTest.java
```diff
@@ -62,6 +62,39 @@ public void intMaxOverflow() throws Throwable {
...
+    /* Ensure resizing capacities between INT_MAX/2 and INT_MAX are clamped to INT_MAX */
+    assertNewCapacity(
+        instance, (Integer.MAX_VALUE / 2), (Integer.MAX_VALUE / 2) + 1, Integer.MAX_VALUE);
+    assertNewCapacity(
+        instance,
+        (int) (Integer.MAX_VALUE * 0.75),
+        (int) (Integer.MAX_VALUE * 0.75) + 1,
+        Integer.MAX_VALUE);
+    assertNewCapacity(instance, Integer.MAX_VALUE - 10, Integer.MAX_VALUE - 9, Integer.MAX_VALUE);
+    assertNewCapacity(instance, Integer.MAX_VALUE - 1, Integer.MAX_VALUE, Integer.MAX_VALUE);
```
chockalingamc, line 96: This would pass even with the old logic. Perhaps we should do 
```
assertNewCapacity(instance, Integer.MAX_VALUE - 2, Integer.MAX_VALUE - 1, Integer.MAX_VALUE);
```
to assert the edgecase of the new logic.

alexw91: Reverting the implementation of `increaseCapacity` to a left shift by 1 reproduces the failure that we are trying to avoid, and causes the unit tests to fail. This is because if `currentCapacity >= INT_MAX/2` then the left shift causes the value to become negative instead of being clamped to `INT_MAX`.

```
private int increaseCapacity(int currentCapacity) {
      return currentCapacity << 1;
  }
```

chockalingamc: Right, we cannot just left shift by 1 in `increaseCapacity`. That would just reproduce the old behavior. 
We need to remove `increaseCapacity` method and just use 
```
Math.max(currentCapacity >= limit>>1 ? limit : currentCapacity<<1, minimumNewCapacity);
```
in `calculateNewCapacity`

src/com/amazon/corretto/crypto/provider/AccessibleByteArrayOutputStream.java
alexw91: This logic adds a ternary operator inside another ternary operator. The full logic (including the outer ternary check for `doNotAllocateMoreThanNeeded`) would end up being:

```java
int newCapacity =
    doNotAllocateMoreThanNeeded
        ? minimumNewCapacity
        : Math.max(
            currentCapacity >= limit >> 1 ? limit : currentCapacity << 1, minimumNewCapacity);

return newCapacity;
```

I'd prefer to avoid having doubly-nested ternary operations for readability. I've tested both and they are equivalent, but I think the code that I have right now is more readable.

geedo0 (dismissed):

PR #481, opened 2025-07-16
Improve support for IDEs that can import Gradle projects.

chockalingamc:
Define correct main and test sources, and test dependencies.

Issue: https://sim.amazon.com/issues/ACCP-35


build.gradle
```diff
@@ -914,3 +910,24 @@ task generateEclipseClasspath {
...
+    test {
+        java {
+            srcDirs 'tst'
+        }
+    }
+}
+// The Java plugin's compileJava and compileTestJava tasks will compile the above main and test code
+// into the default 'build/classes' location. Since we use cmake for all compilation, we can skip Gradle's tasks.
+compileJava.enabled = false
+compileTestJava.enabled = false
```
geedo0, line 933: Neat trick, this was my biggest concern around declaring the source sets.

geedo0 (approved):

build.gradle
chockalingamc: Yeah, without this, Gradle will perform an additional compilation of the listed main and test sources. But that didn't affect our build process significantly. Just a few extra seconds to compile and some additional artifacts under `build/classes` which are ignored. But good to avoid all that to begin with.

alexw91 (approved):

PR #484, opened 2025-07-21
Add conditional building of targets to CMake

prasden:
**Description of changes:** This PR enables conditional compilation and inclusion/exclusion of files based on target JDK version. Users can specify their target JDK through a flag, which controls the inclusion of future KEM-related and files that use JDK 17+ APIs for builds targeting JDK 17 or higher. The `jdk17plus` folder was left out of this PR as only KEM stub files were in it to test the conditional build system. 

`CMakeLists.txt` was modified to check if the target JDK version supports KEM to conditionally include/exclude the KEM files when compiling, and `build.gradle` was modified to allow the user to pass in a flag for the desired target runtime.
 
**Usage**:  Run `./gradlew clean build -DTARGET_JDK_VERSION=X` where X can be set to 8, 11, 17, or 21. On default where no flag is passed in, the target is JDK 8.

**Testing**:  Using, `./gradlew clean build_objects -DTARGET_JDK_VERSION=X` the inclusion/exclusion of KEM files in the final JAR files was checked through the command `jar -tf build/cmake/AmazonCorrettoCryptoProvider.jar | grep -i kem` to ensure that KEM stub files were included in JDK 17/21 builds and excluded in JDK 8/11 builds. `TARGET_JDK_VERSION` was set to 21/17/11/8 while on build time JDK 21, and 17/11/8 while on build time JDK 17 while testing and all builds succeeded.

To ensure the target Java version was what was specified through the flag, the command `javap -verbose -cp /path/to/AmazonCorrettoCryptoProvider.jar com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider | head -10` was used to check the "major version" of the resulting jar file, which can be [matched to a corresponding Java version](https://docs.oracle.com/javase/specs/jvms/se24/html/jvms-4.html#jvms-4.1-200-B.2). 


build.gradle
```diff
@@ -196,6 +196,9 @@ dependencies {
 
 defaultTasks 'release'
 
+ext.buildJdk = System.getProperty("java.version").split('\\.')[0] as Integer
```
geedo0: This is incorrect for really old JDKs. e.g. for JDK8, this ends up resolving to "1" because its `java.version` is 1.8.x. You have logic in your PR that checks for JDK8 specifically, so none of that will work. It's only until JDK11 where they adopted a sensible versioning scheme.

nit: You randomly mix single/double quotes in this line. Stick to single quotes unless you need string interpolation or you need to meet a local style convention.

build.gradle
```diff
@@ -330,6 +333,8 @@ task executeCmake(type: Exec) {
     args "-DTEST_RUNNER_JAR=${configurations.testRunner.singleFile}"
     args "-DCMAKE_VERBOSE_MAKEFILE:BOOL=${isVerbose ? 'ON' : 'OFF'}"
     args "-DAWS_LC_VERSION_STRING=" + awsLcGitVersionId
+    args "-DTARGET_JDK_VERSION=" + targetMinJdk
+    args "-DBUILD_JDK_VERSION=" + buildJdk
```
geedo0: Use string interpolation `"-DTARGET_JDK_VERSION=${targetMinJdk}"`

build.gradle
```diff
@@ -196,6 +196,9 @@ dependencies {
 
 defaultTasks 'release'
 
+ext.buildJdk = System.getProperty("java.version").split('\\.')[0] as Integer
+ext.targetMinJdk = System.getProperty('targetMinJdk') ?: project.findProperty('targetMinJdk') ?: '8'
```
geedo0: Please follow the local conventions of the project. This is very important for maintainability, readability, and usability. Don't introduce new conventions even if you disagree with them. Take the time to familiarize yourself with the look/feel of this code base as you're writing the code.

I don't understand why you're trying to support both System and Project properties. We use exclusively Java System properties with the -D flag and these are all named with UPPER_SNAKE_CASE (e.g. ﻿AWSLC_SRC_DIR). 

Add documentation for this new system property to the README.

CMakeLists.txt
```diff
@@ -160,34 +160,99 @@ add_custom_command(
     COMMENT "Generating hash function SPI classes..."
     )
 
+if((TARGET_JDK_VERSION VERSION_LESS "17") OR ((TARGET_JDK_VERSION VERSION_GREATER_EQUAL "18") AND (TARGET_JDK_VERSION VERSION_LESS "21")))
```
geedo0, line 165: The reasoning here is non-obvious and would benefit from an explanatory comment.

CMakeLists.txt
```diff
@@ -160,34 +160,99 @@ add_custom_command(
...
+    set(KEM_SUPPORTED FALSE)
+else()
+    set(KEM_SUPPORTED TRUE)
+endif()
+
+# These two macros below conditionally include/exclude the Jdk17plus based onif the KEM API is supported in the release target
```
geedo0: Am I missing something? Or do these two macros do the same thing but with a different parameter/function names? Refactor this to be more generic, keep your code DRY. You can accept more parameters to support arbitrary filtering based on the passed JDK version. It's not specific to KEMs, a JDK, or even SOURCE/TEST so don't name things as such. You basically just want a macro that can filter an arbitrary list of sources based on a given JDK. If you're going to create macros, you should be writing them in a way that encourages reuse.

CMakeLists.txt
```diff
@@ -160,34 +160,99 @@ add_custom_command(
...
+    filter_kem_sources(ACCP_SRC)
 else()
     file(GLOB_RECURSE ACCP_SRC CONFIGURE_DEPENDS "src/com/amazon/corretto/crypto/provider/*.java")
     file(GLOB_RECURSE ACCP_UTILS_SRC CONFIGURE_DEPENDS "src/com/amazon/corretto/crypto/utils/*.java")
+    filter_kem_sources(ACCP_SRC)
 endif()
-set(ACCP_SRC ${ACCP_SRC} ${ACCP_UTILS_SRC} ${GENERATED_JAVA_SRC})
 
+
+set(ACCP_SRC ${ACCP_SRC} ${ACCP_UTILS_SRC} ${GENERATED_JAVA_SRC})
```
geedo0: Avoid adding spurious code motion changes like this it adds a lot of noise to the PR.

CMakeLists.txt
```diff
@@ -241,6 +306,7 @@ else()
     )
 endif()
 
+# All subsequent Java targets are compatible with Java 8 and beyond
```
geedo0: Why did you move this comment here? The comment itself is also no longer relevant since we're changing the target JDKs.

CMakeLists.txt
```diff
@@ -160,34 +160,99 @@ add_custom_command(
...
+    filter_kem_sources(ACCP_SRC)
 endif()
-set(ACCP_SRC ${ACCP_SRC} ${ACCP_UTILS_SRC} ${GENERATED_JAVA_SRC})
 
+
+set(ACCP_SRC ${ACCP_SRC} ${ACCP_UTILS_SRC} ${GENERATED_JAVA_SRC})
 set(BASE_JAVA_COMPILE_FLAGS ${CMAKE_JAVA_COMPILE_FLAGS} -h "${JNI_HEADER_DIR}" -Werror)
 
+# This macro checks if the build time JDK is 21 above and the target is 17. If so, it uses
+# -source and -target instead of --release to properly support the backported KEM API on JDK 17.
```
geedo0: Can you explain what you mean by "properly support"? Why doesn't the other method work? Why can't we always just use the source/target method to simplify the logic?

CMakeLists.txt
```diff
@@ -160,34 +160,99 @@ add_custom_command(
...
+if(NOT DEFINED TARGET_JDK_VERSION OR TARGET_JDK_VERSION STREQUAL "")
+    set(TARGET_JDK_VERSION "8")
+endif()
+
+set(ORIGINAL_TARGET_JDK_VERSION ${TARGET_JDK_VERSION})
+
 # Java targets defined here are compiled for Java supporting modules
-set(CMAKE_JAVA_COMPILE_FLAGS ${BASE_JAVA_COMPILE_FLAGS} -Xlint --release 9)
+if(TARGET_JDK_VERSION EQUAL 8)
+    set(MODULE_COMPILE_FLAGS ${BASE_JAVA_COMPILE_FLAGS} -Xlint:-options --release 9)
```
geedo0: I'm confused, isn't the point of your `set_conditional_compile_flags` macro supposed to avoid the need for this logic?

CMakeLists.txt
```diff
@@ -160,34 +160,99 @@ add_custom_command(
...
+    if(NOT DEFINED TARGET_JDK_VERSION OR TARGET_JDK_VERSION STREQUAL "")
+        set(TARGET_JDK_VERSION "8")
+    endif()
+    if(NOT DEFINED BUILD_JDK_VERSION OR BUILD_JDK_VERSION STREQUAL "")
+        if(DEFINED Java_VERSION_MAJOR AND NOT Java_VERSION_MAJOR STREQUAL "")
+            set(BUILD_JDK_VERSION ${Java_VERSION_MAJOR})
+        else()
+            set(BUILD_JDK_VERSION "8")
+        endif()
+    endif()
```
geedo0: Can you explain why you're just arbitrarily changing the build JDK like this? The build JDK is just whatever JDK ends up running Gradle, you can't really just change that by overwriting these variables. That's just asking for trouble...

Also, chill out with all this overly cautious checking that this default `TARGET_JDK_VERSION` is not set, you've got at least 2 other places in this PR that set a default values for it. At some point this all just makes the code harder to read/maintain from all the noise.

CMakeLists.txt
prasden: --release does not supported backported APIs when you are on a newer JDK build. On build time JDK 21, --release fails to recognize the KEM API on a 17 runtime. We could modify to always use source/target to make it simpler. However, the tradeoff with always using source/target is that it can cause runtime failures due to it not preventing incompatible APIs, where as --release controls API access which could be safer for runtime.

CMakeLists.txt
prasden: The JDK 8 check is for specifically for setting the release to 9 for module compilation, which requires JDK 9+ features. `set_conditional_compile_flags` is for the code only jar which also handles the special case of a JDK 21 build and 17 runtime.

CMakeLists.txt
prasden: However, the tradeoff with always using source/target is that it can cause runtime failures due to it not preventing incompatible APIs, where as --release controls API access which could be safer for runtime.

CMakeLists.txt
prasden: For the checks on the CI, `./gradlew coverage` command was failing when CMake variables like `TARGET_JDK_VERSION` weren't set, causing javac to get weird malformed arguments and failed the check. I added defensive checks to ensure no empty flags were passed.

CMakeLists.txt
prasden: Added a comment to clarify.

build.gradle
prasden: Edited to follow conventions.

CMakeLists.txt
prasden: Consolidated into a single macro.

CMakeLists.txt
prasden: Consolidated this into a macro to avoid duplicate logic.

CMakeLists.txt
prasden: Resolved this by passing in `BUILD_JDK_VERSION` and `TARGET_JDK_VERSION` to the `coverage_cmake` task in `build.gradle` as well.

CMakeLists.txt
```diff
@@ -160,27 +160,73 @@ add_custom_command(
     COMMENT "Generating hash function SPI classes..."
     )
 
+# The KEM API was introduced in JDK 21 and backported to JDK 17
+# It is not supported on anything below JDK 17 or on JDK 18-20
+if((TARGET_JDK_VERSION VERSION_LESS "17") OR ((TARGET_JDK_VERSION VERSION_GREATER_EQUAL "18") AND (TARGET_JDK_VERSION VERSION_LESS "21")))
+    set(KEM_SUPPORTED FALSE)
```
alexw91: Wanted to take a step back and ask if there's any KEM-specific logic that is needed during our build? Or if all the logic really is just wrapping Java version checks? 

Why not just have a `jdk17plus` directory that is added to the build when the target is JDK 17 or greater? Are we doing anything differently for the KEM API's than for any other API's that might later be added to the `jdk17plus` directory?

README.md
```diff
@@ -441,6 +441,9 @@ Thus, these should all be set on the JVM command line using `-D`.
   to `true` if the keys are only used by other ACCP services AND they are not type cast to `EdECKey`.
   It is worth noting that the key generated by KeyFactory service of SunEC can be used by ACCP services
   such as Signature.
+* `TARGET_JDK_VERSION`
```
alexw91: The rest of these Java system properties are read at runtime in Java code to change ACCP's behavior. I don't think this property belongs in this section.

CMakeLists.txt
alexw91: We might want to rename `KEM_SUPPORTED` to `INCLUDE_JDK17PLUS_DIR` or some other similar name.

prasden: The current logic follows that, where on target is 17, or 21+, the `jdk17plus` directory is added to the build. The variable can be renamed to be more precise, to better represent future additions to the directory.

prasden: Revised variable name.

README.md
prasden: removed

build.gradle
```diff
@@ -196,6 +196,18 @@ dependencies {
 
 defaultTasks 'release'
 
+ext.buildJdk = {
+    def version = System.getProperty("java.version")
```
geedo0: Syle nit: The rest of the file uses the `System.properties['java.version']` subscripting operator rather than the explicit `get()` calls.

build.gradle
```diff
@@ -196,6 +196,18 @@ dependencies {
...
+    def parts = version.split('\\.')
+    if (parts[0] == "1") {
+        return parts[1] as Integer
+    }
```
geedo0: style nit: Use egyption style bracing `} else {` like the rest of the code.

build.gradle
```diff
@@ -196,6 +196,18 @@ dependencies {
 
 defaultTasks 'release'
```
geedo0, line 198: Add documentation to the README on how to build ACCP for JDK17+ and what exactly BUILD/TARGET JDK means.

CMakeLists.txt
```diff
@@ -160,27 +160,73 @@ add_custom_command(
...
-set(CMAKE_JAVA_COMPILE_FLAGS ${BASE_JAVA_COMPILE_FLAGS} -Xlint --release 9)
+# This macro checks if the build time JDK is 21 above and the target is 17. If so, it uses
+# -source and -target instead of --release to properly support the backported KEM API on JDK 17.
+# For JDK 8 targets, it uses --release 9 for module compilation and --release 8 for code compilation.
+# Otherwise, releases on user specified target.
+macro(set_conditional_compile_flags OUTPUT_VAR)
+    set(IS_MODULE_COMPILATION FALSE)
+    if(${ARGC} GREATER 1 AND "${ARGV1}" STREQUAL "MODULE")
+        set(IS_MODULE_COMPILATION TRUE)
+    endif()
```
geedo0: All this makes the code unnecessarily obtuse. For one, this method is not generic/reusable at all. It's extremely specific to these two places in the code where you need to handle module support and conditional JDK17 stuff. It makes it a lot harder to follow the code as a human. The use of variable arguments as a signal that you need to handle module support is extremely odd and not apparent at all in the function signature or comments.

CMakeLists.txt
```diff
@@ -160,27 +160,73 @@ add_custom_command(
...
+endmacro()
+
 # NOTE: CMake introduced the CONFIGURE_DEPENDS feature in 3.12, so condition
 #       its use on CMake version. If the feature is not available, you may need
 #       to `touch CMakeLists.txt` after adding/removing files for them to be
 #       detected by CMake.
 if (${CMAKE_VERSION} VERSION_LESS "3.12.0")
     file(GLOB_RECURSE ACCP_SRC "src/com/amazon/corretto/crypto/provider/*.java")
     file(GLOB_RECURSE ACCP_UTILS_SRC "src/com/amazon/corretto/crypto/utils/*.java")
+    filter_sources_by_condition(ACCP_SRC ".*/jdk17plus/.*" ${INCLUDE_JDK17PLUS_DIR} "${CMAKE_CURRENT_SOURCE_DIR}/src/com/amazon/corretto/crypto/provider/jdk17plus")
```
geedo0, line 191: Does this force us to have `jdk17plus` in the package namespace?

CMakeLists.txt
```diff
@@ -160,27 +160,73 @@ add_custom_command(
...
+        set(IS_MODULE_COMPILATION TRUE)
+    endif()
+    
+    if(TARGET_JDK_VERSION EQUAL 8)
+        if(IS_MODULE_COMPILATION)
+            set(${OUTPUT_VAR} ${BASE_JAVA_COMPILE_FLAGS} -Xlint:-options --release 9)
+        else()
+            set(${OUTPUT_VAR} ${BASE_JAVA_COMPILE_FLAGS} -Xlint:-options --release 8)
+        endif()
+    elseif(INCLUDE_JDK17PLUS_DIR AND TARGET_JDK_VERSION EQUAL 17 AND BUILD_JDK_VERSION GREATER_EQUAL 21)
```
geedo0: What does this logic do that isn't already inherent in `INCLUDE_JDK17PLUS_DIR`?

CMakeLists.txt
prasden: No, it does not need to be part of the package namespace. Everything inside can be part of the same `com.amazon.corretto.crypto.provider` package.

CMakeLists.txt
prasden: `INCLUDE_JDK17PLUS_DIR` only checks if the target is JDK 17 or JDK 21+. This also adds a specific check if the target is specifically 17 and the build time version is 21 or above, to use -source/-target instead of release to support backported APIs. I've removed the redundant condition of checking if `INCLUDE_JDK17PLUS_DIR` is true.

CMakeLists.txt
prasden: I see what you're saying and have modified this macro to be less obfuscated. I've edited the macro to always accept a parameter which is either true or false to indicate whether or not the current compilation is a module compilation, rather than checking for the existence of a second parameter. I've also added descriptive source comments and named the macro parameters better to reflect the functionality.

CMakeLists.txt
```diff
@@ -160,27 +160,73 @@ add_custom_command(
...
-set(CMAKE_JAVA_COMPILE_FLAGS ${BASE_JAVA_COMPILE_FLAGS} -Xlint --release 9)
+# This macro sets the Java compile flags for ACCP based on JDK version and compilation type
+# On JDK 8: uses --release 9 for module compilation as modules are supported in JDK 9+, and uses --release 8 for code
+# JDK 17 with KEM support: when targeting JDK 17 on newer build-time JDKs (21+), uses -source/-target instead of --release.
+# This is because using --release 17 would exclude backported APIs and cause runtime errors when accessing KEM functionality or other JDK 17 APIs.
+# On default, it uses the specified target JDK version.
+# Parameters:
+#   FLAGS_VAR - CMake variable name to store the resulting compiler flags
+#   IS_MODULE - TRUE for module compilation, FALSE for regular code
+macro(set_java_compile_flags FLAGS_VAR IS_MODULE)
```
alexw91: I don't think you need to take in `FLAGS_VAR` if it is always set to `CMAKE_JAVA_COMPILE_FLAGS`. This macro signature can just be `macro(set_java_compile_flags IS_MODULE)`

alexw91 (approved):
I left 1 comment about removing unnecessary macro parameter. Once that fix is made, I'm okay with shipping this PR.

CMakeLists.txt
prasden: Removed and modified macro calls.

alexw91 (dismissed):

README.md
```diff
@@ -442,6 +442,14 @@ Thus, these should all be set on the JVM command line using `-D`.
   It is worth noting that the key generated by KeyFactory service of SunEC can be used by ACCP services
   such as Signature.
 
+## Build Parameters
+
+ACCP supports build-time parameters that control compilation behavior and compatibility. 
+#### Target JDK Version
+The runtime target that ACCP will be targeting on specifies the minimum JDK version required to run the program and determines which language features and APIs can be used. This can be specified for ACCP using the `-DTARGET_JDK_VERSION` flag when compiling ACCP from source.  For example, to target JDK 17 compatibility when compiling with gradle, you would use: <br>
+<br> `./gradlew release -DTARGET_JDK_VERSION=17`.\
```
geedo0, line 450: Again, stick to the local convention. We don't use inline HTML tags anywhere else in this README so you shouldn't either. A simple newline character would suffice here.

PR #485, opened 2025-07-22
Exposed SHA3 through EVP API

havishr:
Changes: Implemented hash_evp_template.cpp.template, which exposes the EVP_MD API. Exposed SHA3_224, SHA3_256, SHA3_384, and SHA3_512. Allows for the use of these algorithms in an isolated context at the moment. 

Choices: Chose to use the specific constants described in the csrc files for each algorithm because they are not all exposed through EVP_MD. Manually copying the context from the Java state to EVP_MD_CTX and vice versa. Is there a better way of doing this? 

Testing: All testing was done the same as the other hash functions. Individual classes for each algorithm that contain KATs and the use HashFunctionTester. Did CAVP testing with the files from https://csrc.nist.gov/projects/cryptographic-algorithm-validation-program/secure-hashing. Had to set a minimum Java version for HashFunctionTester because SUN JDK8 does not expose SHA3 so trying to compare the implementations throws a NoSuchAlgorithmException. 


.project
WillChilds-Klein, line 1: can we remove this from the PR?

.DS_Store
WillChilds-Klein, line 1: can we remove this from the PR?

README.md
```diff
@@ -333,7 +333,7 @@ When changing between FIPS and non-FIPS builds, be sure to do a full `clean` of
 * release: **Default target** depends on build, test, and coverage
 * overkill: Run **all** tests (no coverage)
 * generateEclipseClasspath: Generates a `.classpath` file which is understandable by Eclipse and VS Code to make development easier. (This should ideally be run prior to opening ACCP in your IDE.)
-* single_test: Runs a single unit test. The test is selected with the Java system property `SINGLE_TEST`. For example: `./gradlew single_test -DSINGLE_TEST=com.amazon.corretto.crypto.provider.test.EcGenTest`
+* single_test: Runs a single unit test. The test is selected with the Java system property `SINGLE_TEST`. For example: `./gradlew singleTest -DSINGLE_TEST=com.amazon.corretto.crypto.provider.test.EcGenTest`
```
WillChilds-Klein:
```suggestion
* singleTest: Runs a single unit test. The test is selected with the Java system property `SINGLE_TEST`. For example: `./gradlew singleTest -DSINGLE_TEST=com.amazon.corretto.crypto.provider.test.EcGenTest`
```

build-tools/bin/generate-java-hash-spi
```diff
@@ -8,9 +8,9 @@ mode=$1
 src=$2
 path=$3
 
-for i in SHA-512 SHA-384 SHA-256 SHA-1 MD5; do
+for i in SHA-512 SHA-384 SHA-256 SHA-1 MD5 SHA3-224 SHA3-256 SHA3-384 SHA3-512; do
     CODENAME=`echo $i | tr -d "-"`
-
+
```
WillChilds-Klein:
```suggestion

```

tst/com/amazon/corretto/crypto/provider/.DS_Store
WillChilds-Klein, line 1: can we remove this from the PR?

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -86,6 +86,10 @@ private void buildServiceMap() {
     addService("MessageDigest", "SHA-256", "SHA256Spi");
     addService("MessageDigest", "SHA-1", "SHA1Spi");
     addService("MessageDigest", "MD5", "MD5Spi");
+    addService("MessageDigest", "SHA3-224", "SHA3224Spi");
+    addService("MessageDigest", "SHA3-256", "SHA3256Spi");
+    addService("MessageDigest", "SHA3-384", "SHA3384Spi");
+    addService("MessageDigest", "SHA3-512", "SHA3512Spi");
```
WillChilds-Klein, line 100: note to reviewers -- i confirmed that these 4 service names align with JCA-standard names for SHA3

https://docs.oracle.com/en/java/javase/11/docs/specs/security/standard-names.html

tst/com/amazon/corretto/crypto/provider/test/BaseSHATest.java
```diff
@@ -0,0 +1,130 @@
...
+  protected abstract String getNullDigest();
+
+  protected abstract String getTestVector();
+
+  protected abstract String getCavpShortFile();
+
+  protected abstract String getCavpLongFile();
+
+  protected MessageDigest getDigest() throws Exception {
+    System.out.print(getAlgorithm());
```
WillChilds-Klein: can we drop this print statement?

csrc/hash_evp_template.cpp.template
```diff
@@ -0,0 +1,243 @@
...
+#define TRANSFORM_sha512 SHA512
+#define TRANSFORM_sha512_224 SHA512_224
+#define TRANSFORM_sha512_256 SHA512_256
+#define TRANSFORM_sha1 SHA1
+#define TRANSFORM_md5 MD5
+#define TRANSFORM_md4 MD4
+#define TRANSFORM_ripemd160 RIPEMD160
+#define TRANSFORM_shake128 SHAKE128
+#define TRANSFORM_shake256 SHAKE256
+#define TRANSFORM_blake2b256 BLAKE2B256
```
WillChilds-Klein: why do we need to define these for algorithms that we don't implement with this template?

csrc/hash_evp_template.cpp.template
```diff
@@ -0,0 +1,243 @@
...
+// OPENSSL_EXPORT const EVP_MD *EVP_sha384(void);
+// OPENSSL_EXPORT const EVP_MD *EVP_sha512(void);
+// OPENSSL_EXPORT const EVP_MD *EVP_sha512_224(void);
+// OPENSSL_EXPORT const EVP_MD *EVP_sha512_256(void);
+// OPENSSL_EXPORT const EVP_MD *EVP_sha3_224(void);
+// OPENSSL_EXPORT const EVP_MD *EVP_sha3_256(void);
+// OPENSSL_EXPORT const EVP_MD *EVP_sha3_384(void);
+// OPENSSL_EXPORT const EVP_MD *EVP_sha3_512(void);
+// OPENSSL_EXPORT const EVP_MD *EVP_shake128(void);
+// OPENSSL_EXPORT const EVP_MD *EVP_shake256(void);
```
WillChilds-Klein: can we remove these comments?

WillChilds-Klein (commented):
Looking good so far!

csrc/hash_evp_template.cpp.template
havishr: Its there for future functionality if needed, but it isn't part of this implementation, so I can remove it.

csrc/hash_evp_template.cpp.template
```diff
@@ -0,0 +1,207 @@
...
+{
+    try {
+        raii_env env(pEnv);
+
+        CTX ctx;
+        
+        java_buffer contextBuffer = java_buffer::from_array(env, contextArray);
+
+        {
+            CHECK_OPENSSL(EVP_DigestInit(&ctx, MD()));
```
WillChilds-Klein: is this init necessary? we will copy into `ctx.md_data` on L72

csrc/hash_evp_template.cpp.template
```diff
@@ -0,0 +1,207 @@
...
+    JNIEnv* pEnv, jclass, jbyteArray contextArray, jbyteArray digestArray, jint offset)
+{
+    try {
+        raii_env env(pEnv);
+        
+        CTX ctx;
+
+        java_buffer contextBuffer = java_buffer::from_array(env, contextArray);
+        {            
+            CHECK_OPENSSL(EVP_DigestInit(&ctx, MD()));
```
WillChilds-Klein: same question about whether we need this here and on L148

csrc/hash_evp_template.cpp.template
```diff
@@ -0,0 +1,207 @@
...
+
+JNIEXPORT void JNICALL JNI_NAME(finish)(
+    JNIEnv* pEnv, jclass, jbyteArray contextArray, jbyteArray digestArray, jint offset)
+{
+    try {
+        raii_env env(pEnv);
+        
+        CTX ctx;
+
+        java_buffer contextBuffer = java_buffer::from_array(env, contextArray);
```
WillChilds-Klein: here and elsewhere, can we add a comment explaining why we use `java_buffer` here while `hash_template.cpp.template` uses `bounce_buffer`s?

csrc/hash_evp_template.cpp.template
havishr: The init is necessary. Trying to copy without the init causes a segfault.

csrc/hash_evp_template.cpp.template
havishr: Same answer as above

.project
geedo0, line 1: While I don't think anyone here (or anywhere) uses Eclipse anymore, I don't think this PR is the place to remove it.

csrc/hash_evp_template.cpp.template
```diff
@@ -0,0 +1,211 @@
+#include "openssl/base.h"
+#include "buffer.h"
+#include "env.h"
+#include "generated-headers.h"
+#include "util.h"
+
+
```
geedo0: Style nit: Here and elsewhere in this file there is superfluous whitespace. Try to stick to just 1 newline, there's really no benefit for using more than that.

csrc/hash_evp_template.cpp.template
```diff
@@ -0,0 +1,211 @@
...
+// Transform DIGEST_NAME from "sha3_224" format to "SHA3224" format for JNI class names
+// This macro needs to be defined for each digest algorithm
```
geedo0: This seems like an anti-pattern for what should be a template. You're basically requiring that we make a change to the template each time you need some new algorithm. This should be pulled up.

WillChilds-Klein (dismissed):

csrc/sha3_224.cpp
```diff
@@ -0,0 +1,6 @@
+#include <openssl/digest.h>
+#define DIGEST_NAME sha3_224 
+#define DIGEST_LENGTH 28
+#define DIGEST_BLOCK_SIZE 144
+#define MD_CTX_SIZE 400
```
geedo0: It would be good to define these constants somewhere, like a comment in the template.

Where did you get the DIGEST_BLOCK_SIZE values from?

MD_CTX_SIZE seems like it would be the same for all the hashes right? Maybe pull this down into the template layer. Better yet, replace it with a `sizeof` call on the actual type to avoid any drift between LC and this template..

csrc/hash_evp_template.cpp.template
```diff
@@ -0,0 +1,211 @@
...
+    JNIEnv* pEnv, jclass, jbyteArray digestArray, jbyteArray dataArray, jint bufOffset, jint dataLength)
+{
+    // As this method needs to be extremely high speed, we are omitting use of java_buffer
+    // to avoid the extra JNI calls it requires. Instead we are trusting that dataLength
+    // is correct.
+    try {
+        raii_env env(pEnv);
+        
+        SecureBuffer<CTX,1> ctx;
+        const size_t scratchSize = DIGEST_BLOCK_SIZE; // Size is arbitrarily chosen
```
geedo0: 🤔 ?

tst/com/amazon/corretto/crypto/provider/test/BaseSHATest.java
```diff
@@ -0,0 +1,129 @@
...
+
+    assertArrayEquals(Hex.decodeHex(getTestVector().toCharArray()), digest.digest());
+  }
+
+  @Test
+  public void testFastPath() throws Exception {
+    MessageDigest digest = getDigest();
+
+    assertArrayEquals(
+        Hex.decodeHex(getTestVector().toCharArray()), digest.digest("testing".getBytes()));
```
geedo0: This is an odd way to lay things out. We should factor out the pre-image "testing" similar to the "TestVector" otherwise, implementers of this class need to implicitly know that their pre-image must always be the String "testing".

tst/com/amazon/corretto/crypto/provider/test/BaseSHATest.java
```diff
@@ -0,0 +1,129 @@
...
+
+  @Test
+  public void cavpShortVectors() throws Throwable {
+    try (final InputStream is = new GZIPInputStream(TestUtil.getTestData(getCavpShortFile()))) {
+      new HashFunctionTester(getAlgorithm()).test(RspTestEntry.iterateOverResource(is));
+    }
+  }
+
+  @Test
+  public void cavpLongVectors() throws Throwable {
```
geedo0: These are both the same. You could further parameterize these to accept a list of input files that get fed into the same cavp test case.

tst/com/amazon/corretto/crypto/provider/test/SHA3_384Test.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class SHA3_384Test extends BaseSHATest {
+
+  private static final String ALGORITHM = "SHA3-384";
+  private static final String NULL_DIGEST =
+      "0c63a75b845e4f7d01107d852e4c2485c51a50aaaa94fc61995e71bbee983a2ac3713831264adb47fb6bd1e058d5f004";
+  private static final String TEST_VECTOR =
+      "e15a44d4e12ac138db4b8d77e954d78d94de4391ec2d1d8b2b8ace1a2f4b3d2fb9efd0546d6fcafacbe5b1640639b005";
+  private static final String CAVP_SHORT_FILE = "SHA3_384ShortMsg.rsp.gz";
+  private static final String CAVP_LONG_FILE = "SHA3_384LongMsg.rsp.gz";
```
geedo0: Personally, I'd inline these. Don't like the additional layer of indirection here.

PR #487, opened 2025-08-06
add ML-KEM with only self-tests

prasden:
This PR adds ML-KEM support to ACCP by implementing KEM base classes (`EvpKemKey, EvpKemPrivateKey, EvpKemPublicKey`) and ML-KEM operations (`MlKemSpi, MlKemGen`) using JNI bindings from AWS-LC's native ML-KEM implementation. The implementation supports all three ML-KEM parameter sets (512, 768, 1024) with full encapsulation/decapsulation functionality and includes test coverage for key generation, encapsulation, decapsulation, parameter set and ciphertext size validation. 

BouncyCastle interoperability testing was intentionally excluded since AWS-LC does not yet support ASN.1 encoding for ML-KEM keys, requiring raw key formats for now. Testing was done through `MlKemTest.java,` where `shouldRegisterMLKEM` was set to true in `AmazonCorrettoCryptoProvider.java,` using `./gradlew clean singleTest -DSINGLE_TEST=com.amazon.corretto.crypto.provider.test.MlKemTest -DTARGET_JDK_VERSION=17.`  All testing was performed exclusively with ACCP as the provider.


csrc/java_evp_keys.cpp
```diff
@@ -714,4 +714,62 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpMlDsaPr
...
+JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpKemPrivateKey_getRawPrivateKey(
+    JNIEnv* pEnv, jclass, jlong keyHandle)
+{
+    jbyteArray result = NULL;
+    try {
+        raii_env env(pEnv);
+        EVP_PKEY* key = reinterpret_cast<EVP_PKEY*>(keyHandle);
+
+        size_t keyLen;
+        CHECK_OPENSSL(EVP_PKEY_get_raw_private_key(key, NULL, &keyLen) == 1);
```
alexw91: `CHECK_OPENSSL` already checks that Openssl's return value is 1 on success and 0 on failure, you don't need these extra `== 1` statements used here and elsewhere.

csrc/mlkem_spi.cpp
```diff
@@ -0,0 +1,117 @@
...
+            throw_java_ex(EX_RUNTIME_CRYPTO, "Failed to create EVP context");
+        }
+        
+        size_t ciphertext_len, shared_secret_len;
+        CHECK_OPENSSL(EVP_PKEY_encapsulate(ctx, NULL, &ciphertext_len, NULL, &shared_secret_len));
+        
+        int paramSet = ciphertextLengthToParameterSet(ciphertext_len);
+        if (paramSet == -1) {
+            throw_java_ex(EX_RUNTIME_CRYPTO, "Unknown ML-KEM parameter set");
+        }
```
alexw91: This check is likely unnecessary. You are checking if `paramSet == -1` here in native code, and if it equals -1 you throw an exception. However, in the corresponding Java code you are doing the opposite conversion, catching the exception and converting it to -1. You should throw the exception, but probably from the Java code, not from here.

src/com/amazon/corretto/crypto/provider/EvpKemKey.java
```diff
@@ -0,0 +1,42 @@
...
+                kemUtilsClass.getDeclaredMethod("nativeGetParameterSet", long.class);
+            method.setAccessible(true);
+            result = use(ptr -> (Integer) method.invoke(null, ptr));
+          } catch (Exception e) {
+            result = -1;
+          }
+          parameterSet = result;
+        }
+      }
+    }
```
alexw91: 1. This code can probably be called once in the constructor and cached, instead of lazily loaded at runtime. This would avoid the need for the extra thread synchronization code since only 1 thread can call the constructor.
2. This would also allow you to immediately throw an exception from the constructor if the pointer inside InternalKey is invalid and doesn't map to a valid MLKEM parameter size. 

As it is now, I think if you pass in an invalid pointer `getParameterSet()` might return `-1`, and as a result `getAlgorithm()` might return the string `ML-KEM--1`

src/com/amazon/corretto/crypto/provider/EvpKemKey.java
```diff
@@ -0,0 +1,42 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider;
+
+public abstract class EvpKemKey extends EvpKey {
+
+  private volatile Integer parameterSet;
```
alexw91: I think a lot of places where you are using `Integer parameterSet` might be better served by a new `MlkemParameter` Java Enum with values of `MLKEM_512`, `MLKEM_768`, and `MLKEM_1024`. 

Something along the lines of:
```
enum MlkemParameter {
   MLKEM_512(...),
   MLKEM_768(...),
   MLKEM_1024(...);
   
   static MlkemParameter fromParameterSize(int parameterSet) { ... }
   
   int getPublicKeySize() { ... }
   int getSecretKeySize() { ... }
   int getCiphertextSize() { ... }
   int getParameterSize() { ... }
}
```

src/com/amazon/corretto/crypto/provider/MlKemGen.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.security.InvalidAlgorithmParameterException;
+import java.security.KeyPair;
+import java.security.KeyPairGeneratorSpi;
+import java.security.SecureRandom;
+import java.security.spec.AlgorithmParameterSpec;
+
+class MlKemGen extends KeyPairGeneratorSpi {
+  private final AmazonCorrettoCryptoProvider provider_;
+  private int parameterSet = -1;
```
alexw91: Same comment as elsewhere, use a Java enum for the parameter size, not a raw int or Integer.

src/com/amazon/corretto/crypto/provider/jdk17plus/KemUtils.java
```diff
@@ -0,0 +1,136 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider;
+
+import java.security.InvalidAlgorithmParameterException;
+import java.security.spec.AlgorithmParameterSpec;
+import java.security.spec.NamedParameterSpec;
+
+/** Utility class containing ML-KEM constants from aws-lc/crypto/fipsmodule/ml_kem/ml_kem.h . */
+public final class KemUtils {
```
alexw91: Why does this class need to be in the `jdk17plus` directory?

prasden: When running on an older target such as JDK 8,  a runtime error is thrown due to`NamedParameterSpec` not being found as it was introduced in JDK 11.

csrc/java_evp_keys.cpp
prasden: was following convention from Ed keys: https://github.com/corretto/amazon-corretto-crypto-provider/blob/d1a2ddb7397633e9185ac4dd74a8afa055e13b92/csrc/java_evp_keys.cpp#L347 
Follow up, can I edit EdKey to consolidate into a single `getRawKey` function for both key types and classes?

csrc/mlkem_spi.cpp
prasden: Changed in next commit to throw exception on Java side.

src/com/amazon/corretto/crypto/provider/EvpKemKey.java
prasden: Changed the check to be called in the constructor and stored the result.

src/com/amazon/corretto/crypto/provider/EvpKemKey.java
prasden: Added MlKemParameter

src/com/amazon/corretto/crypto/provider/EvpKemKey.java
```diff
@@ -0,0 +1,51 @@
...
+  private static final long serialVersionUID = 1;
+
+  private static native int getParameterSet(long ptr);
+
+  EvpKemKey(final InternalKey key, final boolean isPublicKey) {
+    super(key, EvpKeyType.KEM, isPublicKey);
+    this.parameterSet = initializeParameterSet();
+  }
+
+  private MlKemParameter initializeParameterSet() {
```
alexw91: 1. I don't think this function should return null. We should throw an exception instead and avoid setting values to null whenever possible. We should be aiming to [make invalid states unrepresentable](https://geeklaunch.io/blog/make-invalid-states-unrepresentable/). 
   - I think that the EvpKemKey constructor should either succeed and have all parameters contain valid values, or the EvpKemKey constructor should fail with an exception. That way you can remove the various null checks that you currently have.

2. I think this function should be moved into the `MlKemParameter` enum class, and be renamed to `public static MlKemParameter fromInternalKey(InternalKey internalKey)`. Then inside this function, instead of using a bare `use()` call, you can call `internalKey.use(ptr -> (Integer) method.invoke(null, ptr));`. 
     - This would make it clearer that this function is converting an `InternalKey` into a `MlKemParameter`, and allows this function to be more easily reused if there are other places where that conversion needs to happen.

src/com/amazon/corretto/crypto/provider/EvpKemKey.java
```diff
@@ -0,0 +1,51 @@
...
+          "ML-KEM parameter set not available (JDK8 compatibility mode)");
+    }
+    return parameterSet;
+  }
+
+  @Override
+  public String getAlgorithm() {
+    if (parameterSet == null) {
+      return "ML-KEM"; // Generic algorithm name when parameter set unknown
+    }
```
alexw91: Is there a case where this is expected to happen?

src/com/amazon/corretto/crypto/provider/jdk17plus/MlKemSpi.java
```diff
@@ -0,0 +1,177 @@
...
+import javax.crypto.KEMSpi;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
+
+abstract class MlKemSpi implements KEMSpi {
+
+  protected final MlKemParameter parameterSet;
+  protected final int publicKeySize;
+  protected final int privateKeySize;
+  protected final int ciphertextSize;
```
alexw91: These can be deleted. You can call `parameterSet.get___Size()` instead

src/com/amazon/corretto/crypto/provider/MlKemParameter.java
```diff
@@ -0,0 +1,54 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider;
+
+public enum MlKemParameter {
+  // (parameterSize, publicKeySize, secretKeySize, ciphertextSize) - constants defined in AWS-LC
+  MLKEM_512(512, 800, 1632, 768),
```
alexw91: You can add the EVP_NID values from https://github.com/aws/aws-lc/blob/main/include/openssl/nid.h#L4346-L4356 and add a `.getEvpNid()` function.

src/com/amazon/corretto/crypto/provider/EvpKeyType.java
```diff
@@ -19,7 +19,8 @@ enum EvpKeyType {
   RSA("RSA", 6, RSAPublicKey.class, RSAPrivateKey.class),
   EC("EC", 408, ECPublicKey.class, ECPrivateKey.class),
   EdDSA("EdDSA", 949, PublicKey.class, PrivateKey.class),
-  MLDSA("ML-DSA", 993, PublicKey.class, PrivateKey.class);
+  MLDSA("ML-DSA", 993, PublicKey.class, PrivateKey.class),
+  KEM("ML-KEM", 988, PublicKey.class, PrivateKey.class);
 
   final String jceName;
   final int nativeValue;
```
alexw91, line 28: `nativeValue` should be renamed to `evpKeyNidValue` for clarity. I thought this was a pointer value on my first read of the code.

src/com/amazon/corretto/crypto/provider/EvpKeyType.java
```diff
@@ -19,7 +19,8 @@ enum EvpKeyType {
   RSA("RSA", 6, RSAPublicKey.class, RSAPrivateKey.class),
   EC("EC", 408, ECPublicKey.class, ECPrivateKey.class),
   EdDSA("EdDSA", 949, PublicKey.class, PrivateKey.class),
-  MLDSA("ML-DSA", 993, PublicKey.class, PrivateKey.class);
+  MLDSA("ML-DSA", 993, PublicKey.class, PrivateKey.class),
+  KEM("ML-KEM", 988, PublicKey.class, PrivateKey.class);
```
alexw91: Now that I've had time to more thoroughly read this code, unfortunately I think you will probably end up needing to create 1 enum value per key size (MLKEM_512, MLKEM_768, MLKEM_1024), each with their own unique EVP NID value. 

This will allow you to re-use
1. Decoding with [EvpKeyFactory.pkcs82Evp()](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/EvpKeyFactory.java#L35) 
2. Encoding with [EvpKey.encodePublicKey()](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/EvpKey.java#L51) and [EvpKey.encodePrivateKey()](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/EvpKey.java#L53)

But you probably won't need to do this until AWS-LC supports https://datatracker.ietf.org/doc/draft-ietf-lamps-kyber-certificates/. 

Otherwise, if we want to have only 1 enum value for MLKEM in this EvpKeyType Java class, AWS-LC will need to create a generic NID value for MLKEM like they do for the other EVP key types, and you can use that generic EVP NID here.

src/com/amazon/corretto/crypto/provider/EvpKemKey.java
prasden: Added this generic case when this file is compiled as part of the target but is not used for any KEM operations. Will remove this and catch an exception instead.

PR #489, opened 2025-08-29
Add support for ML-KEM

prasden:
This PR adds ML-KEM support to ACCP by implementing KEM base classes and ML-KEM from AWS-LC. The implementation supports all three ML-KEM parameter sets (512, 768, 1024) with full encapsulation/decapsulation functionality.

The [KEM](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/javax/crypto/KEM.html) and [KemSpi](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/javax/crypto/KEMSpi.html) APIs are only supported in JDK 17 and 21+, requiring the `TARGET_JDK_VERSION` to be set to JDK 17 or 21+ when building ACCP for ML-KEM.  

ACCP can only export ML-KEM keys in `expandedKey` format as per the [draft standard](https://www.ietf.org/archive/id/draft-ietf-lamps-dilithium-certificates-12.html#appendix-C.2), as currently AWS-LC does not support exporting KEM keys in seed format. This means that Bouncy Castle can import all ACCP keys but ACCP can only import Bouncy Castle ML-KEM keys that are in `expandedKey` format. 

Unit tests are written in`MlKemTest.java`, and were verified using  `./gradlew clean singleTest -DSINGLE_TEST=com.amazon.corretto.crypto.provider.test.MlKemTest -DTARGET_JDK_VERSION=17.`  `MlKemTest.java` tests ML-KEM key generation, encapsulation/decapsulation, parameter set validation, valid ciphertext sizes and shared secret generation. It also verifies compatibility with BouncyCastle through ACCP's ML-KEM KeyFactory.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #494, opened 2025-09-10
Update test to handle a bug in Signature.verify() implementation by SunEC EdDSASignature until JDK 21

chockalingamc:
*Issue #332 *

*Description of changes:*

According to the `Signature.verify()` [interface](https://github.com/openjdk/jdk/blob/jdk-17%2B35/src/java.base/share/classes/java/security/Signature.java#L782-L786), if anything is wrong with the signature the implementation should throw a `SignatureException`, which should ideally apply for the case of `null` signature as well.

But until JDK 20, the `sun.security.ec.ed.EdDSASignature` [implementation](https://github.com/openjdk/jdk/blob/jdk-20%2B35/src/jdk.crypto.ec/share/classes/sun/security/ec/ed/EdDSASignature.java#L213-L215) skipped signature verification and returned a `false` instead, when its `message` object was null (default value until assigned via an update() call).
Bug report: https://bugs.openjdk.org/browse/JDK-8300399

In JDK 21, the [implementation](https://github.com/openjdk/jdk/blob/jdk-21%2B35/src/jdk.crypto.ec/share/classes/sun/security/ec/ed/EdDSASignature.java#L213-L215) was [fixed](https://github.com/openjdk/jdk/commit/b317658d69a477df04ded3cc2e107970f8a6e20d) to perform verification for empty (null) message, and thereby comply with Signature.verify() interface to throw `SignatureException` for a `null` signature.


chockalingamc:
> A similar change was made in the ML-KEM PR: https://github.com/corretto/amazon-corretto-crypto-provider/pull/489/files#diff-d021808ce2d8389cda58179eb4c3b83ffe854d3f8346e127ad958860f051a170

Thanks for that ! I captured some more context to this change in my comments, and an assert on the exception message. Could we merge this in and pull in the update to the ML-KEM PR ?

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
```diff
@@ -481,12 +482,24 @@ public void testNullInputs() throws GeneralSecurityException {
...
+      // Bug report: https://bugs.openjdk.org/browse/JDK-8300399
+      assertFalse(result);
     } catch (SignatureException e) {
-      // JDK 21 throws SignatureException for null signature
+      // In JDK 21, the implementation was fixed to perform verification for empty (null) message
+      // https://github.com/openjdk/jdk/commit/b317658d69a477df04ded3cc2e107970f8a6e20d
+      // and thereby comply with Signature.verify() interface to throw SignatureException for a
+      // null signature
+      // https://github.com/openjdk/jdk/blob/jdk-21%2B35/src/jdk.crypto.ec/share/classes/sun/security/ec/ed/EdDSASignature.java#L213-L215
+      assertEquals("signature was null", e.getMessage());
```
WillChilds-Klein, line 502: nitpick: we can also assert java version expectations here and after 495.

PR #497, opened 2025-09-18
Update README.md w/ ML-KEM

WillChilds-Klein:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ KeyAgreement:
 * ECDH
 
 KEM algorithms:
-* ML-KEM
+* ML-KEM (JDK 17+ only)
 
 SecretKeyFactory:
 * HkdfWithHmacSHA1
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -92,6 +92,9 @@ KeyGenerator:
 KeyAgreement:
 * ECDH
 
+KEM algorithms:
+* ML-KEM (JDK 17+ only)
```
geedo0: Pedantic - This won't work on JDK 18/19/20 despite being 17+

Should we have documentation explicitly calling out that the `TARGET_JDK_VERSION` property needs to be set at build time for this to work? i.e. Running ACCP from Maven on JDK17 won't work.

WillChilds-Klein: Mm, that's a good point. Will update.

alexw91 (approved):

Issue #493, opened 2025-09-08
AES KW support


sergey-gubenko:
Add support for the AesKeyWrapSpi.
AES KW is defined in [RFC-3394](https://datatracker.ietf.org/doc/html/rfc3394).
This new cipher should be initialized as AES with KW and NoPadding. 

Current ACCP support AES key wrap with padding:

https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/AesKeyWrapPaddingSpi.java

I have source code for the above that I can contribute available.

sergey-gubenko:
Thanks, I would say it's largely a derivative, the KW logic is similar to what KWP does today, except for the actual key wrap unwrap logic, I'm using `AES_wrap_key` instead of `AES_wrap_key_padded`, test cases are also different.

sergey-gubenko:
Hi Gerardo, can you, please, take a look at the AES KW PR when you have a chance? 

https://github.com/corretto/amazon-corretto-crypto-provider/pull/498

sergey-gubenko:
Hi Gerardo, thanks a lot for taking a look at the PR, how do you want me to resolve the comments, in a separate commit or update the existing ones?

sergey-gubenko:
Hi Gerardo, sorry for basic questions, what is the right process here, are we waiting for someone else from your team to review this?

geedo0:
Confirmed that one of my colleagues is in the midst of reviewing it now. We need two approvals to merge. Do you need a new release of ACCP into Maven after this is merged?

PR #498, opened 2025-09-19
Aes wrap not padded

sergey-gubenko:
*Issue #493*

*Description of changes:*

This PR adds a new CipherSpi implementation, AesKeyWrapSpi. This
cipher can be instantiated as AES with mode KW, and NoPadding. AWS-LC
only provides one-shot functions for key wrapping and unwrapping, so we
buffer any incremental data fed to the Cipher until finalization, at
which point we pass all buffered data over the JNI and call AWS-LC. AES
KW is defined in [RFC-3394][1], which we rely on for output buffer
sizing and other implementation-adjacent details/input validation.

As described in [this document][2], NIST provides Known Answer Tests for
AES KW. We pull down [these test vectors][3] and include `gzip`'s of
them in the repo. We then add AesKwKatTest to read these test cases in,
and execute them against ACCP. This commit follows prior art for KAT
tests in AesKwpKatTest.

[1]: https://datatracker.ietf.org/doc/html/rfc3394
[2]: https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Algorithm-Validation-Program/documents/mac/KWVS.pdf
[3]: https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Algorithm-Validation-Program/documents/mac/kwtestvectors.zip

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,9 @@ KeyGenerator:
 KeyAgreement:
 * ECDH
 
+KEM algorithms:
+* ML-KEM (JDK 17+ LTS, also see [ML-KEM Considerations](#ml-kem-considerations))
+
 SecretKeyFactory:
 * HkdfWithHmacSHA1
 * HkdfWithHmacSHA256
@@ -124,6 +127,14 @@ Mac algorithms with precomputed key and associated secret key factories (expert 
 * HmacSHA256WithPrecomputedKey
 * HmacSHA1WithPrecomputedKey
 * HmacMD5WithPrecomputedKey
+
+### ML-KEM Considerations
+
+JDK's [KEM interface](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/javax/crypto/KEM.html) was backported to JDK17, but not to earlier JDK versions. ACCP JARs support all LTS JDK versions since JDK8, so **ACCP's default build configuration and maven artifacts** omit ML-KEM as it relies on the KEM interface. To enable ML-KEM in locally built JARs, you'll need to build with a JDK ≥17 and specify `TARGET_JDK_VERSION` ≥17 like so:
+
+```
+./gradlew -DTARGET_JDK_VERSION=17 build
+``` 
 
 # Notes on ACCP-FIPS
 ACCP-FIPS is a variation of ACCP which uses AWS-LC-FIPS 2.x as its cryptographic module. This version of AWS-LC-FIPS has FIPS certificate [4816](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4816).


src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -92,6 +92,7 @@ private void buildServiceMap() {
     addService("Cipher", "AES_128/GCM/NoPadding", "AesGcmSpi");
     addService("Cipher", "AES_256/GCM/NoPadding", "AesGcmSpi");
 
+    addService("Cipher", "AES/KW/NoPadding", "AesKeyWrapSpi", /*attributes*/ null, "AesWrap");
     addService(
         "Cipher", "AES/KWP/NoPadding", "AesKeyWrapPaddingSpi", /*attributes*/ null, "AesWrapPad");
```
geedo0, line 97: Not your concern - What a sad naming convention for this Cipher Transformation 😞

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
```diff
@@ -0,0 +1,346 @@
...
+    }
+  }
+
+  @Override
+  protected void engineInit(int opmode, Key key, SecureRandom ignored) throws InvalidKeyException {
+    implInit(opmode, key, null);
+  }
+
+  @Override
+  protected void engineInit(int opmode, Key key, AlgorithmParameters params, SecureRandom ignored2)
```
geedo0: nit: We can drop the `2` here from `ignored2` and on line 149.

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
```diff
@@ -0,0 +1,346 @@
...
+      }
+      this.jceKey = (SecretKey) key;
+    }
+    this.opmode = opmode;
+    this.iv = iv;
+    this.buffer.reset();
+  }
+
+  private void validateIv(byte[] iv) throws InvalidAlgorithmParameterException {
+    if (iv != null && iv.length < 8) {
```
geedo0: The constraint should be `iv.length != 8`. IVs of any other length are not valid.

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
geedo0, line 1: When reviewing, I diffed this against `AesKeyWrapPaddingSpi.java` and can confirm that this was largely a copy/paste of the padded version of the algorithm with modifications to support the unwrapped version.

```
diff -u src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java src/com/amazon/corretto/crypto/provider/AesKeyWrapPaddingSpi.java
```

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
```diff
@@ -0,0 +1,346 @@
...
+    return out;
+  }
+
+  private int implDoFinal(byte[] in, int inOffset, int inLen, byte[] out, int outOffset)
+      throws BadPaddingException {
+    implUpdate(in, inOffset, inLen);
+
+    final int outLen;
+    try {
+      if ((buffer.size() % 8) != 0) {
```
geedo0, line 273: We must also check that `buffer.size() >= 16` as per the RFC's minimum length requirement.

```
   The only restriction the key wrap algorithm places on n is that n be
   at least two.  
```

csrc/aes_kw.cpp
geedo0, line 1: Confirmed that this is based on `aes_kwp.cpp` with minor differences to handle the IV parameter and different interface.

```
diff -u csrc/aes_kw.cpp csrc/aes_kwp.cpp
```

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,550 @@
...
+          24,
+          32, // AES keys
+          512,
+          1024,
+          2048,
+          4096 // RSA keys
+          );
+
+  public static List<Arguments> getParamsGeneric() {
+    final int[] ivSizes = {0, 8};
```
geedo0: style nit: define the IV sizes like the other parameters, as a `private static final` outside this function's scope. It should also be the outermost part of the loop just to align with the order of the args (no functional difference, just style).

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,550 @@
...
+    final byte[] plaintext = c.doFinal();
+    assertArraysHexEquals(secret, plaintext);
+  }
+
+  @Test
+  public void testNativeProviderAlias() throws Exception {
+    // this test asserts that all expected aliases for the AES KW cipher
+    // are adequatly supplied by the native provider
+    for (String alias : KW_CIPHER_ALIASES) {
+      Cipher.getInstance(alias, TestUtil.NATIVE_PROVIDER);
```
geedo0: You can also add an additional assert here that checks that you did in fact get this AES-KW implementation. Not sure if `Cipher.getAlgorithm()` returns the full transform or simply the alias used to fetch it. Not important if this poses any amount of trouble.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,550 @@
...
+    }
+  }
+
+  @Test
+  public void testEngineGetOtputSize() throws Exception {
```
geedo0: typo

```suggestion
  public void testEngineGetOutputSize() throws Exception {
```

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,550 @@
...
+    for (int inputSize : inputSizes) {
+      c.update(new byte[inputSize]);
+      bytesBuffered += inputSize;
+      assertTrue(c.getOutputSize(0) == bytesBuffered + 8);
+    }
+    int finalOutputSize = c.getOutputSize(0);
+    byte[] ciphertext = c.doFinal();
+    assertEquals(ciphertext.length, finalOutputSize);
+
+    c.init(Cipher.DECRYPT_MODE, kek);
```
geedo0: Generally, I prefer creating a new Cipher object rather than re-using it for a new mode. It's probably fine since the tests are passing, but it's more controlled and consistent. Applies to here and the subsequent tests.

```suggestion
    c = getCipher(TestUtil.NATIVE_PROVIDER);
    c.init(Cipher.DECRYPT_MODE, kek);
```

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,550 @@
...
+
+  @Test
+  public void testEngineGetParametersAndIv() throws Exception {
+    final IvParameterSpec iv = TestUtil.genIv(1, 16);
+    final Cipher c = getCipher(TestUtil.NATIVE_PROVIDER);
+    assertTrue(c.getParameters() == null);
+    assertTrue(c.getIV() == null);
+    final int[] modes =
+        new int[] {
+          Cipher.ENCRYPT_MODE, Cipher.ENCRYPT_MODE, Cipher.ENCRYPT_MODE, Cipher.ENCRYPT_MODE
```
geedo0: Needs to be updated to parameterized across all cipher modes.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,550 @@
...
+    final byte[] ciphertext = c.doFinal();
+
+    c.init(Cipher.DECRYPT_MODE, kek);
+    assertThrows(
+        IllegalArgumentException.class, () -> c.doFinal(ciphertext, 0, ciphertext.length + 8));
+    assertThrows(IllegalArgumentException.class, () -> c.doFinal(ciphertext, 1, ciphertext.length));
+    assertThrows(
+        IllegalArgumentException.class, () -> c.doFinal(ciphertext, ciphertext.length + 8, 0));
+    assertArraysHexEquals(plaintext, c.doFinal(ciphertext, 0, ciphertext.length));
+  }
```
geedo0: This test does not appear to test any unique logic inside `AesKeyWrapSpi`. Suggest removing it.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,550 @@
...
+
+    // engineSetMode
+    assertThrows(GeneralSecurityException.class, () -> sneakyInvoke(spi, "engineSetMode", "OTHER"));
+    assertThrows(
+        GeneralSecurityException.class, () -> sneakyInvoke(spi, "engineSetPadding", "Other"));
+    sneakyInvoke(spi, "engineSetMode", "KW");
+  }
+
+  @Test
+  public void testBadInputs_getBlockAndKeySize() throws Throwable {
```
geedo0, line 317: pedantic: Some of these asserts, test GoodInputs

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,550 @@
...
+      assertThrows(BadPaddingException.class, () -> c.doFinal(secretBytes1));
+      assertThrows(BadPaddingException.class, () -> c.doFinal(secretBytes2));
+    }
+  }
+
+  @Test
+  public void testBadInputs_ivSize() throws Exception {
+    final IvParameterSpec iv =
+        TestUtil.genIv(
+            1, 5); // RFC-3394 iv must point to an 8 byte value or be NULL to use the default
```
geedo0: Add a test case for an IV of length 10 to assert failure of long IVs.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
geedo0, line 1: Thanks, these a very thorough tests. Like the other files, appears to be derived from `AesKeyWrapPaddingTest.java`.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,550 @@
...
+    final SecretKey key = getAesKey(128 / 8);
+    final AlgorithmParameterSpec iv = new GCMParameterSpec(128, TestUtil.getRandomBytes(16));
+    // using incorrect parameter type for the IV buffer
+    assertThrows(
+        InvalidAlgorithmParameterException.class,
+        () -> sneakyInvoke(spi, "engineInit", Cipher.WRAP_MODE, key, iv, null));
+  }
+
+  @Test
+  public void testBadInputs_bufferSize() throws Exception {
```
geedo0, line 366: Need a test case to assert that wrapping a secret that's <16 bytes fails.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTestThread.java
```diff
@@ -0,0 +1,65 @@
...
+import static com.amazon.corretto.crypto.provider.test.TestUtil.assertArraysHexEquals;
+
+import java.security.GeneralSecurityException;
+import java.security.Key;
+import java.security.SecureRandom;
+import java.util.List;
+import javax.crypto.Cipher;
+import javax.crypto.SecretKey;
+
+public class AesKeyWrapTestThread extends Thread {
```
geedo0, line 14: Any reason why you pulled this into a separate source file rather than just inlining it like in `AesKeyWrapPaddingTest`?

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
```diff
@@ -0,0 +1,550 @@
...
+
+    c.init(Cipher.UNWRAP_MODE, key1);
+    final Key unwrappedKey = c.unwrap(wrappedKey, "Generic", Cipher.SECRET_KEY);
+    assertArraysHexEquals(secret.getEncoded(), unwrappedKey.getEncoded());
+  }
+
+  @Test
+  public void threadStorm() throws GeneralSecurityException, InterruptedException {
+    final byte[] rngSeed = TestUtil.getRandomBytes(20);
+    System.out.println("RNG Seed: " + Arrays.toString(rngSeed));
```
geedo0: Let's drop this line. At some point we need to get rid of all these printlns from the test code, but I'd like to prevent adding more of them.

tst/com/amazon/corretto/crypto/provider/test/AesKwKatTest.java
geedo0, line 1: I confirmed that the *.rsp.gz files committed here contain all of the same test vectors from NIST and have not been tampered with beyond renaming, zipping, and fixing the `PLAINTEXT LENGTH` line to use an underscore. The process I used was simply unzipping the source KATs, unzipping the committed KATs, and comparing the files.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTestThread.java
sergey-gubenko: My original change was use the `AesKeyWrapTestThread` thread object in `AesKeyWrapPaddingTest` and `AesKeyWrapTest` test classes, we've later decided not to include the `AesKeyWrapPaddingTest` change in this PR to keep the changes strictly AES KW related.

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
sergey-gubenko: ok to leave the test name as is? Or I can rename it to `testInputs_getBlockAndKeySize` or something like this, will probably need to do the same in KWP

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
sergey-gubenko: `testBadInputs_bufferSize`  starts with the secret size == 16 and tests that wrap with secret size 15 and 17 are failing

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
sergey-gubenko: I copied this test from KWP tests, but looking at what it does it looks like modes are configured incorrectly, the original intention might have been to test wrap and encrypt

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
sergey-gubenko: `testBadInputs_ivSize5` is testing IV size 5 and `testBadInputs_ivSize10` is IV size 10

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
sergey-gubenko: I removed this this test

src/com/amazon/corretto/crypto/provider/AesKeyWrapSpi.java
sergey-gubenko: sorry, missed this one, just updated my "review comments" commit

tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
geedo0: I don't feel strongly about it by any means.

csrc/aes_kw.cpp
```diff
@@ -0,0 +1,101 @@
...
+        key.get_bytes(env, keybuf.buf, 0, key.len());
+        if (AES_set_encrypt_key(keybuf.buf, key.len() * 8, &aes_key) != 0) {
+            throw_openssl(EX_RUNTIME_CRYPTO, "AES key init failed");
+        }
+
+        jni_borrow ivbuf = (!iv) ? jni_borrow() : jni_borrow(env, iv, "iv");
+        jni_borrow inbuf(env, input, "input");
+        jni_borrow outbuf(env, output, "output");
+
+        size_t outlen = AES_wrap_key(&aes_key, ivbuf.data(), outbuf.data(), inbuf.data(), inbuf.len());
```
alexw91, line 50: I'm not 100% sure from just reading the code, but there might be a bug here around handling null vs zero-length IV's. 
  1. Here it looks like ACCP converts a `null` IV into a zero length byte buffer (which presumably has a non-null memory address?).
  2. ACCP passes this zero length buffer to `AES_wrap_key()`
  3. [AWS-LC checks for a null IV](https://github.com/aws/aws-lc/blob/eaaa97b058fa60a8d3f95939fd3bacaab09947d6/crypto/fipsmodule/aes/key_wrap.c#L60-L78), and if it is null, converts the null IV to a default IV as defined in RFC 3394 section 2.2.3.1. This check in AWS-LC may not work correctly if `ivbuf.data()` is non-null but `ivbuf.len()` is zero. 

Are we sure we're handling this case correctly? I'm not sure if `java_buffer().data()` returns `null` or not.

sergey-gubenko: this would be a `jni_borrow::data()` call:

```
java_buffer iv = (ivArray != nullptr) ? java_buffer::from_array(env, ivArray) : java_buffer();
...
jni_borrow ivbuf = (!iv) ? jni_borrow() : jni_borrow(env, iv, "iv");
```

if iv was nullptr, ivbuf would be initialized with a `jni_borrow()`

in this case, this a "clear" buffer with `m_pData == nullptr`

data() call should return null in this case:

`uint8_t* data() { return (uint8_t*)m_pData; }`

sergey-gubenko: I've tested the `ivArray == null` condition, `ivbuf.data()` is also null in this case

geedo0: FWIW - I had the same thought and chased this down myself while reviewing it. @sergey-gubenko wrote it up nicely, but also the unit tests/KATs prove that it's working right.

PR #499, opened 2025-10-01
Register ACCP key factory and keypair generator only if the feature is explicitly enabled

chockalingamc:
*Description of changes:*
ACCP's KeyFactory for X25519 returns keys of type EvpXECPublicKey and EvpXECPrivateKey that do not implement XECKey interfaces. So, the factory and kpgs should be registered only for clients that explicitly require this feature.


src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -573,6 +577,8 @@ public AmazonCorrettoCryptoProvider() {
     this.shouldRegisterEdKeyFactory =
         Utils.getBooleanProperty(PROPERTY_REGISTER_ED_KEYFACTORY, false);
 
+    this.shouldRegisterXEC = Utils.getBooleanProperty(PROPERTY_REGISTER_XEC, false);
```
WillChilds-Klein, line 580: can we condition this on JDK version instead of a system property?

geedo0: This approach is consistent with existing precedent. I wouldn't force such a change now (especially if it means some of these features are enabled dynamically while others are locked behind configs), but I would support a future refactor that applies this consistently.

geedo0 (approved):

WillChilds-Klein:
Closing in favor of version-based enablement. See #500 and #541.

PR #501, opened 2025-10-15
Common thread class in AES KWP tests plus a test fix

sergey-gubenko:
*Description of changes:*

Changing the KWP test class to use common `AesKeyWrapTestThread` thread class in the `threadStorm` test. `AesKeyWrapTestThread` thread class was introduced with AES KW tests.
Making `testEngineGetParametersAndIv` test parameterized plus testing all 4 cipher modes. Current implementation is testing the `ENCRYPT_MODE` 4 times.

### DIFF
--- a/tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapPaddingTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapPaddingTest.java
@@ -669,7 +669,9 @@ public final class AesKeyWrapPaddingTest {
       }
     }
     if (!results.isEmpty()) {
-      final AssertionError ex = new AssertionError("Throwable while testing threads");
+      final AssertionError ex =
+          new AssertionError(
+              "Throwable while testing threads, RNG seed: " + Arrays.toString(rngSeed));
       for (Throwable t : results) {
         t.printStackTrace();
         ex.addSuppressed(t);
--- a/tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapTest.java
@@ -527,7 +527,9 @@ public final class AesKeyWrapTest {
       }
     }
     if (!results.isEmpty()) {
-      final AssertionError ex = new AssertionError("Throwable while testing threads");
+      final AssertionError ex =
+          new AssertionError(
+              "Throwable while testing threads, RNG seed: " + Arrays.toString(rngSeed));
       for (Throwable t : results) {
         t.printStackTrace();
         ex.addSuppressed(t);


tst/com/amazon/corretto/crypto/provider/test/AesKeyWrapPaddingTest.java
```diff
@@ -628,7 +634,6 @@ public void testBadInputs_wrapAndUnwrap() throws Exception {
   @Test
   public void threadStorm() throws GeneralSecurityException, InterruptedException {
     final byte[] rngSeed = TestUtil.getRandomBytes(20);
-    System.out.println("RNG Seed: " + Arrays.toString(rngSeed));
```
WillChilds-Klein, line 631: can we leave the seed println? otherwise there's no way to reproduce potential failures.

geedo0: Has this ever happened and proved more valuable than having cleaner log outputs? The way this works is that this seed gets printed out in a way that's not well correlated to the test that it relates to which makes what your proposing difficult to begin with. I'm also not a fan of using non-deterministic testing as part of critical path PR CI. A compromise solution may be to just include the seed in the assertion failure message. At least that way, it gets printed inline with the actual test failure.

sergey-gubenko: we can potentially add this into the `AssertionError` exception message at the line `672`:

`final AssertionError ex = new AssertionError("Throwable while testing threads, seed:" + Arrays.toString(rngSeed));`

sergey-gubenko: this is how the error message is going to look like:

```
JUnit Jupiter:AesKeyWrapTest:threadStorm()
    MethodSource [className = 'com.amazon.corretto.crypto.provider.test.AesKeyWrapTest', methodName = 'threadStorm', methodParameterTypes = '']
    => java.lang.AssertionError: Throwable while testing threads, RNG seed:[28, 101, 20, -36, -71, -85, -19, -24, -114, -91, 110, 127, -62, 91, -11, -36, 124, 50, -59, 94]
       com.amazon.corretto.crypto.provider.test.AesKeyWrapTest.threadStorm(AesKeyWrapTest.java:530)
       java.base/java.lang.reflect.Method.invoke(Method.java:580)

```

sergey-gubenko: I've updated both KW and KWP `threadStorm` test to include seed in the assertion exception message:

https://github.com/corretto/amazon-corretto-crypto-provider/pull/501/commits/555a62296eec63a5450ea54a8a61e5cbb68a7ef8

WillChilds-Klein (approved):

PR #504, opened 2025-11-14
Expose XAES-256-GCM in ACCP

ttunglee:
*Issue #, if available:*

### **TL;DR** 

Briefly, XAES-256-GCM is AES-256-GCM but with larger nonce size (20 to 24 bytes) instead of 12 bytes, and it is integrated with a key derivation function, which derives a subkey prior to encryption and decryption to use rather than directly using the main key for encryption/decryption. Therefore, all test cases that apply to AES-256-GCM also apply to XAES-256-GCM, although the output ciphertexts of the two will be different. 

### **XAES-256-GCM** 
XAES-256-GCM is extended AES-GCM with a derived-key mode proposed by [Filippo Valsorda](https://words.filippo.io/xaes-256-gcm/) in 2023, followed by [a specification](https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM.md) released in 2024. [Benhamouda et al.](https://eprint.iacr.org/2025/878.pdf) provides a formal proof for the security of XAES-256-GCM beyond the birthday bound with $$2^{80}$$ encryptions without key rotation in the unique-key/block model, and [Kampanakis et al.](https://eprint.iacr.org/2025/758.pdf) introduces an extension to additionally support CMAC-based Key Commitment. The trade-off compared to the standard AES-GCM is a small computation cost incurred by invoking a key derivation function (KDF) for each encryption. Specifically, it requires three AES invocations for key derivation (in which one can be precomputed), and three more for key commitment. XAES-256-GCM uses FIPS-approved algorithms (CMAC-based KDF in [NIST SP 800-38B](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38b.pdf) and AES-GCM in [NIST SP 800-38D](https://nvlpubs.nist.gov/nistpubs/legacy/sp/nistspecialpublication800-38d.pdf)). 

XAES-256-GCM is implemented in branch xaes-256-gcm of AWS-LC repo: https://github.com/aws/aws-lc/tree/xaes-256-gcm, for both EVP_CIPHER and EVP_AEAD APIs. 

See PR: https://github.com/aws/aws-lc/pull/2750 

### **Description of changes**
- Add `xaes_256_gcm.cpp`, which is copied from `aes_gcm.cpp` with a few modifications. 
- Add `XAes256GcmSpi.java`, which is copied from `AesGcmSpi.java` with a few modifications for nonce size.
- Add `XAes256GcmTest.java`, which is copied from `AesTest.java` with a few modifications for nonce size, and added two cases for 20-byte and 23-byte nonce size. Different from the original `AesTest.java` that alternates executing AES-GCM cipher in AWS-LC library and `javax.crypto`, XAES-256-GCM is only available in AWS-LC library. 
- Add `XAes256GcmKatTest.java`, which is copied from `AesCfbTest.java`, but added KATs from https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM.md 
 
### ** Testing**
`./gradlew singleTest -DSINGLE_TEST=com.amazon.corretto.crypto.provider.test.XAes256GcmTest`
`./gradlew singleTest -DSINGLE_TEST=com.amazon.corretto.crypto.provider.test.XAes256GcmKatTest`


build.gradle
```diff
@@ -14,7 +14,7 @@ plugins {
 
 group = 'software.amazon.cryptools'
 version = '2.5.0'
-ext.awsLcMainTag = 'v1.60.0'
+ext.awsLcMainTag = 'xaes-256-gcm'
```
WillChilds-Klein, line 17: this is fine for the purposes of this PR while XAES changes remain on a non-mainline AWS-LC branch.

I've converted this PR to "draft" status until/if upstream XAES changes are merged in AWS-LC.

ttunglee: Thank you!!!

PR #505, opened 2025-11-24
Add POSIX permissions checks to Loader

geedo0:
*Issue #, if available:*
CryptoAlg-3511

*Description of changes:*
Assertions around the POSIX permissions and owner of the temporary
directory created by `Loader` have been added to prevent a race
condition between creation of the directory and use of it for loading
the ACCP native library.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/Loader.java
+++ b/src/com/amazon/corretto/crypto/provider/Loader.java
@@ -13,6 +13,7 @@ import java.nio.file.StandardOpenOption;
 import java.nio.file.attribute.FileAttribute;
 import java.nio.file.attribute.PosixFilePermission;
 import java.nio.file.attribute.PosixFilePermissions;
+import java.nio.file.attribute.UserPrincipal;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


src/com/amazon/corretto/crypto/provider/Loader.java
```diff
@@ -429,6 +430,33 @@ private static byte[] bootstrapRng(int numBytes) throws IOException {
     throw new AssertionError("Unable to read enough entropy");
   }
 
+  private static void verifyDirectoryOwnershipAndPermissions(Path directory) throws IOException {
+    UserPrincipal directoryOwner = Files.getOwner(directory);
+    String currentUsername = System.getProperty("user.name");
```
WillChilds-Klein, line 435: is this sufficiently robust? JVM system properties can be overridden. i don't have a better idea than shelling out to `whoami` or equivalent, but that's not particularly portable...

so, not a blocker, just something for us to hopefully discuss a bit.

geedo0: What threat are you trying to protect against that isn't worse in other ways? We have precedent of using standard Java system properties as nearby as 10 lines below here.

WillChilds-Klein: Not a security concern -- if an attacker can set JVM system properties, they've already won. Mildly worried about a user setting this system property for some obscure reason, leading to usability issues. I gen-AI'd up a quick demo and confirmed that this property can be overridden without issue.

I'm fine with trusting the user to not foot-gun themselves.

```
> please compile and run Username.java with a java system property of user.name=nonsense, then `cat` Username.java

> I'll help you compile and run Username.java with the specified system property, then show the source code. Let me first check if the file exists and then proceed with the compilation and execution.


🛠️  Using tool: fs_read (trusted)
 ⋮
 ● Reading file: Username.java, all lines
 ⋮
 ● Completed in 0.0s


> Now I'll compile and run the Java file with the specified system property:


🛠️  Using tool: execute_bash
 ⋮
 ● I will run the following shell command: javac Username.java
 ⋮
 ↳ Purpose: Compile Username.java


Allow this action? Use 't' to trust (always allow) this tool for the session. [y/n/t]:

> y

Picked up JAVA_TOOL_OPTIONS: -Dlog4j2.formatMsgNoLookups=true

 ⋮
 ● Completed in 0.583s





🛠️  Using tool: execute_bash
 ⋮
 ● I will run the following shell command:
java -Duser.name=nonsense Username
 ⋮
 ↳ Purpose: Run Username.java with user.name=nonsense system property


Allow this action? Use 't' to trust (always allow) this tool for the session. [y/n/t]:

> y

Picked up JAVA_TOOL_OPTIONS: -Dlog4j2.formatMsgNoLookups=true
nonsense

 ⋮
 ● Completed in 0.123s





🛠️  Using tool: execute_bash (trusted)
 ⋮
 ● I will run the following shell command: cat Username.java
 ⋮
 ↳ Purpose: Display Username.java source code

public class Username {
    public static void main(String[] args) {
        System.out.println(System.getProperty("user.name"));
    }
}

 ⋮
 ● Completed in 0.43s
```

Issue #516, opened 2026-02-12
ACCP RSASSA-PSS signature implementation rejects RSASSA-PSS keys with `InvalidKeySpecException: Incorrect key type`


msteindorfer:
**Summary:** ACCP's RSASSA-PSS signature implementation only works with regular RSA keys, not with RSASSA-PSS-specific keys. This violates JCA compatibility since `SunRsaSign` handles both key types correctly.

**Expected behavior:** ACCP should accept both RSA and RSASSA-PSS keys for RSASSA-PSS signatures, matching `SunRsaSign` behavior.

**Actual behavior:** `InvalidKeyException` / `InvalidKeySpecException` thrown at `engineInitVerify()` when using RSASSA-PSS keys, regardless of call order. This applies at least to keys obtained via `KeyPairGenerator.getInstance("RSASSA-PSS").initialize(...).generateKeyPair()` and `KeyFactory.getInstance("RSASSA-PSS").generatePublic(...)`, as tested.

**Reproducer:** The issue can be reproduced with ACCP 2.5.0. It appears that the specific Java major version or distribution doesn't matter here. Please find [RSASSAPSSReproducer.java](https://github.com/user-attachments/files/25263300/RSASSAPSSReproducer.java) and a corresponding execution log attached.

```
$ ls -1
AmazonCorrettoCryptoProvider-2.5.0-osx-aarch_64.jar
RSASSAPSSReproducer.java

$ java -cp AmazonCorrettoCryptoProvider-2.5.0-osx-aarch_64.jar RSASSAPSSReproducer.java
=== Test compatibility of RSASSA-PSS Signature implementations with RSASSA-PSS keys ===

--- Test 1: Without ACCP (baseline) ---
Generated key algorithm: RSASSA-PSS
Signature provider: SunRsaSign
✓ initVerify() succeeded
✓ setParameter() succeeded
✓ SUCCESS


--- Test 2: With ACCP installed ---
✓ Installed AmazonCorrettoCryptoProvider as provider #1

Generated key algorithm: RSASSA-PSS
Signature provider: AmazonCorrettoCryptoProvider
✗ FAILED with InvalidKeyException
  Message: java.security.spec.InvalidKeySpecException: Incorrect key type
java.security.InvalidKeyException: java.security.spec.InvalidKeySpecException: Incorrect key type
	at com.amazon.corretto.crypto.provider.EvpKeyFactory.engineTranslateKey(EvpKeyFactory.java:126)
[...]

PR #521, opened 2026-03-03
Retry RSA key parsing to handle RSASSA-PSS OID

WillChilds-Klein:
*Issue #, if available:* Addresses #516

# Notes

In #516 @msteindorfer reported that ACCP fails to parse SunRsaSign-generated `RSASSA-PSS` keys. `RSA` and `RSASSA-PSS` keys should be identical in terms of key material, but OpenJDK [gives them different OIDs](https://github.com/openjdk/jdk/blob/a6db3f870218bc016ecc67e3f5e142d6491ac080/src/java.base/share/classes/sun/security/rsa/RSAUtil.java#L46-L47) on serialization, and ACCP only checks for `RSA`'s NID/OID as specified in [EvpKeyType](https://github.com/corretto/amazon-corretto-crypto-provider/blob/afcbb82a13b196dde3f535193aac4a6e4bd3801d/src/com/amazon/corretto/crypto/provider/EvpKeyType.java#L19-L24).

This PR adds a `RSASSA-PSS` retry when `RSA`-typed keys fail to parse and adjust native logic to handle either RSA value returned from AWS-LC's `EVP_PKEY_base_id()`.

# Testing

- new unit tests failed before changes, succeed after

---

### DIFF
--- a/csrc/java_evp_keys.cpp
+++ b/csrc/java_evp_keys.cpp
@@ -104,8 +104,21 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_p
 
         {
             jni_borrow borrow = jni_borrow(env, pkcs8Buff, "pkcs8Buff");
-            result.set(der2EvpPrivateKey(borrow, derLen, evpType, shouldCheckPrivate, EX_INVALID_KEY_SPEC));
-            if (EVP_PKEY_base_id(result) != evpType) {
+            // RSASSA-PSS keys use OID id-RSASSA-PSS (1.2.840.113549.1.1.10) which
+            // causes d2i_PrivateKey(EVP_PKEY_RSA, ...) to fail. Retry with
+            // EVP_PKEY_RSA_PSS if the initial parse fails for an RSA key type.
+            try {
+                result.set(der2EvpPrivateKey(borrow, derLen, evpType, shouldCheckPrivate, EX_INVALID_KEY_SPEC));
+            } catch (java_ex&) {
+                if (evpType == EVP_PKEY_RSA) {
+                    result.set(
+                        der2EvpPrivateKey(borrow, derLen, EVP_PKEY_RSA_PSS, shouldCheckPrivate, EX_INVALID_KEY_SPEC));
+                } else {
+                    throw;
+                }
+            }
+            int baseId = EVP_PKEY_base_id(result);
+            if (baseId != evpType && !(evpType == EVP_PKEY_RSA && baseId == EVP_PKEY_RSA_PSS)) {
                 throw_java_ex(EX_INVALID_KEY_SPEC, "Incorrect key type");
             }
         }
@@ -134,7 +147,8 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_x
         {
             jni_borrow borrow = jni_borrow(env, x509Buff, "x509Buff");
             result.set(der2EvpPublicKey(borrow, derLen, EX_INVALID_KEY_SPEC));
-            if (EVP_PKEY_base_id(result) != evpType) {
+            int baseId = EVP_PKEY_base_id(result);
+            if (baseId != evpType && !(evpType == EVP_PKEY_RSA && baseId == EVP_PKEY_RSA_PSS)) {
                 throw_java_ex(EX_INVALID_KEY_SPEC, "Incorrect key type");
             }
         }
--- a/csrc/java_evp_keys.cpp
+++ b/csrc/java_evp_keys.cpp
@@ -111,6 +111,7 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_p
                 result.set(der2EvpPrivateKey(borrow, derLen, evpType, shouldCheckPrivate, EX_INVALID_KEY_SPEC));
             } catch (java_ex&) {
                 if (evpType == EVP_PKEY_RSA) {
+                    ERR_clear_error();
                     result.set(
                         der2EvpPrivateKey(borrow, derLen, EVP_PKEY_RSA_PSS, shouldCheckPrivate, EX_INVALID_KEY_SPEC));
                 } else {
--- a/csrc/keyutils.h
+++ b/csrc/keyutils.h
@@ -88,6 +88,12 @@ EVP_PKEY* der2EvpPublicKey(const unsigned char* der, const int derLen, const cha
 bool checkKey(const EVP_PKEY* key);
 static bool inline BN_null_or_zero(const BIGNUM* bn) { return nullptr == bn || BN_is_zero(bn); }
 
+// Returns true if keyType is EVP_PKEY_RSA or EVP_PKEY_RSA_PSS. RSASSA-PSS keys
+// (OID 1.2.840.113549.1.1.10) are functionally identical to RSA keys but use a
+// different AlgorithmIdentifier, causing EVP_PKEY_base_id() to return
+// EVP_PKEY_RSA_PSS instead of EVP_PKEY_RSA.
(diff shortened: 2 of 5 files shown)


sgmenda (dismissed):

csrc/java_evp_keys.cpp
```diff
@@ -104,8 +104,21 @@ JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_EvpKeyFactory_p
...
-            if (EVP_PKEY_base_id(result) != evpType) {
+            // RSASSA-PSS keys use OID id-RSASSA-PSS (1.2.840.113549.1.1.10) which
+            // causes d2i_PrivateKey(EVP_PKEY_RSA, ...) to fail. Retry with
+            // EVP_PKEY_RSA_PSS if the initial parse fails for an RSA key type.
+            try {
+                result.set(der2EvpPrivateKey(borrow, derLen, evpType, shouldCheckPrivate, EX_INVALID_KEY_SPEC));
+            } catch (java_ex&) {
+                if (evpType == EVP_PKEY_RSA) {
+                    result.set(
+                        der2EvpPrivateKey(borrow, derLen, EVP_PKEY_RSA_PSS, shouldCheckPrivate, EX_INVALID_KEY_SPEC));
```
justsmth, line 116: Should there be a `ERR_clear_error()` prior to a retry to avoid confusion when diagnosing failures?

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -873,6 +873,82 @@ public byte[] getEncoded() {
...
+    final KeyPair pssKeyPair = kpg.generateKeyPair();
+
+    // Translate RSASSA-PSS keys through ACCP's RSA KeyFactory
+    final KeyFactory accpKf = KeyFactory.getInstance("RSA", NATIVE_PROVIDER);
+    final PublicKey accpPub =
+        accpKf.generatePublic(
+            new java.security.spec.X509EncodedKeySpec(pssKeyPair.getPublic().getEncoded()));
+    final PrivateKey accpPriv =
+        accpKf.generatePrivate(
+            new java.security.spec.PKCS8EncodedKeySpec(pssKeyPair.getPrivate().getEncoded()));
```
justsmth, line 935: NP: Verify the name with e.g. `assertEquals("RSA", accpPub.getAlgorithm());`? 

This also documents the (arguably surprising) behavior, so if someone later adds a proper `RSASSA-PSS` KeyFactory, then this assertion starts failing.

csrc/keyutils.h
```diff
@@ -88,6 +88,12 @@ EVP_PKEY* der2EvpPublicKey(const unsigned char* der, const int derLen, const cha
 bool checkKey(const EVP_PKEY* key);
 static bool inline BN_null_or_zero(const BIGNUM* bn) { return nullptr == bn || BN_is_zero(bn); }
 
+// Returns true if keyType is EVP_PKEY_RSA or EVP_PKEY_RSA_PSS. RSASSA-PSS keys
+// (OID 1.2.840.113549.1.1.10) are functionally identical to RSA keys but use a
+// different AlgorithmIdentifier, causing EVP_PKEY_base_id() to return
+// EVP_PKEY_RSA_PSS instead of EVP_PKEY_RSA.
+static inline bool isRsaKeyType(int keyType) { return keyType == EVP_PKEY_RSA || keyType == EVP_PKEY_RSA_PSS; }
```
justsmth, line 95: An `id-RSASSA-PSS`-tagged key is supposed to be usage-restricted to PSS signatures. By collapsing both key types into a single `isRsaKeyType` predicate, are we also implicitly allowing PSS-constrained keys to be used for `RSA/ECB/OAEPPadding`, `SHA256withRSA` PKCS#1 v1.5, etc.

Is this allowance consistent with other JCE implementations?

justsmth (commented):
[`EvpRsaKey` hard-codes `EvpKeyType.RSA`](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/EvpRsaKey.java#L17-L19), so after a PSS-encoded key flows through ACCP's RSA KeyFactory, `getAlgorithm()` returns `"RSA"` — even though the underlying `EVP_PKEY` is still `EVP_PKEY_RSA_PSS` and `encodeRsaPrivateKey`/`EVP_PKEY2PKCS8` will re-emit the `id-RSASSA-PSS` OID on export.

Should there be a separate type for `EVP_PKEY_RSA_PSS`?

csrc/java_evp_keys.cpp
WillChilds-Klein: yes, good call.

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
WillChilds-Klein: sure, added assertions for pub/priv keys.

csrc/keyutils.h
WillChilds-Klein: > An id-RSASSA-PSS-tagged key is supposed to be usage-restricted to PSS signatures.

why? PSS is just a padding scheme; as far as I know key material is identical between RSA and RSASSA-PSS keys.

> Is this allowance consistent with other JCE implementations?

looks like RSASSA-PSS keys can be used with other RSA flavors. courtesy of claude:

```
$ cat PssKeyMisuseTest.java
import java.security.KeyPair;
import java.security.KeyPairGenerator;
import java.security.Provider;
import java.security.Security;
import java.security.Signature;
import javax.crypto.Cipher;

public class PssKeyMisuseTest {

    public static void main(String[] args) throws Exception {
        KeyPairGenerator kpg = KeyPairGenerator.getInstance("RSASSA-PSS");
        Provider kpgProvider = kpg.getProvider();
        System.out.println("KeyPairGenerator(RSASSA-PSS) provider: " + kpgProvider.getName());

        kpg.initialize(2048);
        KeyPair kp = kpg.generateKeyPair();

        System.out.println("Public  key algorithm: " + kp.getPublic().getAlgorithm());
        System.out.println("Private key algorithm: " + kp.getPrivate().getAlgorithm());
        System.out.println();

        byte[] plaintext = "hello pkcs1".getBytes();

        System.out.println("--- 1) PKCS1-padded encryption (RSA/ECB/PKCS1Padding) ---");
        try {
            Cipher cipher = Cipher.getInstance("RSA/ECB/PKCS1Padding");
            System.out.println("Cipher provider: " + cipher.getProvider().getName());
            cipher.init(Cipher.ENCRYPT_MODE, kp.getPublic());
            byte[] ct = cipher.doFinal(plaintext);
            System.out.println("Ciphertext length: " + ct.length);
        } catch (Exception e) {
            System.out.println("FAILED: " + e.getClass().getName() + ": " + e.getMessage());
        }
        System.out.println();

        System.out.println("--- 2) PKCS1-padded signature (SHA256withRSA) ---");
        try {
            Signature sig = Signature.getInstance("SHA256withRSA");
            System.out.println("Signature provider: " + sig.getProvider().getName());
            sig.initSign(kp.getPrivate());
            sig.update(plaintext);
            byte[] s = sig.sign();
            System.out.println("Signature length: " + s.length);
        } catch (Exception e) {
            System.out.println("FAILED: " + e.getClass().getName() + ": " + e.getMessage());
        }
    }
}


$ javac PssKeyMisuseTest.java && java PssKeyMisuseTest 2>/dev/null
Picked up JAVA_TOOL_OPTIONS: -Dlog4j2.formatMsgNoLookups=true
KeyPairGenerator(RSASSA-PSS) provider: SunRsaSign
Public  key algorithm: RSASSA-PSS
Private key algorithm: RSASSA-PSS

--- 1) PKCS1-padded encryption (RSA/ECB/PKCS1Padding) ---
Cipher provider: SunJCE
Ciphertext length: 256

--- 2) PKCS1-padded signature (SHA256withRSA) ---
Signature provider: SunRsaSign
Signature length: 256
```

justsmth (approved):

sgmenda (approved):

PR #524, opened 2026-03-06
Support elliptic-curve aliases by comparing OIDs instead of names

msteindorfer:
*Issue #, if available:* N/A

*Description of changes:*

Since JDK 24, the `EcParametersTest` test fails since, instead expected curve name `secp256r1`, the actual value is `prime256v1`. The latter is a valid alternative name for the former.

The test case already covers the corner case that whenever JCE returns the OID instead of the human-readable name, it will compare both OIDs. This commit expands the same logic to always instead of names, default to OID comparison.

*Further  comments:*

While this PR contains just the test fix to accept the different aliased names, it should separately be followed up if expected and actual elliptic curve names must be identical, or if OID equality is sufficient.

_Note, since pull-request builds only cover JDK 8, 11 and 17, this issue doesn't show in CI/CD coverage. @WillChilds-Klein I suggest to extend pull-request build testing also with the most recent two LTS versions, JDK 21 and 25._

### DIFF
--- a/tst/com/amazon/corretto/crypto/provider/test/EcGenTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/EcGenTest.java
@@ -378,7 +378,8 @@ public class EcGenTest {
 
   public static void assertECEquals(
       final String message, final ECGenParameterSpec expected, final ECGenParameterSpec actual) {
-    assertEquals(expected.getName(), actual.getName(), message);
+    assertEquals(
+        TestUtil.getCurveOid(expected.getName()), TestUtil.getCurveOid(actual.getName()), message);
   }
 
   private static class TestThread extends Thread {


WillChilds-Klein (approved):

WillChilds-Klein:
> @WillChilds-Klein I suggest to extend pull-request build testing also with the most recent two LTS versions, JDK 21 and 25.

yes. we now have a JDK21 CI job for ARM. I'm [working on adding one for JDK25](https://github.com/corretto/amazon-corretto-crypto-provider/pull/539), but it looks like we'll need to upgrade to Gradle 9 for that which is a bit involved. Coming soon.

geedo0 (approved):

msteindorfer:
@WillChilds-Klein @geedo0, thanks to both of you for the reviews and approvals of the various smaller PRs.

PR #531, opened 2026-04-03
Add MlKemUtils with ML-KEM seed private key expansion

geedo0:
*Issue #, if available:*
P407404249

*Description of changes:*
Add MlKemUtils.expandPrivateKey() utility that converts an ML-KEM
private key from the 64-byte seed format to the expanded decapsulation
key format, as defined in FIPS 203 and RFC 9935.

The implementation leverages AWS-LC's existing seed expansion support
in kem_priv_decode (KEM_KEY_set_raw_keypair_from_seed) which
automatically expands seed-format keys during PKCS8 parsing, and
kem_priv_encode which always writes the expanded format.

This mirrors the existing MlDsaUtils.expandPrivateKey() for ML-DSA.

### DIFF
--- a/src/com/amazon/corretto/crypto/utils/MlKemUtils.java
+++ b/src/com/amazon/corretto/crypto/utils/MlKemUtils.java
@@ -1,0 +1,33 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.utils;
+
+import java.security.PrivateKey;
+
+/** Public utility methods for ML-KEM operations. */
+public final class MlKemUtils {
+  private MlKemUtils() {} // private constructor to prevent instantiation
+
+  private static native byte[] expandPrivateKeyInternal(byte[] key);
+
+  /**
+   * Returns an expanded ML-KEM private key, whether the key passed in is based on a seed or
+   * expanded. It returns the PKCS8-encoded expanded key.
+   *
+   * <p>The seed format is a 64-byte value (d || z) as defined in FIPS 203. The expanded
+   * decapsulation key is derived using ML-KEM.KeyGen_internal(d, z) (Algorithm 16).
+   *
+   * <p>See <a href="https://csrc.nist.gov/pubs/fips/203/final">FIPS 203</a>
+   *
+   * <p>See <a href="https://datatracker.ietf.org/doc/rfc9935/">RFC 9935</a>
+   *
+   * @param key an ML-KEM private key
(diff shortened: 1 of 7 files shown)


tst/com/amazon/corretto/crypto/provider/test/MlKemUtilsTest.java
```diff
@@ -0,0 +1,104 @@
...
+      return false;
+    } catch (Exception e) {
+      return true;
+    }
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"ML-KEM-512", "ML-KEM-768", "ML-KEM-1024"})
+  @DisabledIf("mlKemDisabled")
+  public void testExpandSeedKey(String algorithm) throws Exception {
```
WillChilds-Klein, line 81: can we do a decaps with both seed and expanded keys against the same ciphertext and assert they yield the same shared secret?

src/com/amazon/corretto/crypto/utils/MlKemUtils.java
```diff
@@ -0,0 +1,33 @@
...
+   * decapsulation key is derived using ML-KEM.KeyGen_internal(d, z) (Algorithm 16).
+   *
+   * <p>See <a href="https://csrc.nist.gov/pubs/fips/203/final">FIPS 203</a>
+   *
+   * <p>See <a href="https://datatracker.ietf.org/doc/rfc9935/">RFC 9935</a>
+   *
+   * @param key an ML-KEM private key
+   * @return a byte[] containing the PKCS8-encoded expanded private key
+   */
+  public static byte[] expandPrivateKey(PrivateKey key) {
```
WillChilds-Klein, line 27: should we add a check that ML-KEM is enabled?

WillChilds-Klein (commented):
could we please set [`--target-jdk-version 17`](https://github.com/corretto/amazon-corretto-crypto-provider/blob/eeb168e3c5c05f50d1eeb42bd49a1f0b25a723d9/tests/ci/run_accp_basic_tests.sh#L38) [here](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/.github/workflows/codecov-ci.yml#L29) to get coverage on ML-KEM stuff?

src/com/amazon/corretto/crypto/utils/MlKemUtils.java
geedo0: Despite being strongly linked to the KEM use case, this utility doesn't use any of the JDK KEM APIs so it will compile/work on any JDK version.

WillChilds-Klein (dismissed):

WillChilds-Klein (approved):

sgmenda (approved):
lgtm

PR #534, opened 2026-04-16
Several small bug fixes

dkostic:
- Fix use-after-free in GCM encryptDoFinal error path for borrowed contexts
- Use SecureAlloc for ECDH shared secret vector
- Fix RSA_blinding_off called after RSA_free in key rebuild path
- Fix heap-allocated exception (throw new -> throw) in bounce_buffer
- Fix secureZero called with zero length in rdrand
- Fix inverted null check in DRBG cleanup
- Fix swapped fcntl arguments in entropy fallback path
- Fix incomplete buffer zeroing on entropy read error
- Fix unlikely() macro parenthesization in buffer bounds checks
- Reset RSA cipher buffer on same-key re-initialization

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/RsaCipher.java
+++ b/src/com/amazon/corretto/crypto/provider/RsaCipher.java
@@ -334,6 +334,8 @@ class RsaCipher extends CipherSpi {
               keySizeBytes_ = (key_.getModulus().bitLength() + 7) / 8;
               buffer_ = new AccessibleByteArrayOutputStream();
               parseKey();
+          } else {
+              buffer_ = new AccessibleByteArrayOutputStream();
           }
         }
     }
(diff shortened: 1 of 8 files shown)


src/com/amazon/corretto/crypto/provider/RsaCipher.java
```diff
@@ -334,6 +334,8 @@ protected void engineInit(final int opmode, final Key key, final SecureRandom ra
               keySizeBytes_ = (key_.getModulus().bitLength() + 7) / 8;
               buffer_ = new AccessibleByteArrayOutputStream();
               parseKey();
+          } else {
+              buffer_ = new AccessibleByteArrayOutputStream();
```
WillChilds-Klein, line 338: can we just pull the assignment out of the `if` block rather than add the `else` block? `parseKey()` doesn't reference `buffer_`.

WillChilds-Klein (approved):

PR #535, opened 2026-04-17
Throw IllegalBlockSizeException for undersized RSA ciphertext

geedo0:
*Description of changes:*
When decrypting an RSA ciphertext shorter than the key size, the native
layer hit RSA_R_DATA_LEN_NOT_EQUAL_TO_MOD_LEN and fell through to a
generic error path that threw RuntimeCryptoException. This violates the
JCE contract — Cipher.doFinal() declares IllegalBlockSizeException and
BadPaddingException, not RuntimeCryptoException.

The input length != modulus length condition is a block size mismatch,
not a padding error, so IllegalBlockSizeException is the right choice.
This is consistent with the Java-side guard that already throws
IllegalBlockSizeException when the input is longer than the key size.

### DIFF
--- a/tst/com/amazon/corretto/crypto/provider/test/RsaCipherTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/RsaCipherTest.java
@@ -296,6 +296,22 @@ public class RsaCipherTest {
     assertThrows(IllegalBlockSizeException.class, () -> nativeEncrypt.doFinal(ciphertext));
   }
 
+  @ParameterizedTest
+  @MethodSource("paddingXlengthParams")
+  public void undersizedCiphertext(
+      final String padding,
+      final Integer keySize,
+      final OAEPParameterSpec oaep,
+      final String ignoredName)
+      throws GeneralSecurityException {
+    final Cipher dec = getNativeCipher(padding);
+    dec.init(Cipher.DECRYPT_MODE, getKeyPair(keySize).getPrivate(), oaep);
+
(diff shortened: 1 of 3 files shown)


tst/com/amazon/corretto/crypto/provider/test/RsaCipherTest.java
```diff
@@ -296,6 +296,22 @@ public void overlargeCiphertext(
...
+      final Integer keySize,
+      final OAEPParameterSpec oaep,
+      final String ignoredName)
+      throws GeneralSecurityException {
+    final Cipher dec = getNativeCipher(padding);
+    dec.init(Cipher.DECRYPT_MODE, getKeyPair(keySize).getPrivate(), oaep);
+
+    byte[] ciphertext = new byte[(keySize / 8) - 1];
+    Arrays.fill(ciphertext, (byte) 1);
+    assertThrows(IllegalBlockSizeException.class, () -> dec.doFinal(ciphertext));
```
WillChilds-Klein, line 312: np: should we assert that default java does this too?

WillChilds-Klein (approved):

tst/com/amazon/corretto/crypto/provider/test/RsaCipherTest.java
geedo0: I don't think this behavior is important enough to test across all platforms.

skmcgrail (approved):

PR #536, opened 2026-04-19
Remove NONEwithRSA input length restriction, DigestInfo

WillChilds-Klein:
*Issue #, if available:*

*Description of changes:*

# Summary

Our initial implementation of `NONEwithRSA` is compatible with hash-based signature implementations (e.g. `SHA256withRSA`) from JDK-default and BC implementations, but is incompatible with those providers' respective `NONEwithRSA` implementations in two important ways:

1. serialized/expected signatures wrapped in a `DigestInfo` DER structure rather than raw signature bytes
2. strict length input message length validation against configured digest algorithm, defaulting to SHA256

This change aligns behavior with SunJCE and BC providers, adding tests asserting that.

This issue initially surfaced via an internal integration test failure failing to negotiate TLSv1.0, the JSSE implementation of which relied on `NONEwithRSA` for its underlying cryptography. I posted an [internal CR](https://code.amazon.com/reviews/CR-268688190/) confirming the efficacy of this fix.

# Source changes

## sign.cpp

- PKCS1 sign path now uses `RSA_sign_raw` instead of `RSA_sign` signs raw bytes directly without `DigestInfo` wrapping
- PKCS1 verify path now uses `RSA_verify_raw` + `CRYPTO_memcmp` instead of `RSA_verify` compares raw bytes directly w/o `DigestInfo` DER wrapper
- Digest-length validation (`length != hLen`) moved from shared top-level check into PSS-only branch
- New top-level guard rejects input larger than `RSA_size - 11` (11 bytes PKCS1 padding overhead)

## NoneWithRsa.java

- Buffer changed from fixed 64-byte capacity to unbounded (new `AccessibleByteArrayOutputStream()`)
- `engineUpdate`: digest-length enforcement now conditional on `paddingType == RSA_PKCS1_PSS_PADDING`; PKCS1 accepts any message length small enough for RSA modulus
- `engineSetParameter`: throws `UnsupportedOperationException for `NONEwithRSA (matching SunJCE/BC); PSS path validates PSSParameterSpec and returns before the throw

# Test changes

## New tests

- `testMessageLargerThanModulusThrows` asserts `keySize/8 - 11` bytes (msg just small enough w/ padding) succeeds and `keySize/8 - 10` bytes (too large w/ padding) fails for both ACCP and SunJCE
- `testNoneWithRsaSetParameterThrowsCompat` confirms SunJCE, BC, and ACCP all throw `UnsupportedOperationException` on setParameter

## Modified tests

- `testNoneWithRsaInputValidation` now asserts arbitrary-length inputs (16, 33 bytes. much smaller than key modulus size.) sign and verify successfully
- `testNoneWithRsaDifferentDigests` / `testNoneWithRsaSignSha512Digest` removed `setParameter` calls; sign raw digests directly
- `testNoneWithRsaInteropWithSunRsaSign` and `testNoneWithRsaInteropWithBouncyCastle` refactored into single parameterized `testNoneWithRsaInteropWithOtherProviders` testing ACCP SunJCE/BC cross-verification
- `testNoneWithRsaCrossAlgorithmInterop` now asserts NONEwithRSA and SHA256withRSA are NOT interoperable (`DigestInfo` mismatch), but the trailing 32 digest bytes in the decrypted plaintext are identical
- `testNoneWithRsaSetParameterRejectsAllSpecs` updated to expect `UnsupportedOperationException` for all param types including `PSSParameterSpec`
- `testNoneWithRsaSetParameterWithBufferedData` updated to expect `UnsupportedOperationException` regardless of buffer state

---

### DIFF
--- a/csrc/sign.cpp
+++ b/csrc/sign.cpp
@@ -648,14 +648,10 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_NoneWithRs
             throw_java_ex(EX_SIGNATURE_EXCEPTION, "Failed to get RSA key");
         }
 
-        const size_t rsaSize = RSA_size(rsa);
-        if ((size_t)length > (rsaSize - 11)) {  // PKCS1 padding is 11 bytes for raw NONEwithRSA
-            throw_java_ex(EX_SIGNATURE_EXCEPTION, "Message too long for RSA key");
-        }
-
         java_buffer digestBuf = java_buffer::from_array(env, digestArr, offset, length);
 
         // Allocate output buffer for signature
+        const size_t rsaSize = RSA_size(rsa);
         std::vector<uint8_t, SecureAlloc<uint8_t> > signature(rsaSize);
 
         // Borrow digest for the crypto call, release before creating Java array
@@ -676,6 +672,9 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_NoneWithRs
                 break;
             }
             case RSA_PKCS1_PADDING: {
+                if ((size_t)length > (rsaSize - 11)) {  // PKCS1 padding is 11 bytes for raw NONEwithRSA
+                    throw_java_ex(EX_SIGNATURE_EXCEPTION, "Message too long for RSA key");
+                }
                 int rc = RSA_sign_raw(rsa, &sigLen, signature.data(), rsaSize,
                     digest.data(), length, RSA_PKCS1_PADDING);
                 if (rc != 1) {
@@ -743,7 +742,9 @@ JNIEXPORT jboolean JNICALL Java_com_amazon_corretto_crypto_provider_NoneWithRsa_
             break;
         }
         case RSA_PKCS1_PADDING: {
-            // Decrypt signature and strip PKCS1 padding
+            if ((size_t)length > (RSA_size(rsa) - 11)) {
+                throw_java_ex(EX_SIGNATURE_EXCEPTION, "Message too long for RSA key");
+            }
             size_t outLen = 0;
             std::vector<uint8_t> buf(RSA_size(rsa));
             if (1 != RSA_verify_raw(rsa, &outLen, buf.data(), buf.size(), sig.data(), sig.len(), RSA_PKCS1_PADDING)
--- a/csrc/sign.cpp
+++ b/csrc/sign.cpp
@@ -648,39 +648,39 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_NoneWithRs
             throw_java_ex(EX_SIGNATURE_EXCEPTION, "Failed to get RSA key");
         }
 
-        size_t hLen = EVP_MD_size(md);
-
-        if (static_cast<size_t>(length) != hLen) {
-            throw_java_ex(EX_SIGNATURE_EXCEPTION, "Digest length mismatch");
+        const size_t rsaSize = RSA_size(rsa);
+        if ((size_t)length > (rsaSize - 11)) {  // PKCS1 padding is 11 bytes for raw NONEwithRSA
+            throw_java_ex(EX_SIGNATURE_EXCEPTION, "Message too long for RSA key");
         }
 
         java_buffer digestBuf = java_buffer::from_array(env, digestArr, offset, length);
 
         // Allocate output buffer for signature
-        size_t maxOut = RSA_size(rsa);
-        std::vector<uint8_t, SecureAlloc<uint8_t> > signature(maxOut);
+        std::vector<uint8_t, SecureAlloc<uint8_t> > signature(rsaSize);
 
         // Borrow digest for the crypto call, release before creating Java array
+        size_t sigLen = 0;
         {
             jni_borrow digest(env, digestBuf, "digest");
 
             switch (paddingType) {
             case RSA_PKCS1_PSS_PADDING: {
+                if (static_cast<size_t>(length) != EVP_MD_size(md)) {
+                    throw_java_ex(EX_SIGNATURE_EXCEPTION, "Digest length mismatch");
+                }
                 const EVP_MD* mgf_md = reinterpret_cast<const EVP_MD*>(mgfMd);
-                size_t sigLen = 0;
-                if (RSA_sign_pss_mgf1(rsa, &sigLen, signature.data(), maxOut, digest.data(), hLen, md, mgf_md, saltLen)
+                if (RSA_sign_pss_mgf1(rsa, &sigLen, signature.data(), rsaSize, digest.data(), length, md, mgf_md, saltLen)
                     != 1) {
                     throw_openssl("RSA_sign_pss_mgf1 failed");
                 }
-                signature.resize(sigLen);
                 break;
             }
             case RSA_PKCS1_PADDING: {
-                unsigned int sigLen = 0;
-                if (RSA_sign(EVP_MD_type(md), digest.data(), hLen, signature.data(), &sigLen, rsa) != 1) {
-                    throw_openssl("RSA_sign failed");
+                int rc = RSA_sign_raw(rsa, &sigLen, signature.data(), rsaSize,
+                    digest.data(), length, RSA_PKCS1_PADDING);
+                if (rc != 1) {
+                    throw_openssl("RSA_sign_raw failed");
                 }
-                signature.resize(sigLen);
                 break;
             }
             default:
@@ -688,6 +688,7 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_NoneWithRs
             }
         }
 
+        signature.resize(sigLen);
         return vecToArray(env, signature);
 
     } catch (java_ex& ex) {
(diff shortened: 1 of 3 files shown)


csrc/sign.cpp
```diff
@@ -648,46 +648,47 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_NoneWithRs
             throw_java_ex(EX_SIGNATURE_EXCEPTION, "Failed to get RSA key");
         }
 
-        size_t hLen = EVP_MD_size(md);
-
-        if (static_cast<size_t>(length) != hLen) {
-            throw_java_ex(EX_SIGNATURE_EXCEPTION, "Digest length mismatch");
+        const size_t rsaSize = RSA_size(rsa);
+        if ((size_t)length > (rsaSize - 11)) {  // PKCS1 padding is 11 bytes for raw NONEwithRSA
+            throw_java_ex(EX_SIGNATURE_EXCEPTION, "Message too long for RSA key");
```
justsmth: This guard is PKCS#1-specific (as the comment indicates), but it fires unconditionally for all `paddingType` values including PSS? I'd move this into the `RSA_PKCS1_PADDING` case below.

src/com/amazon/corretto/crypto/provider/NoneWithRsa.java
justsmth, line 37: Should this javadoc be updated to reflect the new behavior?

tst/com/amazon/corretto/crypto/provider/test/NoneWithRsaTest.java
```diff
@@ -1686,67 +1692,55 @@ public void testNoneWithRsaSetParams() throws Exception {
     assertTrue(sig.verify(signature));
   }
 
-  @Test
-  public void testNoneWithRsaInteropWithSunRsaSign() throws Exception {
-    KeyPair kp = generateKeyPair(2048);
-    byte[] message = "SunRsaSign interop test".getBytes();
+  private static Stream<Provider> noneWithRsaInteropProviders() {
+    return Stream.of(Security.getProvider("SunJCE"), TestUtil.BC_PROVIDER);
+  }
```
justsmth, line 1697: NP: Should this be "SunRsaSign"?

csrc/sign.cpp
```diff
@@ -740,13 +735,25 @@ JNIEXPORT jboolean JNICALL Java_com_amazon_corretto_crypto_provider_NoneWithRsa_
...
-            result = RSA_verify_pss_mgf1(rsa, digest.data(), hLen, md, mgf_md, saltLen, sig.data(), sig.len());
+            result = RSA_verify_pss_mgf1(rsa, digest.data(), length, md, mgf_md, saltLen, sig.data(), sig.len());
             break;
         }
-        case RSA_PKCS1_PADDING:
-            result = RSA_verify(EVP_MD_type(md), digest.data(), hLen, sig.data(), sig.len(), rsa);
+        case RSA_PKCS1_PADDING: {
+            // Decrypt signature and strip PKCS1 padding
+            size_t outLen = 0;
+            std::vector<uint8_t> buf(RSA_size(rsa));
```
justsmth, line 749: NP: Should this use `SecureAlloc`?

src/com/amazon/corretto/crypto/provider/NoneWithRsa.java
justsmth: NP: Doc update -- for PKCS1, any length up to `rsaSize - 11` is accepted?

csrc/sign.cpp
```diff
@@ -740,13 +735,25 @@ JNIEXPORT jboolean JNICALL Java_com_amazon_corretto_crypto_provider_NoneWithRsa_
         int result = 0;
         switch (paddingType) {
         case RSA_PKCS1_PSS_PADDING: {
+            if (static_cast<size_t>(length) != EVP_MD_size(md)) {
```
justsmth, line 737: NP: Sanity check lengths in "verify" like is done in "sign"?

src/com/amazon/corretto/crypto/provider/NoneWithRsa.java
justsmth: NP: Doc update?

tst/com/amazon/corretto/crypto/provider/test/NoneWithRsaTest.java
```diff
@@ -1757,27 +1751,56 @@ public void testNoneWithRsaCrossAlgorithmInterop() throws Exception {
...
+
+    // However, the underlying signed content shares the same digest bytes.
+    // Decrypt both signatures with raw RSA to recover plaintext, padding stripped.
+    Cipher rsa = Cipher.getInstance("RSA/ECB/PKCS1Padding", ACCP);
+    rsa.init(Cipher.DECRYPT_MODE, kp.getPublic());
+    byte[] nonePlain = rsa.doFinal(noneSig);
+    assertEquals(digest.length, nonePlain.length); // raw SHA-256 digest, 32 bytes
+    rsa.init(Cipher.DECRYPT_MODE, kp.getPublic());
+    byte[] sha256Plain = rsa.doFinal(sha256Sig);
+    assertEquals(digest.length + 19, sha256Plain.length); // +19 bytes for DigestInfo DER
```
justsmth, line 1792: NP: 19 is a magic number?

tst/com/amazon/corretto/crypto/provider/test/NoneWithRsaTest.java
WillChilds-Klein: no. oddly enough, NONEwithRSA comes from SunJCE

https://github.com/corretto/corretto-17/blob/2a7c13cb67f3d1fa63e303091197bdb84b944298/src/java.base/share/classes/com/sun/crypto/provider/SunJCE.java#L149

src/com/amazon/corretto/crypto/provider/NoneWithRsa.java
WillChilds-Klein: ack, will update.

tst/com/amazon/corretto/crypto/provider/test/NoneWithRsaTest.java
WillChilds-Klein: explained in comment on the right

csrc/sign.cpp
WillChilds-Klein: yes, this should be moved into the PKCS1_PADDING case.

src/com/amazon/corretto/crypto/provider/NoneWithRsa.java
WillChilds-Klein: yes, thank you

csrc/sign.cpp
WillChilds-Klein: `buf` stores the signature bytes (public material), so no i don't think so

csrc/sign.cpp
WillChilds-Klein: it's not necessary (verification would just fail), but wouldn't hurt. added.

justsmth (approved):

tst/com/amazon/corretto/crypto/provider/test/NoneWithRsaTest.java
justsmth: Should there be a named constant for the 19?

csrc/sign.cpp
justsmth: `sig` is the signature bytes?

tst/com/amazon/corretto/crypto/provider/test/NoneWithRsaTest.java
WillChilds-Klein: It's explained in a comment and not reused anywhere, so IMO no.

csrc/sign.cpp
WillChilds-Klein: poor phrasing on my part. `RSA_verify_raw` will do RSA exponentiation with the public key, thereby "decrypting" the signature's digest into `buf` for comparison against `digest`. because neither the signature nor public key are private, we don't need a `SecureAlloc` here.

src/com/amazon/corretto/crypto/provider/NoneWithRsa.java
```diff
@@ -185,41 +190,33 @@ protected boolean isBufferEmpty() {
...
-   * The MGF, salt length, and trailer fields are ignored since they are PSS-specific and have no
-   * meaning for RSASSA-PKCS1-v1_5 signatures. This allows callers to use a uniform parameter
-   * interface across both padding modes.
-   *
    * @throws InvalidAlgorithmParameterException if the parameter type is not supported or the digest
    *     algorithm is unrecognized
    * @throws IllegalStateException if called while the buffer contains data
    */
   @Override
   protected synchronized void engineSetParameter(final AlgorithmParameterSpec params)
```
geedo0, line 198: So there is a compatibility change introduced with this PR for how NONEwithRSA is handled. AFAICT from all our release paths, there has been no release that contained this. Are you aware of any customers that may have integrated the old behavior?

geedo0 (approved):

src/com/amazon/corretto/crypto/provider/NoneWithRsa.java
WillChilds-Klein: > So there is a compatibility change introduced with this PR for how NONEwithRSA is handled. AFAICT from all our release paths, there has been no release that contained this.

Correct.

> Are you aware of any customers that may have integrated the old behavior?

No.

PR #537, opened 2026-04-20
Implement NONEwithEd25519ph

WillChilds-Klein:
*Issue #, if available:* t/P415299769

*Description of changes:*

# Notes

This commit implements a non-[JCA-standard](https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html) variant of Ed25519ph that skips the inner SHA512 hash of the message (i.e. `PH(M)`) [specified](https://www.rfc-editor.org/rfc/rfc8032.html#section-5.1) in RFC 8032:

```
   For Ed25519ph, phflag=1 and PH is SHA512 instead.  That is, the input
   is hashed using SHA-512 before signing with Ed25519.
...
5.1.6.  Sign

   The inputs to the signing procedure is the private key, a 32-octet
   string, and a message M of arbitrary size.
...
   2.  Compute SHA-512(dom2(F, C) || prefix || PH(M)), where M is the
       message to be signed.  Interpret the 64-octet digest as a little-
       endian integer r.
...
```

NIST SP 186-5 [similarly stipulates](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf#%5B%7B%22num%22%3A122%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C70%2C700%2C0%5D) this inner hash for Ed25519ph.

Because the inner hash is skipped, the caller is responsible for hashing the message with SHA512 before passing that digest into `NONEwithEd25519ph`. If the caller requires FIPS compliance, the caller is responsible for ensuring proper validation of the cryptographic module used for their SHA512 hashing.

This variant is compatible with golang's Ed25519ph [implementation](https://pkg.go.dev/crypto/ed25519#PrivateKey.Sign) which also skips the inner SHA512 call.

# Testing
- Added KAT coverage from RFC 8032
- interop test with regular ACCP's Ed25519ph, transitively with other providers through existing interop tests

---


skmcgrail (dismissed):

src/com/amazon/corretto/crypto/provider/EvpSignatureRaw.java
```diff
@@ -120,6 +120,12 @@ static final class Ed25519 extends EvpSignatureRaw {
     }
   }
 
+  static final class NONEwithEd25519ph extends EvpSignatureRaw {
```
geedo0, line 127: Maybe there should be a documentation comment here on what exactly this implements and how to use it? Also, we should update the4 algorithm list in the README.md.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
```diff
@@ -522,6 +567,122 @@ public void testNullInputs() throws GeneralSecurityException {
...
+    }
+  }
+
+  @Test
+  public void noneWithEd25519phRejectsWrongInputLength() throws GeneralSecurityException {
+    assumeTrue(ed25519phIsEnabled());
+    final KeyPair kp = nativeGen.generateKeyPair();
+    final Signature sig = Signature.getInstance("NONEwithEd25519ph", NATIVE_PROVIDER);
+
+    for (int badLength : new int[] {0, 1, 32, 63, 65, 128, 1024}) {
```
geedo0: nit: A bit over the top, but I guess no harm.

geedo0 (commented):
Just want the README update before approving.

tst/com/amazon/corretto/crypto/provider/test/EdDSATest.java
WillChilds-Klein: yeah, agreed. i'll cut this down.

src/com/amazon/corretto/crypto/provider/EvpSignatureRaw.java
WillChilds-Klein: good call.

geedo0 (approved):

skmcgrail (approved):

PR #540, opened 2026-04-29
Add DigestUtils.digestInfoWrap static utility

WillChilds-Klein:
*Issue #, if available:*

*Description of changes:*

This PR provides a static utility function `DigestUtils.digestInfoWrap` to wrap a digest in a `DigestInfo` DER encoding wrapper per RFC 8017 Sec. 9.2. We provide a few basic KAT tests for common digest algorithms.

---

### DIFF
--- a/csrc/util_class.cpp
+++ b/csrc/util_class.cpp
@@ -254,6 +254,53 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_utils_MlDsaUtils_co
 #endif // !defined(FIPS_BUILD) || defined(EXPERIMENTAL_FIPS_BUILD)
 
 /*
+ * Class:     com_amazon_corretto_crypto_utils_DigestUtils
+ * Method:    digestInfoWrapInternal
+ * Signature: (Ljava/lang/String;[B)[B
+ */
+JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_utils_DigestUtils_digestInfoWrapInternal(
+    JNIEnv* pEnv, jclass, jstring digestName, jbyteArray digestBytes)
+{
+    jbyteArray result = nullptr;
+    try {
+        raii_env env(pEnv);
+        if (!digestName) {
+            throw_java_ex(EX_ILLEGAL_ARGUMENT, "digestName must not be null");
+        }
+        jni_string md_name(env, digestName);
+        const EVP_MD* md = EVP_get_digestbyname(md_name.native_str);
+        if (!md) {
+            throw_java_ex(EX_ILLEGAL_ARGUMENT, "Unsupported digest algorithm");
+        }
+        const int hash_nid = EVP_MD_type(md);
+
+        jsize digest_len = env->GetArrayLength(digestBytes);
+        jbyte* digest = env->GetByteArrayElements(digestBytes, nullptr);
+        CHECK_OPENSSL(digest);
+
+        uint8_t* out_msg = nullptr;
+        size_t out_msg_len = 0;
+        int is_alloced = 0;
+        if (1 != RSA_add_pkcs1_prefix(
+                &out_msg, &out_msg_len, &is_alloced, hash_nid, (const uint8_t*)digest, (size_t)digest_len)) {
+            throw_java_ex(EX_ILLEGAL_ARGUMENT, "RSA_add_pkcs1_prefix failed");
+        }
+        if (!(result = env->NewByteArray((jsize)out_msg_len))) {
+            throw_java_ex(EX_OOM, "Unable to allocate DigestInfo array");
+        }
+        env->SetByteArrayRegion(result, 0, (jsize)out_msg_len, (const jbyte*)out_msg);
+        if (is_alloced && out_msg) {
+            OPENSSL_free(out_msg);
+        }
+        env->ReleaseByteArrayElements(digestBytes, digest, JNI_ABORT);
+    } catch (java_ex& ex) {
+        ex.throw_to_java(pEnv);
+        return nullptr;
+    }
+    return result;
+}
+
+/*
  * Class:     com_amazon_corretto_crypto_utils_EcUtils
  * Method:    encodeRfc5915EcPrivateKeyInternal
  * Signature: ([B)[B
--- a/csrc/util_class.cpp
+++ b/csrc/util_class.cpp
@@ -273,7 +273,8 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_utils_DigestUtils_d
         jni_string md_name(env, digestName);
         const EVP_MD* md = EVP_get_digestbyname(md_name.native_str);
         if (md == nullptr) {
-            throw_java_ex(EX_ILLEGAL_ARGUMENT, "Unsupported digest algorithm");
+            throw_java_ex(
+                EX_ILLEGAL_ARGUMENT, std::string("Unsupported digest algorithm: ") + md_name.native_str);
         }
         const int hash_nid = EVP_MD_type(md);
 
--- a/csrc/util_class.cpp
+++ b/csrc/util_class.cpp
@@ -298,13 +298,13 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_utils_DigestUtils_d
             }
             env->SetByteArrayRegion(result, 0, (jsize)out_msg_len, (const jbyte*)out_msg);
         } catch (...) {
-            if (is_alloced && out_msg) {
+            if (is_alloced) {
                 OPENSSL_free(out_msg);
             }
             env->ReleaseByteArrayElements(digestBytes, digest, JNI_ABORT);
             throw;
         }
-        if (is_alloced && out_msg) {
+        if (is_alloced) {
             OPENSSL_free(out_msg);
         }
         env->ReleaseByteArrayElements(digestBytes, digest, JNI_ABORT);
--- a/csrc/util_class.cpp
+++ b/csrc/util_class.cpp
@@ -17,8 +17,10 @@ extern "C" {
  * Signature: (Ljava/nio/ByteBuffer;Ljava/nio/ByteBuffer;)J
  */
 
-JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_Utils_getNativeBufferOffset(
-    JNIEnv* env, jclass, jobject bufA, jobject bufB)
+JNIEXPORT jlong JNICALL Java_com_amazon_corretto_crypto_provider_Utils_getNativeBufferOffset(JNIEnv* env,
+                                                                                             jclass,
+                                                                                             jobject bufA,
+                                                                                             jobject bufB)
 {
     const jlong JINT_MAX = (1L << 31) - 1L;
     const jlong JINT_MIN = -(1L << 31);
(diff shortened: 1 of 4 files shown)


src/com/amazon/corretto/crypto/provider/Utils.java
```diff
@@ -69,12 +69,17 @@ static int getDigestLength(final String digestName) {
     return getDigestLength(getEvpMdFromName(digestName));
   }
 
-  static String jceDigestNameToAwsLcName(final String jceName) {
+  public static String jceDigestNameToAwsLcName(final String jceName) {
```
geedo0: nit: The logic here seems pretty brittle, maybe we can do something more regex based to constrain it? Would be nice to be able to identify validation issues here and reject them before going to the native layer.

csrc/util_class.cpp
```diff
@@ -253,6 +256,65 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_utils_MlDsaUtils_co
...
+        if (digest == nullptr) {
+            throw_java_ex(EX_OOM, "Unable to pin or copy digestBytes array");
+        }
+
+        uint8_t* out_msg = nullptr;
+        size_t out_msg_len = 0;
+        int is_alloced = 0;
+        try {
+            if (1
+                != RSA_add_pkcs1_prefix(&out_msg, &out_msg_len, &is_alloced, hash_nid, (const uint8_t*)digest,
```
geedo0, line 292: Does `digest_len` get checked anywhere? And is there no consequence if the user passes in some bogus lengthed array?

I see that you have a test for this and it allegedly passes, so I assume this exists.

geedo0 (dismissed):

csrc/util_class.cpp
WillChilds-Klein: > Does digest_len get checked anywhere? And is there no consequence if the user passes in some bogus lengthed array?

Yes. It's checked [here](https://github.com/aws/aws-lc/blob/6af5337f678501c9ec87dd86ab3f42c8fe5717da/crypto/fipsmodule/rsa/rsa.c#L737-L753) in `RSA_add_pkcs1_prefix`.

src/com/amazon/corretto/crypto/provider/Utils.java
WillChilds-Klein: I expect these AWS-LC digest names to be stable. For new names that may diverge from this convention in the future, a "digest not supported" error will alert us/the user of the need to modify this implementation detail.

To ensure this remains an implementation detail, I'll remove the Public qualifiers from this class and method.

src/com/amazon/corretto/crypto/utils/DigestUtils.java
```diff
@@ -0,0 +1,56 @@
...
+    if (digestBytes == null) {
+      throw new IllegalArgumentException("digestBytes must not be null");
+    }
+    return digestInfoWrapInternal(normalizeDigestName(digestName), digestBytes);
+  }
+
+  private static String normalizeDigestName(final String jceName) {
+    if (jceName == null) {
+      return null;
+    }
```
justsmth: NP: It's fine to keep, but the `jceName == null` is not needed as this is guarded by the caller, which is also a `static` function. (Returning `null` here would propagate into `digestInfoWrapInternal` causing a very ambiguous error to surface.)

src/com/amazon/corretto/crypto/provider/Utils.java
justsmth, line 78: Should uses of this function be switched to using `normalizeDigestName` -- this is missing the SHA3 special cases.

src/com/amazon/corretto/crypto/utils/DigestUtils.java
```diff
@@ -0,0 +1,56 @@
...
+    final String upper = jceName.toUpperCase();
```
justsmth, line 45: NP: `jceName.toUpperCase(Locale.ROOT)`?

csrc/util_class.cpp
```diff
@@ -253,6 +256,65 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_utils_MlDsaUtils_co
...
+    try {
+        raii_env env(pEnv);
+        if (digestName == nullptr) {
+            throw_java_ex(EX_ILLEGAL_ARGUMENT, "digestName must not be null");
+        }
+        jni_string md_name(env, digestName);
+        const EVP_MD* md = EVP_get_digestbyname(md_name.native_str);
+        if (md == nullptr) {
+            throw_java_ex(EX_ILLEGAL_ARGUMENT, "Unsupported digest algorithm");
+        }
```
justsmth, line 278: NPs: 
* `ERR_clear_error` prior to throw -- symmetry with `RSA_add_pkcs1_prefix` below?
* Improve error message: "Unsupported digest algorithm: " + md_name.native_st

csrc/util_class.cpp
```diff
@@ -253,6 +256,65 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_utils_MlDsaUtils_co
...
+        try {
+            if (1
+                != RSA_add_pkcs1_prefix(&out_msg, &out_msg_len, &is_alloced, hash_nid, (const uint8_t*)digest,
+                                        (size_t)digest_len)) {
+                ERR_clear_error();
+                throw_java_ex(EX_ILLEGAL_ARGUMENT, "RSA_add_pkcs1_prefix failed");
+            }
+            if (!(result = env->NewByteArray((jsize)out_msg_len))) {
+                throw_java_ex(EX_OOM, "Unable to allocate DigestInfo array");
+            }
```
justsmth, line 299: When `NewByteArray` returns null, JNI has already set `OutOfMemoryError` as pending?

tst/com/amazon/corretto/crypto/provider/test/DigestUtilsTest.java
```diff
@@ -0,0 +1,313 @@
...
+  public void testNullDigestName() {
+    assertThrows(
+        IllegalArgumentException.class, () -> DigestUtils.digestInfoWrap(null, new byte[32]));
+  }
+
+  @Test
+  public void testNullDigestBytes() {
+    assertThrows(IllegalArgumentException.class, () -> DigestUtils.digestInfoWrap("SHA-256", null));
+  }
+
```
justsmth, line 356: NP: Worth adding a case like `DigestUtils.digestInfoWrap("sha-256", …)` and asserting it produces the same output as the `"SHA-256"` variant, so we catch it if someone ever removes the `toUpperCase` call?

tst/com/amazon/corretto/crypto/provider/test/DigestUtilsTest.java
```diff
@@ -0,0 +1,313 @@
...
+    0x03,
+    0x04,
+    0x02,
+    0x0a,
+    0x05,
+    0x00,
+    0x04,
+    0x40
+  };
+
```
justsmth, line 241: Also include "SHA-224" and "SHA-512/224" prefixes?

tst/com/amazon/corretto/crypto/provider/test/DigestUtilsTest.java
```diff
@@ -0,0 +1,313 @@
...
+    byte[] wrapped = DigestUtils.digestInfoWrap("SHA-512", digest);
+    assertArrayEquals(concat(SHA512_PREFIX, digest), wrapped);
+  }
+
+  @Test
+  public void testSha512_256() throws Exception {
+    byte[] digest = MessageDigest.getInstance("SHA-512/256").digest("hello".getBytes());
+    byte[] wrapped = DigestUtils.digestInfoWrap("SHA-512/256", digest);
+    assertArrayEquals(concat(SHA512_256_PREFIX, digest), wrapped);
+  }
```
justsmth, line 302: Does this fall through to the SUN provider? ACCP doesn't register `SHA-512/256` (or SHA-224, or any SHA-512/t variant) as a [`MessageDigest` service](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java#L86-L91)?
```
  private void buildServiceMap() {
    addService("MessageDigest", "SHA-512", "SHA512Spi");
    addService("MessageDigest", "SHA-384", "SHA384Spi");
    addService("MessageDigest", "SHA-256", "SHA256Spi");
    addService("MessageDigest", "SHA-1", "SHA1Spi");
    addService("MessageDigest", "MD5", "MD5Spi");
...
```

WillChilds-Klein: yes, but it doesn't really matter. any correctly sized byte array works for this test.

src/com/amazon/corretto/crypto/utils/DigestUtils.java
WillChilds-Klein: that's true, this can be removed.

src/com/amazon/corretto/crypto/provider/Utils.java
WillChilds-Klein: i considered this, but decided "no" because that would require making this class public (different java package from DigestUtils) and I'd prefer not to expose it. i don't think we should modify this method today to handle SHA3 because ACCP doesn't support SHA3 and that would be dead code.

src/com/amazon/corretto/crypto/utils/DigestUtils.java
WillChilds-Klein: I also saw this in my local claude review :) there are other instances of `toUpperCase()` and if l10n/i18n is an issue we should do them all in a batch.

csrc/util_class.cpp
WillChilds-Klein: also came up in local claude review, but `EVP_get_digestbyname` doesn't set an error so `ERR_clear_error` is moot.

csrc/util_class.cpp
WillChilds-Klein: maybe? i'm not sure that's guaranteed by JDK's public contracts. either way, not sure what other exception we'd throw here.

tst/com/amazon/corretto/crypto/provider/test/DigestUtilsTest.java
WillChilds-Klein: yeah, good call

tst/com/amazon/corretto/crypto/provider/test/DigestUtilsTest.java
WillChilds-Klein: will call this out in a test comment.

tst/com/amazon/corretto/crypto/provider/test/DigestUtilsTest.java
WillChilds-Klein: sure, why not

csrc/util_class.cpp
WillChilds-Klein: will adjust error message

justsmth (dismissed):

justsmth (approved):

src/com/amazon/corretto/crypto/provider/Utils.java
geedo0: Thanks for keeping this logic private.

geedo0 (approved):

PR #542, opened 2026-05-07
Re-init AES-GCM native context when IV length changes

samuel40791765:
*Issue #, if available:*
  
  `V2200543407`

*Description of changes:*
  
`AesGcmSpi.checkKeyIvPair` decided `sameKey` from the key bytes alone, so re-initializing a `Cipher` with the same key but an IV with a different length took the cached-context fast path in the native layer and kept the prior `ivlen`.
This would either read past the new IV buffer (shorter new IV) or silently truncates the nonce (longer new IV, causing nonce reuse when IVs share a prefix).
  
Treat a differing IV length as a non-match so the fast path is skipped and the native layer runs a full re-init. Two regression tests added to `AesTest` covering both directions.

### DIFF
--- a/tst/com/amazon/corretto/crypto/provider/test/AesTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/AesTest.java
@@ -397,6 +397,34 @@ public class AesTest {
     }
   }
 
+  /**
+   * V2200543407: re-initializing a Cipher with the same key but a different-length IV must produce
+   * the same output as a fresh Cipher. The native EVP_CIPHER_CTX caches its ivlen from the prior
+   * init; if reused, a different-length IV would be truncated or read past the buffer. The two
+   * primer cycles get the native context cached so the regression step exercises that reuse path;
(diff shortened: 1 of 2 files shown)


WillChilds-Klein (dismissed):

geedo0 (approved):

tst/com/amazon/corretto/crypto/provider/test/AesTest.java
```diff
@@ -397,6 +397,34 @@ public void large_overlap_encrypt() {
     }
   }
 
+  /**
+   * V2200543407: re-initializing a Cipher with the same key but a different-length IV must produce
+   * the same output as a fresh Cipher. The native EVP_CIPHER_CTX caches its ivlen from the prior
+   * init; if reused, a different-length IV would be truncated or read past the buffer. The two
+   * primer cycles get the native context cached so the regression step exercises that reuse path;
```
WillChilds-Klein, line 404: doesn't just the first "cycle" cache the context?

WillChilds-Klein (approved):

PR #543, opened 2026-05-19
Add support for PBKDF2

prasden:
Issue #:  *ACCP-77*

*Description of changes:*

This PR adds support for PBKDF2 in ACCP for the five JCA digests `PBKDF2WithHmacSHA1/224/256/384/512` backed by AWS-LC's `PKCS5_PBKDF2_HMAC`. The new SPI `Pbkdf2SecretKeyFactorySpi.java` follows existing KDF patterns.

*Testing* 

Tests include KATs  from AWS-LC's  [pbkdf_test.cc](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/pbkdf/pbkdf_test.cc) covering SHA-1, SHA-256, and SHA-512.

Four SHA-224/SHA-384 vectors come from [Anti-weakpasswords/PBKDF2-GCC-OpenSSL-library](https://github.com/Anti-weakpasswords/PBKDF2-GCC-OpenSSL-library/blob/master/pbkdf2_test.sh).  Each vector in the repo's `pbkdf2_test.sh` script lists  a set of inputs (password, salt, iteration count, output length, digest) and the expected output bytes for that input. Before adding the four vectors from the repo, they were double-checked by building the upstream repo's `pbkdf2` CLI from source. For each of the four entries, the CLI was run with the same password, salt, iteration count, output length, and digest from the upstream `pbkdf2_test.sh` script, to ensure the bytes were identical with the expected values. Python's `hashlib.pbkdf2_hmac` was also used to cross-check. 

Remaining tests check ACCP's interoperability, JCA registration verification, and bad input checks. 

### DIFF
--- a/csrc/pbkdf2.cpp
+++ b/csrc/pbkdf2.cpp
@@ -1,0 +1,37 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+#include "buffer.h"
+#include "env.h"
+#include "generated-headers.h"
+#include <openssl/evp.h>
+
+using namespace AmazonCorrettoCryptoProvider;
+
+extern "C" JNIEXPORT void JNICALL
+Java_com_amazon_corretto_crypto_provider_Pbkdf2SecretKeyFactorySpi_pbkdf2(JNIEnv* env,
+                                                                          jclass,
+                                                                          jbyteArray jPassword,
+                                                                          jint passwordLen,
+                                                                          jbyteArray jSalt,
+                                                                          jint saltLen,
+                                                                          jint iterations,
+                                                                          jint digestCode,
+                                                                          jbyteArray jOutput,
+                                                                          jint outputLen)
+{
+    try {
+        JByteArrayCritical password(env, jPassword);
+        JByteArrayCritical salt(env, jSalt);
+        JByteArrayCritical output(env, jOutput);
+        EVP_MD const* digest = digest_code_to_EVP_MD(digestCode);
+
+        if (PKCS5_PBKDF2_HMAC(reinterpret_cast<const char*>(password.get()), passwordLen, salt.get(), saltLen,
+                              iterations, digest, outputLen, output.get())
+            != 1) {
+            throw_openssl(EX_RUNTIME_CRYPTO, "PBKDF2 failed.");
+        }
+
+    } catch (java_ex& ex) {
+        ex.throw_to_java(env);
+    }
+}
--- a/csrc/pbkdf2.cpp
+++ b/csrc/pbkdf2.cpp
@@ -28,7 +28,7 @@ Java_com_amazon_corretto_crypto_provider_Pbkdf2SecretKeyFactorySpi_pbkdf2(JNIEnv
         if (PKCS5_PBKDF2_HMAC(reinterpret_cast<const char*>(password.get()), passwordLen, salt.get(), saltLen,
                               iterations, digest, outputLen, output.get())
             != 1) {
-            throw_openssl(EX_RUNTIME_CRYPTO, "PBKDF2 failed.");
+            throw_openssl(EX_RUNTIME_CRYPTO, "Heap allocation in PBKDF2 failed.");
         }
 
     } catch (java_ex& ex) {
--- a/src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
+++ b/src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
@@ -47,14 +47,8 @@ class Pbkdf2SecretKeyFactorySpi extends KdfSpi {
     final int iterations = spec.getIterationCount();
     final int keyLengthBits = spec.getKeyLength();
 
-    if (password == null) {
-      throw new InvalidKeySpecException("Password cannot be null");
-    }
     if (salt == null) {
       throw new InvalidKeySpecException("Salt cannot be null");
-    }
-    if (iterations < 1) {
-      throw new InvalidKeySpecException("Positive iteration count required");
     }
     if (keyLengthBits <= 0) {
       throw new InvalidKeySpecException("Positive key length required");
@@ -89,8 +83,7 @@ class Pbkdf2SecretKeyFactorySpi extends KdfSpi {
     final int len = bb.limit();
     final byte[] passwordBytes = new byte[len];
     bb.get(passwordBytes, 0, len);
-    bb.clear();
-    bb.put(new byte[len]);
+    Arrays.fill(bb.array(), bb.arrayOffset(), bb.arrayOffset() + bb.capacity(), (byte) 0);
     return passwordBytes;
   }
 
--- a/src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
+++ b/src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
@@ -1,0 +1,129 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider;
+
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.StandardCharsets;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.KeySpec;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.PBEKeySpec;
+import javax.crypto.spec.SecretKeySpec;
+
+class Pbkdf2SecretKeyFactorySpi extends KdfSpi {
+
+  private final int digestCode;
+  private final String algorithmName;
+
+  private Pbkdf2SecretKeyFactorySpi(final int digestCode, final String algorithmName) {
+    this.digestCode = digestCode;
+    this.algorithmName = algorithmName;
+  }
+
+  private static native void pbkdf2(
+      byte[] jPassword,
+      int passwordLen,
+      byte[] jSalt,
+      int saltLen,
+      int iterations,
+      int digestCode,
+      byte[] jOutput,
+      int outputLen);
+
+  @Override
+  protected SecretKey engineGenerateSecret(final KeySpec keySpec) throws InvalidKeySpecException {
+    if (!(keySpec instanceof PBEKeySpec)) {
+      throw new InvalidKeySpecException("KeySpec must be an instance of PBEKeySpec");
+    }
+
+    final PBEKeySpec spec = (PBEKeySpec) keySpec;
+    final char[] password = spec.getPassword();
+    final byte[] salt = spec.getSalt();
+    final int iterations = spec.getIterationCount();
+    final int keyLengthBits = spec.getKeyLength();
+
+    if (password == null) {
+      throw new InvalidKeySpecException("Password cannot be null");
+    }
+    if (salt == null) {
+      throw new InvalidKeySpecException("Salt cannot be null");
+    }
+    if (iterations < 1) {
+      throw new InvalidKeySpecException("Positive iteration count required");
+    }
+    if (keyLengthBits <= 0) {
+      throw new InvalidKeySpecException("Positive key length required");
+    }
+
+    final byte[] passwordBytes = charsToBytes(password);
+    final byte[] derivedKey = new byte[keyLengthBits / 8];
+
+    try {
+      pbkdf2(
+          passwordBytes,
+          passwordBytes.length,
+          salt,
+          salt.length,
+          iterations,
+          digestCode,
+          derivedKey,
+          derivedKey.length);
+      return new SecretKeySpec(derivedKey, algorithmName);
+    } finally {
+      Arrays.fill(passwordBytes, (byte) 0);
+      Arrays.fill(derivedKey, (byte) 0);
+    }
+  }
+
+  // PBKDF2 takes the password as a byte string, but PBEKeySpec holds
+  // it as a char array, so we must convert to bytes before calling into PBKDF2. We use
+  // UTF-8 to match SunJCE so keys are identical across providers for any PBEKeySpec input.
+  // https://github.com/openjdk/jdk/blob/a73eca9e8b1b96925b4b4d4ccfeffe9891fd8ce1/src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java#L48-L57
+  private static byte[] charsToBytes(final char[] password) {
+    final ByteBuffer bb = StandardCharsets.UTF_8.encode(CharBuffer.wrap(password));
+    final int len = bb.limit();
+    final byte[] passwordBytes = new byte[len];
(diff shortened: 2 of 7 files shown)


src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
```diff
@@ -0,0 +1,129 @@
...
+
+    if (password == null) {
+      throw new InvalidKeySpecException("Password cannot be null");
+    }
+    if (salt == null) {
+      throw new InvalidKeySpecException("Salt cannot be null");
+    }
+    if (iterations < 1) {
+      throw new InvalidKeySpecException("Positive iteration count required");
+    }
```
WillChilds-Klein: can we add coverage for these cases?

csrc/pbkdf2.cpp
```diff
@@ -0,0 +1,37 @@
...
+                                                                          jbyteArray jOutput,
+                                                                          jint outputLen)
+{
+    try {
+        JByteArrayCritical password(env, jPassword);
+        JByteArrayCritical salt(env, jSalt);
+        JByteArrayCritical output(env, jOutput);
+        EVP_MD const* digest = digest_code_to_EVP_MD(digestCode);
+
+        if (PKCS5_PBKDF2_HMAC(reinterpret_cast<const char*>(password.get()), passwordLen, salt.get(), saltLen,
```
WillChilds-Klein, line 28: minor: the header for `PKCS5_PBKDF2_HMAC` [reads](https://github.com/aws/aws-lc/blob/39d3830a15e9ae71499f26bacd77db09da54dc15/include/openssl/evp.h#L493-L498):

```
// PKCS5_PBKDF2_HMAC computes |iterations| iterations of PBKDF2 of |password|
// and |salt|, using |digest|, and outputs |key_len| bytes to |out_key|. It
// returns one on success and zero on allocation failure or if |iterations| is
// 0. It's recommended that |iterations| be set to a much higher number (at
// least hundreds of thousands).
```

you already enforce `iterations > 0` in the java logic. suggest modifying the exception method to `Heap allocation in PBKDF2 failed` or similar

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
```diff
@@ -0,0 +1,129 @@
...
+    if (keyLengthBits <= 0) {
+      throw new InvalidKeySpecException("Positive key length required");
+    }
+
+    final byte[] passwordBytes = charsToBytes(password);
+    final byte[] derivedKey = new byte[keyLengthBits / 8];
```
WillChilds-Klein, line 58: should we validate that `keyLengthBits` is divisible by 8 or does AWS-LC do that for us?

csrc/pbkdf2.cpp
```diff
@@ -0,0 +1,37 @@
...
+                              iterations, digest, outputLen, output.get())
+            != 1) {
+            throw_openssl(EX_RUNTIME_CRYPTO, "PBKDF2 failed.");
```
WillChilds-Klein: do we need to zeroize `output` and `password` before throwing here?

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
```diff
@@ -0,0 +1,129 @@
...
+  // it as a char array, so we must convert to bytes before calling into PBKDF2. We use
+  // UTF-8 to match SunJCE so keys are identical across providers for any PBEKeySpec input.
+  // https://github.com/openjdk/jdk/blob/a73eca9e8b1b96925b4b4d4ccfeffe9891fd8ce1/src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java#L48-L57
+  private static byte[] charsToBytes(final char[] password) {
+    final ByteBuffer bb = StandardCharsets.UTF_8.encode(CharBuffer.wrap(password));
+    final int len = bb.limit();
+    final byte[] passwordBytes = new byte[len];
+    bb.get(passwordBytes, 0, len);
+    bb.clear();
+    bb.put(new byte[len]);
```
WillChilds-Klein: what's the purpose of this allocation?

you call `clear()` on the line above, which [doesn't zeroize](https://docs.oracle.com/javase/8/docs/api/java/nio/Buffer.html#clear--). are you trying to zeroize here? if so, i don't think this allocation accomplishes that. you could use an array-backed ByteBuffer and do an `Arrays.fill(..., 0)` on the backing array.

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
```diff
@@ -0,0 +1,129 @@
...
+      return new SecretKeySpec(derivedKey, algorithmName);
+    } finally {
+      Arrays.fill(passwordBytes, (byte) 0);
+      Arrays.fill(derivedKey, (byte) 0);
+    }
+  }
+
+  // PBKDF2 takes the password as a byte string, but PBEKeySpec holds
+  // it as a char array, so we must convert to bytes before calling into PBKDF2. We use
+  // UTF-8 to match SunJCE so keys are identical across providers for any PBEKeySpec input.
```
WillChilds-Klein, line 80: should we add a SunJCE compatibility test?

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
```diff
@@ -0,0 +1,129 @@
...
+    result.put(
+        getSpiFactoryForAlgName(PBKDF2_WITH_SHA384),
+        new Pbkdf2SecretKeyFactorySpi(Utils.SHA384_CODE, PBKDF2_WITH_SHA384));
+    result.put(
+        getSpiFactoryForAlgName(PBKDF2_WITH_SHA512),
+        new Pbkdf2SecretKeyFactorySpi(Utils.SHA512_CODE, PBKDF2_WITH_SHA512));
+    return Collections.unmodifiableMap(result);
+  }
+
+  static String getSpiFactoryForAlgName(final String alg) {
```
WillChilds-Klein, line 120: nit: should we just call `toUpperCase` in the `put` calls above?

WillChilds-Klein (commented):
Some minor comments, looking good!

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
prasden: We can, but looking at [HKDFSecretKeyFactorySpi](https://github.com/corretto/amazon-corretto-crypto-provider/blob/588c4f3990d68e0e66adb588f1df9c2fff806a34/src/com/amazon/corretto/crypto/provider/HkdfSecretKeyFactorySpi.java#L133) and [CounterKdfSpi](https://github.com/corretto/amazon-corretto-crypto-provider/blob/588c4f3990d68e0e66adb588f1df9c2fff806a34/src/com/amazon/corretto/crypto/provider/CounterKdfSpi.java#L64), this seemed to be established convention. I initially had a refactor as part of this PR to refactor this pattern into a shared helper in `Utils.java` but saw that it added some noise to this PR. I'm not strongly opposed to making a change for your suggestion but my stance is that this change is best for a future refactor PR.

csrc/pbkdf2.cpp
prasden: No, I don't believe so. The `finally` block in `engineGenerateSecret` in our SPI implementation zeros both `passwordBytes` and `derivedKey`, both of which are accessed in this layer with `JByteArrayCritical`. Zeroing here as well would be redundant, unless you suggest switching to zeroize in our cpp layer instead of the java layer.

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
prasden: After rereading [PBEKeySpec's](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/javax/crypto/spec/PBEKeySpec.java) implementation, I found that the null password check was unnecessary as all constructors for `PBEKeySpec` initialize `password` as an empty char array if null/empty length `password` is passed in.  

The iterations < 1 check is similarly unreachable as the 3 and 4 arg constructors throw  `IllegalArgumentException` if `iterationCount` <= 0, and the 1 arg constructor leaves `iterationCount`  at 0 but also leaves `salt` at null, so the `salt == null` check fires first. 

Coverage for the other branches are in `invalidParameterTests`

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
prasden: Added an encoding test to ensure compatibility.

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
prasden: Neither AWS-LC nor SunJCE/BouncyCastle validate divisibility by 8 and silently truncate, so we match that behavior to keep parity.

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
prasden: The intention was to zeroize to leave nothing in the buffers. I replaced `bb.clear()` with your suggestion.

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
WillChilds-Klein: that works.

csrc/pbkdf2.cpp
WillChilds-Klein: I guess the question behind my question was whether the native buffers and java buffers point to the same memory regions, or if there's copying involved. JByteArrayCritical's [ctor](https://github.com/corretto/amazon-corretto-crypto-provider/blob/588c4f3990d68e0e66adb588f1df9c2fff806a34/csrc/buffer.cpp#L40) calls JNI's `GetPrimitiveArrayCritical`, the docs for which [state](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#GetPrimitiveArrayCritical_ReleasePrimitiveArrayCritical):

>  If possible, the VM returns a pointer to the primitive array; otherwise, a copy is made.

So if we get a copy from the JVM, the java layer only zeroizes the original buffer that's _copied from_ and the native buffer isn't zeroized [before it's released](https://github.com/corretto/amazon-corretto-crypto-provider/blob/588c4f3990d68e0e66adb588f1df9c2fff806a34/csrc/buffer.cpp#L50C51-L50C80).

It doesn't look like we have a good way to zeroize the native buffer in the copy case. [HKDF](https://github.com/corretto/amazon-corretto-crypto-provider/blob/588c4f3990d68e0e66adb588f1df9c2fff806a34/csrc/hkdf.cpp#L24-L37) has the same behavior, I'll look into ways to address this outside the scope of this PR.

WillChilds-Klein (dismissed):

prasden:
Codebuild CI on ubuntu machines was previously failing due to the literal non-ASCII characters(á, ó). They are now changed to use escape sequences instead to test compatibility with SunJCE.

WillChilds-Klein (dismissed):

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
```diff
@@ -0,0 +1,122 @@
...
+      int outputLen);
+
+  @Override
+  protected SecretKey engineGenerateSecret(final KeySpec keySpec) throws InvalidKeySpecException {
+    if (!(keySpec instanceof PBEKeySpec)) {
+      throw new InvalidKeySpecException("KeySpec must be an instance of PBEKeySpec");
+    }
+
+    final PBEKeySpec spec = (PBEKeySpec) keySpec;
+    final char[] password = spec.getPassword();
```
justsmth: * `PBEKeySpec.getPassword()` returns a cloned `char[]`, so this method owns that temporary copy and should wipe it before returning. 
* Also, the cleanup scope probably needs to start immediately after `getPassword()`: as written, any later validation failure (for example `salt == null` or invalid key length) exits before cleanup runs. Could we move the rest of the method under a single `try { ... } finally { ... }` and explicitly zero the password chars there?

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
```diff
@@ -0,0 +1,122 @@
...
+      pbkdf2(
+          passwordBytes,
+          passwordBytes.length,
+          salt,
+          salt.length,
+          iterations,
+          digestCode,
+          derivedKey,
+          derivedKey.length);
+      return new SecretKeySpec(derivedKey, algorithmName);
```
justsmth, line 70: Would it be better to return `Pbkdf2Key` to preserve the PBE metadata instead of only the derived bytes?

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
```diff
@@ -0,0 +1,122 @@
...
+import java.security.spec.KeySpec;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.PBEKeySpec;
+import javax.crypto.spec.SecretKeySpec;
+
+class Pbkdf2SecretKeyFactorySpi extends KdfSpi {
```
justsmth, line 18: This SPI inherits from `KdfSpi` -- should this also implement `engineGetKeySpec` and `engineTranslateKey`?

prasden: No, we do not need to. `KdfSpi` defines both methods to throw `UnsupportedOperationException.` PBKDF2 is a one way algorithm so we cannot recover metadata from `engineGetKeySpec().` We currently return a standard `SecretKeySpec` so `engineGetKeySpec()` has nothing to translate into as there is no ACCP specific key.

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
prasden: Currently all three KDF factories -- Hkdf, CounterKdf, ConcatenationKdf return `SecretKeySpec` and I think maintaining that precedent is better and the caller already holds the metadata in the `PBEKeySpec` they pass in so I'd keep this consistent.

src/com/amazon/corretto/crypto/provider/Pbkdf2SecretKeyFactorySpi.java
prasden: Agreed, I instead moved the `getPassword()` call after the null checks so nothing is stored in `password` if the function returns early and zeroed the password in the `finally` block.

WillChilds-Klein (approved):

justsmth (approved):

PR #546, opened 2026-06-02
Reset RsaCipher buffer unconditionally on init and on doFinal failure

WillChilds-Klein:
## Resolves

V2220039808

## Summary

- `RsaCipher.engineInit()` only resets `buffer_` when the `Key` reference changes. With the same `Key` object, stale data from an aborted operation survives re-init and is prepended to the next operation's input -- causing cross-operation data bleed, including into `Cipher.wrap()`. This violates the JCA `Cipher.init()` contract (\"any previously-acquired state is lost\").
- `engineDoFinal()` only resets `buffer_` on the success path. A failed `doFinal` (e.g., `IllegalBlockSizeException`) leaves accumulated data in the buffer, corrupting subsequent operations on the same `Cipher` instance.
- Fix: (1) move `buffer_` allocation outside the `if (key_ != key)` block so it always resets on `init`, (2) wrap the entire `doFinal` body in `try/finally` so the buffer resets on any exit path. Use the capped two-arg `AccessibleByteArrayOutputStream` constructor at the reset site.
- Adds regression tests for both scenarios.

## Test plan

- [x] `./gradlew singleTest -DSINGLE_TEST=com.amazon.corretto.crypto.provider.test.RsaCipherTest` -- 3504 tests, 0 failed (33 aborted are pre-existing NoPadding skips)
- [x] `./gradlew spotlessJavaCheck` clean

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/RsaCipher.java
+++ b/src/com/amazon/corretto/crypto/provider/RsaCipher.java
@@ -212,7 +212,7 @@ class RsaCipher extends CipherSpi {
 
         return result;
       } finally {
-        buffer_ = new AccessibleByteArrayOutputStream(keySizeBytes_, keySizeBytes_);
+        buffer_.reset();
       }
     }
   }
--- a/src/com/amazon/corretto/crypto/provider/RsaCipher.java
+++ b/src/com/amazon/corretto/crypto/provider/RsaCipher.java
@@ -129,89 +129,91 @@ class RsaCipher extends CipherSpi {
     synchronized (lock_) {
       assertInitialized();
 
-      if (buffer_.size() != 0) { // Not one-shot
-        if (input != null) {
-          buffer_.write(input, inputOffset, inputLen);
-        }
-        input = buffer_.getDataBuffer();
-        inputOffset = 0;
-        inputLen = buffer_.size();
-      }
-      // One-shot, no input. Cipher only calls engineDoFinal with null input in doFinal overloads
-      // that don't take an input buffer, and in those cases, inputOffset and inputLen are 0.
-      // We set them here anyways to be safe, because the API makes no such guarantee.
-      else if (input == null) {
-        input = Utils.EMPTY_ARRAY;
-        inputOffset = 0;
-        inputLen = 0;
-      }
-
-      if (output.length - outputOffset < engineGetOutputSize(inputLen)) {
-        throw new ShortBufferException();
-      }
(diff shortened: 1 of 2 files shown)


src/com/amazon/corretto/crypto/provider/RsaCipher.java
```diff
@@ -129,89 +129,91 @@ protected int engineDoFinal(
...
+                        oaepLabel,
+                        finalInput,
+                        finalInputOffset,
+                        finalInputLen,
+                        output,
+                        outputOffset));
+
+        return result;
+      } finally {
+        buffer_ = new AccessibleByteArrayOutputStream(keySizeBytes_, keySizeBytes_);
```
justsmth: Shouldn't this be a buffer_.reset() to zeroize everything rather than relying on GC?

tst/com/amazon/corretto/crypto/provider/test/RsaCipherTest.java
```diff
@@ -1197,6 +1197,66 @@ private void wrapUnwrapRsa(final Cipher wrap, final Cipher unwrap)
...
+    // but fits in the buffer (256 bytes)
+    byte[] tooLarge = new byte[215];
+    Arrays.fill(tooLarge, (byte) 0xCC);
+    c.update(tooLarge);
+    try {
+      c.doFinal();
+      fail("Expected IllegalBlockSizeException");
+    } catch (IllegalBlockSizeException expected) {
+    }
+
```
justsmth: You could just `assertThrows` to be specific about which exception gets thrown:
```
assertThrows(IllegalBlockSizeException.class, () -> c.doFinal());
```

src/com/amazon/corretto/crypto/provider/RsaCipher.java
WillChilds-Klein: yes, `reset()` is the right call here, and we can assume `buffer_` is non-null because init state is asserted on L130.

L316 is fine as-is because it's in `engineInit` and JVM guarantees that the buffer's heap allocation returns a zeroed array

tst/com/amazon/corretto/crypto/provider/test/RsaCipherTest.java
WillChilds-Klein: yes, that's cleaner.

justsmth (approved):

geedo0 (approved):

PR #547, opened 2026-06-05
Zeroize native copies of sensitive JByteArrayCritical buffers

WillChilds-Klein:
# Description of changes

`GetPrimitiveArrayCritical` may return a pointer to the underlying Java
array (a pin) or a copy. `ReleasePrimitiveArrayCritical` does not zeroize
the freed native copy, so secret material can persist on the native heap
after the call returns. The Java-side `Arrays.fill()` only clears the
original Java array, not the native copy.

Add a required `wipe` parameter to `JByteArrayCritical`'s ctor. When
wipe is true and the JVM made a copy, the dtor uses `JNI_COMMIT` to copy
any modifications back to the Java array, `OPENSSL_cleanse`s the native
buffer, then `JNI_ABORT`s to free without re-copying the cleansed bytes.
For inputs (caller did not write through `get()`), the `JNI_COMMIT` is a
no-op for the Java array.

Motivated by [this comment](https://github.com/corretto/amazon-corretto-crypto-provider/pull/543#discussion_r3276611113).

# Callouts

Unfortunately, the new wiping code is [not covered by tests](https://app.codecov.io/gh/corretto/amazon-corretto-crypto-provider/pull/547?src=pr&el=tree&filepath=csrc%2Fbuffer.cpp&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=corretto#93b659ec92f8ad4a90715845bb1bf5e4-R63) because we don't have a good way to force the JVM to return a copy on `GetPrimitiveArrayCritical` rather than pinning.

---

### DIFF
--- a/csrc/aes_xts.cpp
+++ b/csrc/aes_xts.cpp
@@ -86,9 +86,9 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
     jint outputOffset)
 {
     try {
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey);
-        JByteArrayCritical input(env, jinput);
-        JByteArrayCritical output(env, joutput);
+        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, /*wipe=*/true);
+        JByteArrayCritical input(env, jinput, /*wipe=*/true);
+        JByteArrayCritical output(env, joutput, /*wipe=*/false);
 
         AesXtsCipher cipher(true, packedTweakKey.get() + AES_XTS_KEY_INDEX_START, packedTweakKey.get());
         cipher.encrypt(input.get() + inputOffset, inputLen, output.get() + outputOffset);
@@ -107,8 +107,10 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
     jint outputOffset)
 {
     try {
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey);
-        JByteArrayCritical input(env, jinput);
+        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, /*wipe=*/true);
+        // Buffer holds plaintext on entry and ciphertext on exit; the native copy holds
+        // plaintext mid-call, so it must be cleansed before being freed.
+        JByteArrayCritical input(env, jinput, /*wipe=*/true);
 
         AesXtsCipher cipher(true, packedTweakKey.get() + AES_XTS_KEY_INDEX_START, packedTweakKey.get());
         cipher.encrypt(input.get() + inputOffset, inputLen, input.get() + outputOffset);
@@ -128,9 +130,9 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
     jint outputOffset)
 {
     try {
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey);
-        JByteArrayCritical input(env, jinput);
-        JByteArrayCritical output(env, joutput);
+        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, /*wipe=*/true);
+        JByteArrayCritical input(env, jinput, /*wipe=*/false);
+        JByteArrayCritical output(env, joutput, /*wipe=*/true);
 
         AesXtsCipher cipher(false, packedTweakKey.get() + AES_XTS_KEY_INDEX_START, packedTweakKey.get());
         cipher.decrypt(input.get() + inputOffset, inputLen, output.get() + outputOffset);
@@ -151,8 +153,9 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
     jint outputOffset)
 {
     try {
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey);
-        JByteArrayCritical input(env, jinput);
+        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, /*wipe=*/true);
+        // Buffer holds ciphertext on entry and plaintext on exit; cleanse the native copy.
+        JByteArrayCritical input(env, jinput, /*wipe=*/true);
 
         AesXtsCipher cipher(false, packedTweakKey.get() + AES_XTS_KEY_INDEX_START, packedTweakKey.get());
         cipher.decrypt(input.get() + inputOffset, inputLen, input.get() + outputOffset);
--- a/csrc/aes_xts.cpp
+++ b/csrc/aes_xts.cpp
@@ -134,6 +134,7 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
     jint outputOffset)
 {
     try {
+        // |output| (WIPE_OUTPUT) declared first so its dtor runs last, after the other
         JByteArrayCritical output(env, joutput, WipeMode::WIPE_OUTPUT);
         JByteArrayCritical packedTweakKey(env, jPackedTweakKey, WipeMode::WIPE_INPUT);
         JByteArrayCritical input(env, jinput, WipeMode::NO_WIPE);
--- a/csrc/aes_xts.cpp
+++ b/csrc/aes_xts.cpp
@@ -149,6 +149,7 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
         const jsize jOutputArrLen = env->GetArrayLength(joutput);
 
         // |output| (WIPE_OUTPUT) declared first so its dtor runs last, after the other
+        // buffers' critical regions have been released; see csrc/buffer.h.
         JByteArrayCritical output(env, joutput, jOutputArrLen, WipeMode::WIPE_OUTPUT);
         JByteArrayCritical packedTweakKey(env, jPackedTweakKey, jPackedTweakKeyLen, WipeMode::WIPE_INPUT);
         JByteArrayCritical input(env, jinput, jInputArrLen, WipeMode::NO_WIPE);
--- a/csrc/aes_xts.cpp
+++ b/csrc/aes_xts.cpp
@@ -86,9 +86,11 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
     jint outputOffset)
 {
     try {
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, /*wipe=*/true);
-        JByteArrayCritical input(env, jinput, /*wipe=*/true);
-        JByteArrayCritical output(env, joutput, /*wipe=*/false);
+        // No WIPE_OUTPUT here: ciphertext is not sensitive, plaintext input is read-only.
+        // Plain RAII suffices; no commitBack() needed.
+        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, WipeMode::WIPE_INPUT);
+        JByteArrayCritical input(env, jinput, WipeMode::WIPE_INPUT);
+        JByteArrayCritical output(env, joutput, WipeMode::NO_WIPE);
 
         AesXtsCipher cipher(true, packedTweakKey.get() + AES_XTS_KEY_INDEX_START, packedTweakKey.get());
         cipher.encrypt(input.get() + inputOffset, inputLen, output.get() + outputOffset);
@@ -107,10 +109,12 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
     jint outputOffset)
 {
     try {
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, /*wipe=*/true);
-        // Buffer holds plaintext on entry and ciphertext on exit; the native copy holds
-        // plaintext mid-call, so it must be cleansed before being freed.
-        JByteArrayCritical input(env, jinput, /*wipe=*/true);
+        // Same buffer holds plaintext on entry and ciphertext on exit; native copy held
+        // plaintext mid-call, so we WIPE_OUTPUT (cleanse + commit ciphertext back).
+        // Declared first so its dtor runs last, after packedTweakKey's critical region
+        // has been released.
+        JByteArrayCritical input(env, jinput, WipeMode::WIPE_OUTPUT);
+        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, WipeMode::WIPE_INPUT);
 
         AesXtsCipher cipher(true, packedTweakKey.get() + AES_XTS_KEY_INDEX_START, packedTweakKey.get());
         cipher.encrypt(input.get() + inputOffset, inputLen, input.get() + outputOffset);
@@ -130,9 +134,9 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
     jint outputOffset)
 {
     try {
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, /*wipe=*/true);
-        JByteArrayCritical input(env, jinput, /*wipe=*/false);
-        JByteArrayCritical output(env, joutput, /*wipe=*/true);
+        JByteArrayCritical output(env, joutput, WipeMode::WIPE_OUTPUT);
+        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, WipeMode::WIPE_INPUT);
+        JByteArrayCritical input(env, jinput, WipeMode::NO_WIPE);
 
         AesXtsCipher cipher(false, packedTweakKey.get() + AES_XTS_KEY_INDEX_START, packedTweakKey.get());
         cipher.decrypt(input.get() + inputOffset, inputLen, output.get() + outputOffset);
@@ -153,9 +157,10 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
     jint outputOffset)
 {
     try {
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, /*wipe=*/true);
-        // Buffer holds ciphertext on entry and plaintext on exit; cleanse the native copy.
-        JByteArrayCritical input(env, jinput, /*wipe=*/true);
+        // Same buffer holds ciphertext on entry and plaintext on exit; WIPE_OUTPUT cleanses
+        // any native copy and commits the plaintext back.
+        JByteArrayCritical input(env, jinput, WipeMode::WIPE_OUTPUT);
+        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, WipeMode::WIPE_INPUT);
 
         AesXtsCipher cipher(false, packedTweakKey.get() + AES_XTS_KEY_INDEX_START, packedTweakKey.get());
         cipher.decrypt(input.get() + inputOffset, inputLen, input.get() + outputOffset);
--- a/csrc/aes_xts.cpp
+++ b/csrc/aes_xts.cpp
@@ -86,11 +86,16 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
     jint outputOffset)
 {
     try {
+        // GetArrayLength must be called BEFORE any JBAC ctor; see csrc/buffer.h.
+        const jsize jPackedTweakKeyLen = env->GetArrayLength(jPackedTweakKey);
+        const jsize jInputArrLen = env->GetArrayLength(jinput);
+        const jsize jOutputArrLen = env->GetArrayLength(joutput);
+
         // No WIPE_OUTPUT here: ciphertext is not sensitive, plaintext input is read-only.
         // Plain RAII suffices; no commitBack() needed.
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, WipeMode::WIPE_INPUT);
-        JByteArrayCritical input(env, jinput, WipeMode::WIPE_INPUT);
-        JByteArrayCritical output(env, joutput, WipeMode::NO_WIPE);
+        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, jPackedTweakKeyLen, WipeMode::WIPE_INPUT);
+        JByteArrayCritical input(env, jinput, jInputArrLen, WipeMode::WIPE_INPUT);
+        JByteArrayCritical output(env, joutput, jOutputArrLen, WipeMode::NO_WIPE);
 
         AesXtsCipher cipher(true, packedTweakKey.get() + AES_XTS_KEY_INDEX_START, packedTweakKey.get());
         cipher.encrypt(input.get() + inputOffset, inputLen, output.get() + outputOffset);
@@ -109,12 +114,16 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
     jint outputOffset)
 {
     try {
+        // GetArrayLength must be called BEFORE any JBAC ctor; see csrc/buffer.h.
+        const jsize jPackedTweakKeyLen = env->GetArrayLength(jPackedTweakKey);
+        const jsize jInputArrLen = env->GetArrayLength(jinput);
+
         // Same buffer holds plaintext on entry and ciphertext on exit; native copy held
         // plaintext mid-call, so we WIPE_OUTPUT (cleanse + commit ciphertext back).
         // Declared first so its dtor runs last, after packedTweakKey's critical region
         // has been released.
-        JByteArrayCritical input(env, jinput, WipeMode::WIPE_OUTPUT);
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, WipeMode::WIPE_INPUT);
+        JByteArrayCritical input(env, jinput, jInputArrLen, WipeMode::WIPE_OUTPUT);
+        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, jPackedTweakKeyLen, WipeMode::WIPE_INPUT);
 
         AesXtsCipher cipher(true, packedTweakKey.get() + AES_XTS_KEY_INDEX_START, packedTweakKey.get());
         cipher.encrypt(input.get() + inputOffset, inputLen, input.get() + outputOffset);
@@ -134,10 +143,15 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
     jint outputOffset)
 {
     try {
+        // GetArrayLength must be called BEFORE any JBAC ctor; see csrc/buffer.h.
+        const jsize jPackedTweakKeyLen = env->GetArrayLength(jPackedTweakKey);
+        const jsize jInputArrLen = env->GetArrayLength(jinput);
+        const jsize jOutputArrLen = env->GetArrayLength(joutput);
+
         // |output| (WIPE_OUTPUT) declared first so its dtor runs last, after the other
-        JByteArrayCritical output(env, joutput, WipeMode::WIPE_OUTPUT);
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, WipeMode::WIPE_INPUT);
-        JByteArrayCritical input(env, jinput, WipeMode::NO_WIPE);
+        JByteArrayCritical output(env, joutput, jOutputArrLen, WipeMode::WIPE_OUTPUT);
+        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, jPackedTweakKeyLen, WipeMode::WIPE_INPUT);
+        JByteArrayCritical input(env, jinput, jInputArrLen, WipeMode::NO_WIPE);
 
         AesXtsCipher cipher(false, packedTweakKey.get() + AES_XTS_KEY_INDEX_START, packedTweakKey.get());
         cipher.decrypt(input.get() + inputOffset, inputLen, output.get() + outputOffset);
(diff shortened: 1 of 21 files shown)


csrc/buffer.cpp
```diff
@@ -37,17 +37,36 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
+        // JNI_COMMIT copies any modifications back to the Java array but keeps the
+        // native buffer alive, so we can OPENSSL_cleanse it before the JNI_ABORT
+        // release that frees it without re-copying the cleansed bytes back. Inputs
+        // were never written by C code, so the commit is a no-op for the Java array.
+        env_->ReleasePrimitiveArrayCritical(jarray_, ptr_, JNI_COMMIT);
+        if (len_ > 0) {
+            OPENSSL_cleanse(ptr_, len_);
+        }
+        env_->ReleasePrimitiveArrayCritical(jarray_, ptr_, JNI_ABORT);
+    } else {
```
justsmth: Is this safe across all JVMs?

skmcgrail (dismissed):

csrc/buffer.cpp
```diff
@@ -37,17 +37,36 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
-JByteArrayCritical::JByteArrayCritical(JNIEnv* env, jbyteArray jarray)
+JByteArrayCritical::JByteArrayCritical(JNIEnv* env, jbyteArray jarray, bool wipe)
     : env_(env)
     , jarray_(jarray)
+    , len_(0)
+    , is_copy_(JNI_FALSE)
+    , wipe_(wipe)
 {
-    ptr_ = env->GetPrimitiveArrayCritical(jarray, nullptr);
+    len_ = env->GetArrayLength(jarray);
```
justsmth: Only needed for wipe?
```
    if (wipe_) {
        len_ = env->GetArrayLength(jarray);
    }
```

justsmth (dismissed):

csrc/buffer.cpp
WillChilds-Klein: technically, today, yes. but i'd prefer to populate it unconditionally for consistency.

csrc/buffer.cpp
WillChilds-Klein: good catch. claude says no, that it's safe to double-free the same `ptr_` on OpenJDK/corretto is an undocumented implementation detail, not a guaranteed interface. i'll modify this case to do an extra allocation/copy before freeing. not cheap perf-wise, but not the common case so acceptable IMO for increased soundness.

csrc/buffer.cpp
```diff
@@ -37,17 +37,46 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
     return array;
 }
 
-JByteArrayCritical::JByteArrayCritical(JNIEnv* env, jbyteArray jarray)
+JByteArrayCritical::JByteArrayCritical(JNIEnv* env, jbyteArray jarray, bool wipe)
```
prasden: should we add a default for `wipe = true`? right now the it is the caller's responsibility to set wipe to true.

csrc/buffer.cpp
```diff
@@ -37,17 +37,46 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
     : env_(env)
     , jarray_(jarray)
+    , len_(0)
+    , is_copy_(JNI_FALSE)
+    , wipe_(wipe)
 {
-    ptr_ = env->GetPrimitiveArrayCritical(jarray, nullptr);
+    len_ = env->GetArrayLength(jarray);
```
prasden: do we need to initialize `len` to 0 above if we override here?

csrc/buffer.cpp
```diff
@@ -37,17 +37,46 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
+        }
+        env_->ReleasePrimitiveArrayCritical(jarray_, ptr_, JNI_ABORT);
+        // Now outside the critical region; safe to call other JNI functions.
+        if (len_ > 0) {
+            env_->SetByteArrayRegion(jarray_, 0, len_, reinterpret_cast<const jbyte*>(tmp.data()));
+        }
+    } else {
+        env_->ReleasePrimitiveArrayCritical(jarray_, ptr_, 0);
+    }
+}
```
prasden: We are outside the critical region for this object, but I think not for the entire thread -- [here](https://github.com/corretto/amazon-corretto-crypto-provider/pull/547/changes#diff-14709542504a4b90c97aef1bf37ccc50e63856a416df0152a232d4ac8b0a9c17R24-R27) you initialize four `JByteArrayCritical` objects and when the destructors run in reverse order, once you get to the destructor of `secret`, `SetByteArrayRegion` is called while `output` is still being held.

The [docs](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#GetPrimitiveArrayCritical_ReleasePrimitiveArrayCritical) for `Get/ReleasePrimitiveArrayCritical` say

 > After calling GetPrimitiveArrayCritical, the native code should not run for an extended period of time before it calls ReleasePrimitiveArrayCritical. We must treat the code inside this pair of functions as running in a "critical region."... native code must not call other JNI functions, or any system call that may cause the current thread to block and wait for another Java thread.

and we call `SetByteArrayRegion` before the last object is destructed and released, which is problematic, but only in if `is_copy` is true.  Do we need to address this or will this path never be hit? Do correct me if I've understood this wrong.

csrc/buffer.cpp
justsmth: The `/*wipe=*/` at every call site does seem like a "code smell".

Rather than a default, which I'd avoid since it's security-relevant, I think it should force callers to state their intent. How about a scoped enum?

    enum class WipeMode { MUST_WIPE, NO_WIPE };
    // JByteArrayCritical output(env, jOutput, WipeMode::MUST_WIPE);

csrc/buffer.cpp
justsmth: Agree -- this can be hit. And it can't be fixed by reordering the locals either. (In `pbkdf2.cpp:23-25` and `aes_xts.cpp:133-135` the *output* buffer is declared last, so it destructs *first* and runs `SetByteArrayRegion` while the input criticals are still held.)

The commit needs to move out of the critical region entirely rather than relying on a favorable destruction order -- I think this refactor could resolve most of the remaining concerns.

csrc/buffer.cpp
```diff
@@ -37,17 +37,46 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
+        // The JVM returned a separate native copy. Stash the (possibly modified) bytes in
+        // a SecureAlloc-backed vector, OPENSSL_cleanse the native buffer, end the critical
+        // region with JNI_ABORT (so the cleansed bytes don't get copied back), and only
+        // then call SetByteArrayRegion to commit the saved bytes to the Java array. The
+        // intermediate vector is itself cleansed when it goes out of scope (SecureAlloc).
+        //
+        // This avoids relying on calling ReleasePrimitiveArrayCritical twice (once with
+        // JNI_COMMIT and once with JNI_ABORT) on the same pointer, which is not portably
+        // documented across JVM implementations.
+        std::vector<uint8_t, SecureAlloc<uint8_t> > tmp(static_cast<uint8_t*>(ptr_),
```
justsmth: This allocation can throw `std::bad_alloc` (`SecureAlloc::allocate` in `env.h:281,287`, plus `operator new`). Destructors are implicitly `noexcept` in C++11, so a throw here aborts the process -- strictly worse than the pre-PR behavior.

Maybe we can allocate/size the stash buffer up front (before the cleanse/release sequence) so the dtor body is noexcept-safe?

csrc/buffer.cpp
```diff
@@ -37,17 +37,46 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
+                                                        static_cast<uint8_t*>(ptr_) + len_);
+        if (len_ > 0) {
+            OPENSSL_cleanse(ptr_, len_);
+        }
+        env_->ReleasePrimitiveArrayCritical(jarray_, ptr_, JNI_ABORT);
+        // Now outside the critical region; safe to call other JNI functions.
+        if (len_ > 0) {
+            env_->SetByteArrayRegion(jarray_, 0, len_, reinterpret_cast<const jbyte*>(tmp.data()));
+        }
```
justsmth: For pure inputs (`secret`/`prk`/`password`), we never write through `get()`, so this `SetByteArrayRegion` just rewrites identical bytes.

Since inputs would only need cleanse + `JNI_ABORT`, could we distinguish input vs output during construction (possibly extending the `enum WipeMode` suggestions from the other thread), then this skips the redundant work *and* shrink the surface of the critical-region fix above?

csrc/buffer.cpp
```diff
@@ -37,17 +37,46 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
+    ptr_ = env->GetPrimitiveArrayCritical(jarray, &is_copy_);
     if (ptr_ == nullptr) {
         throw java_ex(EX_ERROR, "GetPrimitiveArrayCritical failed.");
     }
 }
 
-JByteArrayCritical::~JByteArrayCritical() { env_->ReleasePrimitiveArrayCritical(jarray_, ptr_, 0); }
+JByteArrayCritical::~JByteArrayCritical()
+{
+    if (wipe_ && is_copy_) {
```
justsmth: This `wipe_ && is_copy_` branch is the one the PR notes is not covered by tests. 

Could we pull the stash/cleanse/commit logic into a free function that takes a buffer + length so we can unit-test it directly?

justsmth (changes requested):

csrc/buffer.cpp
WillChilds-Klein: I intentionally did not include a default here, forcing the caller to consider whether their array contains sensitive material. enum instead of bool sounds fine to me, will implement.

csrc/buffer.cpp
WillChilds-Klein: doesn't hurt. `len_` assignment may change in the future, so better to initialize.

csrc/buffer.cpp
WillChilds-Klein: yes, we should unit test this.

csrc/buffer.cpp
WillChilds-Klein: good point. i'm a bit surprised the compiler didn't catch this... i'll think about how we can avoid the allocation here.

csrc/buffer.cpp
WillChilds-Klein: I added distinct WIPE_INPUT and WIPE_OUTPUT modes

csrc/buffer.cpp
WillChilds-Klein: Yes, that's true. I don't love the solution, but I've separated out `release` and `commitBack`. these get called automatically and idempotently in the dtor, BUT callers will need to invoke them manually in the case of ≥2 `WIPE_OUTPUT` buffers in the same scope. This is less than ideal as it requires the caller to 1/ call these methods manually and 2/ call them in the correct order. The best mechanism I can think of to "enforce" this is thorough documentation. Any other ideas?

PS -- at first I consolidated the `release` and `commitBack` calls into the dtor w/ no public exposure. that works OK for one WIPE_OUTPUT buffer in a given scope, but it makes having 2 impossible because the first buffer's `commitBack`/`SetByteArrayRegion` is called while the second buffer is still in its critical region. this solution is a bit more verbose, but flexible, and can still rely on the dtor in the case of 1 `WIPE_OUTPUT` buffer in scope.

justsmth: Behavior looks correct for all current call sites. My only worry is that a future second `WIPE_OUTPUT` (or reordered decl) silently reintroduces the violation.

A debug thread-local count of open criticals would guard against this, but it has to sit on a path that runs every time -- not behind the copy-only `stash_.empty()` gate, or it almost never executes:

```cpp
// debug-only; compiled out under NDEBUG
static thread_local int s_open_criticals;
// ctor (after GetPrimitiveArrayCritical succeeds): ++s_open_criticals;
// release() (once, guarded by released_):          --s_open_criticals;

void commitBack() {
    if (mode_ != WipeMode::WIPE_OUTPUT) return;
    assert(s_open_criticals == 0 && "commitBack() while a critical region is still open");
    if (stash_.empty()) return;
    env_->SetByteArrayRegion(...);
    stash_.clear();
}
```
The count is independent of pin-vs-copy, so the buggy multi-output case trips it on a normal pinned run. 

Alternatively a small RAII scope type that owns the buffers and orders release/commit would remove the convention entirely, but the assert is probably enough for now.

csrc/buffer.cpp
```diff
@@ -37,19 +37,106 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
+    // bad_alloc, release the critical region ourselves, and rethrow to preserve the
+    // original failure mode.
+    if (mode_ == WipeMode::WIPE_OUTPUT && is_copy_ && len_ > 0) {
+        try {
+            stash_.reserve(static_cast<size_t>(len_));
+        } catch (...) {
+            env_->ReleasePrimitiveArrayCritical(jarray_, ptr_, JNI_ABORT);
+            ptr_ = nullptr;
+            throw;
+        }
```
justsmth, line 107: The `reserve()` failure is rethrown as a raw `std::bad_alloc`, but the call sites only `catch (java_ex&)`. Converting to a `java_ex` here surfaces a clean Java `OutOfMemoryError` and matches how the rest of the ctor reports failure.

```suggestion
        } catch (...) {
            env_->ReleasePrimitiveArrayCritical(jarray_, ptr_, JNI_ABORT);
            ptr_ = nullptr;
            throw java_ex(EX_OOM, "Failed to allocate stash buffer for sensitive output array");
        }
```

csrc/aes_xts.cpp
```diff
@@ -128,9 +134,9 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
     jint outputOffset)
 {
     try {
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey);
-        JByteArrayCritical input(env, jinput);
-        JByteArrayCritical output(env, joutput);
+        JByteArrayCritical output(env, joutput, WipeMode::WIPE_OUTPUT);
```
justsmth: Nit: `hkdf`, `pbkdf2`, and the `*SameBuffer` variants all explain why `WIPE_OUTPUT` is declared first; `dec()` is the only site without it. Since the ordering is important, it's worth documenting:

```suggestion
        // |output| (WIPE_OUTPUT) declared first so its dtor runs last, after the other
        JByteArrayCritical output(env, joutput, WipeMode::WIPE_OUTPUT);
```

csrc/buffer.cpp
WillChilds-Klein: Ah, the thread-wide counter is a good idea. i'll implement that.

csrc/buffer.cpp
```diff
@@ -37,19 +49,127 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
+    // class comment in buffer.h for the call-site pattern (single WIPE_OUTPUT per scope
+    // declared first; multi-WIPE_OUTPUT requires manual release()/commitBack() ordering).
+    //
+    // Not directly unit-tested: assert() calls abort() on failure, which is incompatible
+    // with the in-process test harness, and constructing a JByteArrayCritical needs a real
+    // JNIEnv. Code review of new call sites is the primary defense; this assertion is the
+    // backstop that fires in any debug-mode test run that misuses the API.
+#ifndef NDEBUG
+    assert(s_open_criticals == 0 && "commitBack() called while a critical region is still open on this thread");
+#endif
```
justsmth: The counter plumbing looks right, but the assert ended up after the `stash_.empty()` gate, so it only runs on the copy path (`is_copy_ == true`) — on a pinned HotSpot run `commitBack()` returns before reaching it, which is the case we were trying to cover. Splitting the mode check from the stash check fixes it:

```suggestion
    if (mode_ != WipeMode::WIPE_OUTPUT) {
        return;
    }
    // Tripping this assertion means another JByteArrayCritical's critical region is
    // still open on this thread. SetByteArrayRegion is forbidden in that state. See the
    // class comment in buffer.h for the call-site pattern (single WIPE_OUTPUT per scope
    // declared first; multi-WIPE_OUTPUT requires manual release()/commitBack() ordering).
    //
    // Not directly unit-tested: assert() calls abort() on failure, which is incompatible
    // with the in-process test harness, and constructing a JByteArrayCritical needs a real
    // JNIEnv. Code review of new call sites is the primary defense; this assertion is the
    // backstop that fires in any debug-mode test run that misuses the API.
#ifndef NDEBUG
    assert(s_open_criticals == 0 && "commitBack() called while a critical region is still open on this thread");
#endif
    if (stash_.empty()) {
        return;
    }
```

WillChilds-Klein: Hm, well it gets incremented unconditionally in the ctor, so should we just do the assert at the top of the function before `mode_` check as well? seems like we should enforce alloc/release hygiene regardless of mode.

csrc/buffer.cpp
```diff
@@ -2,10 +2,22 @@
...
 namespace AmazonCorrettoCryptoProvider {
 
+#ifndef NDEBUG
+// Per-thread count of currently-open JByteArrayCritical regions. Incremented after
+// successful GetPrimitiveArrayCritical and decremented in release(). Used by
+// commitBack() to assert that no other critical regions are still open on this thread
+// when SetByteArrayRegion runs -- which would silently violate JNI's "no other JNI
+// calls within a critical region" rule. The count is independent of pin-vs-copy, so a
+// buggy multi-WIPE_OUTPUT scope trips the assertion on a normal pinned run, not just
+// on the rare JVM-copy path. Compiled out under NDEBUG.
```
prasden: What is the rare JVM copy path?

csrc/buffer.cpp
justsmth: I don't think we can put it above the `mode_` check -- it'd false-positive. 

`NO_WIPE`/`WIPE_INPUT` buffers run `commitBack()` during their destructor while later-declared buffers are still in their critical regions, so a non-zero count is legal for them -- they never call `SetByteArrayRegion`, so the `== 0` invariant only applies to `WIPE_OUTPUT`.

So the mode gate should be first, then the `assert`, and then `stash_.empty()` check after that. You could also have `assert(s_open_criticals > 0)` in `release()` before the decrement, but that's a separate check.

csrc/buffer.cpp
```diff
@@ -37,19 +49,127 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
 
-JByteArrayCritical::~JByteArrayCritical() { env_->ReleasePrimitiveArrayCritical(jarray_, ptr_, 0); }
+JByteArrayCritical::~JByteArrayCritical()
+{
+    // RAII fallback for the single-WIPE_OUTPUT-per-scope case: end the critical region
+    // and commit any stashed writes. Multi-WIPE_OUTPUT scopes must drive these manually
+    // (see class comment); calls here will be no-ops if the caller already did.
+    if (!released_) {
+        release();
+    }
```
prasden: I don't think we need this guard. You already check in the `release()` function if that object's release is set to true and return early if it is.

csrc/buffer.cpp
WillChilds-Klein: the case where the JVM isn't able to return pinned memory, so it returns a copy in GetPrimitiveArrayCritical

>  If possible, the VM returns a pointer to the primitive array; otherwise, a copy is made.

https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#GetPrimitiveArrayCritical_ReleasePrimitiveArrayCritical

csrc/buffer.cpp
```diff
@@ -37,19 +49,127 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
+    if (released_) {
+        return;
+    }
+    released_ = true;
+#ifndef NDEBUG
+    --s_open_criticals;
+#endif
+
+    // Pinned or non-sensitive: ptr_ aliases the Java array (or is opaque to us). Mode-0
+    // release commits any writes that already landed in the array; for the pinned WIPE_*
```
prasden: maybe it's just me, but I had to research what mode-0 was. I think a little description of the 0 parameter could help

csrc/aes_xts.cpp
```diff
@@ -86,9 +86,11 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
...
 {
     try {
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey);
-        JByteArrayCritical input(env, jinput);
-        JByteArrayCritical output(env, joutput);
+        // No WIPE_OUTPUT here: ciphertext is not sensitive, plaintext input is read-only.
+        // Plain RAII suffices; no commitBack() needed.
+        JByteArrayCritical packedTweakKey(env, jPackedTweakKey, WipeMode::WIPE_INPUT);
+        JByteArrayCritical input(env, jinput, WipeMode::WIPE_INPUT);
+        JByteArrayCritical output(env, joutput, WipeMode::NO_WIPE);
```
prasden: Not sure if there is a resolution to this, but if a caller accidentally tags a secret with `NO_WIPE` that's a problem. Thorough documentation, as you've done is a good way to prevent this. Asked around with Claude and one option was creating a `SecretInput/OutputArray` explicit type which carries the cleanse and stash behavior itself, or setting one of the `WIPE_X` as the default in the constructor so caller doesn't have to worry about it everytime, but it does increase complexity if a ciphertext is defaulted to wipe when it doesn't need to be. 

`SecretInput/OutputArray` could be two subclasses of `JByteArrayCritical` and their respective destructors could cleanse/release_and _commit_back, which might be cleaner than using the enum to determine and having asserts to check if we are doing the right operation.

csrc/buffer.h
```diff
@@ -557,14 +557,92 @@ template <typename T> class bounce_buffer {
...
+//   WIPE_OUTPUT: Sensitive buffer the caller writes through (derived key, plaintext-out). On the
+//                copy path, the bytes the caller wrote are stashed in a SecureAlloc-backed
+//                vector, the native buffer is OPENSSL_cleansed and discarded via JNI_ABORT, and
+//                the stashed bytes are committed back to the Java array via SetByteArrayRegion
+//                AFTER all critical regions on the thread have been released.
+enum class WipeMode {
+    NO_WIPE,
+    WIPE_INPUT,
+    WIPE_OUTPUT,
+};
```
prasden, line 577: opinionated, but maybe `NO_CLEANSE`, `CLEANSE_SECRET`, `CLEANSE_AND_RETURN_SECRET`, something along those lines? Gives a little bit more distinction on why we need to wipe (secrets), as well as input/output actions

csrc/buffer.cpp
prasden: maybe we can add that here?

prasden: also commitBack() only does work on the copy path. that's the only path that fills `stash_`. On a pin, `stash_` is empty  as we release it immediately and it's a no-op. I think we can make that clearer here

csrc/buffer.cpp
```diff
@@ -37,19 +49,127 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
     , jarray_(jarray)
+    , len_(0)
+    , is_copy_(JNI_FALSE)
+    , mode_(mode)
+    , released_(false)
 {
-    ptr_ = env->GetPrimitiveArrayCritical(jarray, nullptr);
+    len_ = env->GetArrayLength(jarray);
+
+    ptr_ = env->GetPrimitiveArrayCritical(jarray, &is_copy_);
```
prasden: When the first `JByteArrayCritical` object is created, we access the critical region with `GetPrimitiveArrayCritical.` For the next object created in this thread, we violate the contract by calling another JNI method `GetArrayLength` here. I think a fix would be to get the lengths of the buffer first, and then create the objects at each call site, or pass in at the Java layer, which adds more to this PR.

csrc/buffer.cpp
WillChilds-Klein: Sure, i can update the comment.

csrc/buffer.cpp
WillChilds-Klein: Sure, i can update the comment.

csrc/buffer.cpp
WillChilds-Klein: That's true, and idempotency is in the function's contract, so i agree it's safe to drop this conditional.

csrc/aes_xts.cpp
WillChilds-Klein: That's a good idea. How about I defer that refactor to a follow-up so we can see what it looks like compared to today's enum-based approach?

csrc/buffer.cpp
WillChilds-Klein: Hm, yeah, looks like we do need to pass in the length here to avoid interleaving JNI calls. In the follow-up's refactor, we can restrict that length parameter to just the WIPE_OUTPUT analog, as we shouldn't need array length for the other variants.

csrc/buffer.h
WillChilds-Klein: I'm agnostic on this, but if we merge the follow-up refactor these will go away anyway.

csrc/buffer.cpp
WillChilds-Klein: separated out mode/stash emptiness check and added an additional assert in `release()`

csrc/buffer.h
```diff
@@ -557,14 +557,97 @@ template <typename T> class bounce_buffer {
...
+//        //   3. output -> ends output's critical region, then SetByteArrayRegion
+//        //                (safe: no other criticals are open at this point)
+//
+// 2. Multiple WIPE_OUTPUTs in one scope: the caller MUST drive release() and
+//    commitBack() manually so that ALL criticals are closed before ANY commit runs:
+//
+//        JByteArrayCritical out1(env, jA, WipeMode::WIPE_OUTPUT);
+//        JByteArrayCritical out2(env, jB, WipeMode::WIPE_OUTPUT);
+//        JByteArrayCritical password(env, jPassword, WipeMode::WIPE_INPUT);
+//        // ... algorithm writes through out1, out2 ...
```
justsmth: Nit: stale examples.
```suggestion
//        const jsize aLen        = env->GetArrayLength(jA);
//        const jsize bLen        = env->GetArrayLength(jB);
//        const jsize passwordLen = env->GetArrayLength(jPassword);
//        JByteArrayCritical out1(env, jA, aLen, WipeMode::WIPE_OUTPUT);
//        JByteArrayCritical out2(env, jB, bLen, WipeMode::WIPE_OUTPUT);
//        JByteArrayCritical password(env, jPassword, passwordLen, WipeMode::WIPE_INPUT);
//        // ... algorithm writes through out1, out2 ...
```

csrc/buffer.h
```diff
@@ -557,14 +557,97 @@ template <typename T> class bounce_buffer {
...
+// The class supports two usage patterns:
+//
+// 1. Single WIPE_OUTPUT per scope (the common case): rely on RAII. Declare the
+//    WIPE_OUTPUT first so it destructs LAST under C++'s LIFO order, after every other
+//    JByteArrayCritical's critical region has ended. The dtor handles everything.
+//
+//        JByteArrayCritical output(env, jOutput, WipeMode::WIPE_OUTPUT);  // declared FIRST
+//        JByteArrayCritical secret(env, jSecret, WipeMode::WIPE_INPUT);
+//        JByteArrayCritical salt(env, jSalt, WipeMode::NO_WIPE);
+//        // ... algorithm ...
```
justsmth: Nit: stale examples:
```suggestion
//        const jsize outLen    = env->GetArrayLength(jOutput);
//        const jsize secretLen = env->GetArrayLength(jSecret);
//        const jsize saltLen   = env->GetArrayLength(jSalt);
//        // All GetArrayLength calls happen BEFORE the first ctor (see ctor docs below).
//        JByteArrayCritical output(env, jOutput, outLen, WipeMode::WIPE_OUTPUT);  // declared FIRST
//        JByteArrayCritical secret(env, jSecret, secretLen, WipeMode::WIPE_INPUT);
//        JByteArrayCritical salt(env, jSalt, saltLen, WipeMode::NO_WIPE);
//        // ... algorithm ...
```

csrc/aes_xts.cpp
```diff
@@ -128,9 +143,15 @@ extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_AesXt
...
     try {
-        JByteArrayCritical packedTweakKey(env, jPackedTweakKey);
-        JByteArrayCritical input(env, jinput);
-        JByteArrayCritical output(env, joutput);
+        // GetArrayLength must be called BEFORE any JBAC ctor; see csrc/buffer.h.
+        const jsize jPackedTweakKeyLen = env->GetArrayLength(jPackedTweakKey);
+        const jsize jInputArrLen = env->GetArrayLength(jinput);
+        const jsize jOutputArrLen = env->GetArrayLength(joutput);
+
+        // |output| (WIPE_OUTPUT) declared first so its dtor runs last, after the other
```
justsmth, line 151: Nit: sentence seems cutoff (sorry, I think this was my suggestion.)
```suggestion
        // |output| (WIPE_OUTPUT) declared first so its dtor runs last, after the other
        // buffers' critical regions have been released; see csrc/buffer.h.
```

csrc/buffer.cpp
```diff
@@ -37,19 +60,134 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
+}
+
+void JByteArrayCritical::release() noexcept
+{
+    if (released_) {
+        return;
+    }
+    released_ = true;
+#ifndef NDEBUG
+    assert(s_open_criticals > 0 && "release() called with negative open criticals, should never happen");
```
justsmth: Nit: inaccurate assert message:
```suggestion
    assert(s_open_criticals > 0 && "release() would underflow s_open_criticals; unbalanced ctor/release()");
```

csrc/buffer.cpp
```diff
@@ -37,19 +60,134 @@ jbyteArray vecToArray(raii_env& env, const std::vector<uint8_t, SecureAlloc<uint
...
+    //
+    // Not directly unit-tested: assert() calls abort() on failure, which is incompatible
+    // with the in-process test harness, and constructing a JByteArrayCritical needs a real
+    // JNIEnv. Code review of new call sites is the primary defense; this assertion is the
+    // backstop that fires in any debug-mode test run that misuses the API.
+#ifndef NDEBUG
+    assert(s_open_criticals == 0 && "commitBack() called while a critical region is still open on this thread");
+#endif
+   if (stash_.empty()) {
+        return;
```
justsmth: Nit: indentation
```suggestion
   if (stash_.empty()) {
      return;
```

justsmth (dismissed):

csrc/aes_xts.cpp
prasden: Yep that's great

prasden (approved):

justsmth (approved):

PR #548, opened 2026-06-11
Implement DestroyableSecretKey and use it for ML-KEM

WillChilds-Klein:
*Issue #, if available:* V2246228895

## Summary

Add a package-private `DestroyableSecretKey` class — a `SecretKey` impl backed by a defensively-copied `byte[]` whose `destroy()` overwrites the bytes with zeros and marks the instance unusable. Post-destroy `getEncoded()` and `getAlgorithm()` throw `IllegalStateException`; double-destroy throws; serialization is blocked (since `SecretKey extends Serializable`, the default would silently emit raw bytes). Wired it into `MlKemSpi.engineEncapsulate` and `engineDecapsulate` to replace the `SecretKeySpec` they previously returned. Bundled a separate commit adding `extern "C"` to the JNI exports in `csrc/mlkem_spi.cpp` so symbols link correctly on macOS (a pre-existing latent issue that surfaced while testing locally). Added 12 unit tests for the class itself plus 3 parameterized round-trip tests on the public `KEM` API, all passing under `TARGET_JDK_VERSION=17`.

## Alternatives considered

**Subclass `SecretKeySpec` instead of implementing `SecretKey` directly.** Would preserve `instanceof SecretKeySpec` for any consumer doing that check. Rejected because (a) the codebase grep showed no production code path relies on that — only `SecretKey` interface checks — and the named consumer (KMS Humboldt) reaches material via `getEncoded()`; and (b) `SecretKeySpec`'s constructor clones the bytes into a private field we can't reach without reflection across `java.base`'s module boundary, which `setAccessible(true)` would fail on without `--add-opens`. The only working subclass shape would pass dummy bytes to `super(...)` and override every accessor — more like masking the parent than extending it. Custom `SecretKey` is cleaner and the consumers don't care about the type identity.

**Reflection-based zeroization of `SecretKeySpec`'s internal `key` field.** Same module-access problem, plus brittleness against future JDK refactors of `SecretKeySpec`'s internals — silent failure modes are the worst kind for a security feature.

## Why this route

The custom-class approach gives consumers a reliable, deterministic zeroize without leaking the `byte[]` field outside of `getEncoded()` and without depending on JDK internals. It's a small additive class behind a package-private API surface — easy to extend later to other SPIs (HKDF, PBKDF2, etc., once we decide they're worth the change) without touching `SecretKeySpec` semantics. The tradeoff (loss of `instanceof SecretKeySpec`) is acceptable because no ACCP code paths and no known consumers rely on it; per JCE convention, callers handle results as `SecretKey`. The bundled `extern "C"` fix is genuinely orthogonal but small enough not to warrant its own PR — it's split into a separate commit for easy reviewer cherry-pick or revert.

---

### DIFF
--- a/csrc/mlkem_spi.cpp
+++ b/csrc/mlkem_spi.cpp
@@ -19,7 +19,7 @@ static EVP_PKEY_CTX_auto setupMlKemContext(JNIEnv* pEnv, jlong evpKeyPtr)
     return ctx;
 }
 
-JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_MlKemSpi_nativeEncapsulate(
+extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_MlKemSpi_nativeEncapsulate(
     JNIEnv* pEnv, jclass, jlong evpKeyPtr, jbyteArray ciphertextArray, jbyteArray sharedSecretArray)
 {
     try {
@@ -38,7 +38,7 @@ JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_MlKemSpi_nativeE
     }
 }
 
-JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_MlKemSpi_nativeDecapsulate(
+extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_MlKemSpi_nativeDecapsulate(
     JNIEnv* pEnv, jclass, jlong evpKeyPtr, jbyteArray ciphertextArray, jbyteArray sharedSecretArray)
 {
     try {
--- a/src/com/amazon/corretto/crypto/provider/DestroyableSecretKey.java
+++ b/src/com/amazon/corretto/crypto/provider/DestroyableSecretKey.java
@@ -7,7 +7,6 @@ import java.io.NotSerializableException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.util.Arrays;
-import java.util.Objects;
 import javax.crypto.SecretKey;
 import javax.security.auth.Destroyable;
 
@@ -37,8 +36,8 @@ final class DestroyableSecretKey implements SecretKey, Destroyable {
   private volatile boolean destroyed;
 
   DestroyableSecretKey(final byte[] key, final String algorithm) {
-    Objects.requireNonNull(key, "key must not be null");
-    Objects.requireNonNull(algorithm, "algorithm must not be null");
+    Utils.requireNonNull(key, "key must not be null");
+    Utils.requireNonNull(algorithm, "algorithm must not be null");
     if (algorithm.isEmpty()) {
       throw new IllegalArgumentException("algorithm must not be empty");
     }
--- a/src/com/amazon/corretto/crypto/provider/DestroyableSecretKey.java
+++ b/src/com/amazon/corretto/crypto/provider/DestroyableSecretKey.java
@@ -33,7 +33,6 @@ final class DestroyableSecretKey implements SecretKey, Destroyable {
   private static final long serialVersionUID = 1L;
 
   private final String algorithm;
-  // not final: bytes cleared on destroy() via Arrays.fill
   private final byte[] key;
   private volatile boolean destroyed;
 
@@ -100,8 +99,9 @@ final class DestroyableSecretKey implements SecretKey, Destroyable {
   //      synchronization.
   // Identity-based equality (Object.equals) is the safer default for stateful keys.
 
-  // Block serialization. SecretKey extends Serializable, but emitting the raw key bytes
-  // in a serialized form would defeat the destroyable contract.
+  // Even though SecretKey extends Serializable, block implicit serialization. We can relax
+  // this in the future if callers need this behavior and have adequate leakage prevention
+  // measures in place.
   private void writeObject(final ObjectOutputStream out) throws IOException {
     throw new NotSerializableException("DestroyableSecretKey");
   }
--- a/src/com/amazon/corretto/crypto/provider/DestroyableSecretKey.java
+++ b/src/com/amazon/corretto/crypto/provider/DestroyableSecretKey.java
@@ -1,0 +1,96 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider;
+
+import java.io.IOException;
+import java.io.NotSerializableException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.util.Arrays;
+import java.util.Objects;
+import javax.crypto.SecretKey;
+
+/**
+ * A {@link SecretKey} that holds raw key material in a byte array which the consumer can actively
+ * zero out via {@link #destroy()}.
+ *
+ * <p>Unlike {@link javax.crypto.spec.SecretKeySpec}, whose {@link #destroy()} is a no-op that
+ * throws {@link javax.security.auth.DestroyFailedException}, this class implements {@code
+ * destroy()} to overwrite the internal byte array with zeros and mark the instance unusable. After
+ * destruction, {@link #getEncoded()} and {@link #getAlgorithm()} throw {@link
+ * IllegalStateException}.
+ *
+ * <p>This class is used for symmetric keys whose material the consumer may want to wipe
+ * deterministically rather than wait for garbage collection -- for example, ML-KEM shared secrets
+ * that will be used as wrapping keys for higher-value material.
+ *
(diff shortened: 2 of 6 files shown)


tst/com/amazon/corretto/crypto/provider/test/jdk17plus/MlKemTest.java
```diff
@@ -128,6 +128,30 @@ public void testKemRoundTrips(TestParams params) throws Exception {
...
+    NamedParameterSpec paramSpec = new NamedParameterSpec(params.parameterSet);
+
+    KEM.Encapsulator encapsulator = encapsulatorKem.newEncapsulator(params.pub, paramSpec, null);
+    KEM.Encapsulated encapsulated = encapsulator.encapsulate();
+    KEM.Decapsulator decapsulator = decapsulatorKem.newDecapsulator(params.priv, paramSpec);
+    SecretKey recovered = decapsulator.decapsulate(encapsulated.encapsulation());
+
+    for (SecretKey key : new SecretKey[] {encapsulated.key(), recovered}) {
+      org.junit.jupiter.api.Assertions.assertFalse(key.isDestroyed());
+      org.junit.jupiter.api.Assertions.assertNotNull(key.getEncoded());
```
prasden: NP: Import at the top?

WillChilds-Klein: ew. yeah for sure.

src/com/amazon/corretto/crypto/provider/DestroyableSecretKey.java
```diff
@@ -0,0 +1,97 @@
...
+    assertNotDestroyed();
+    return key.clone();
+  }
+
+  @Override
+  public synchronized void destroy() {
+    assertNotDestroyed();
+    Arrays.fill(key, (byte) 0);
+    destroyed = true;
+  }
```
justsmth, line 78: `destroy()` is `synchronized` but the accessors aren't, so there's a minor "Time-of-Check to Time-of-Use" concern.  For example in `getEncoded()`, a concurrent `destroy()` can land between `assertNotDestroyed()` and `key.clone()`, in which case `getEncoded()` returns a zeroed (or partially zeroed) array instead of throwing `IllegalStateException`.

WillChilds-Klein: Good catch! I'll add `synchronized` declarations here and on other methods that depend on destroy state.

src/com/amazon/corretto/crypto/provider/jdk17plus/MlKemSpi.java
```diff
@@ -208,7 +208,7 @@ public SecretKey engineDecapsulate(byte[] encapsulation, int from, int to, Strin
       byte[] sharedSecret = new byte[engineSecretSize()];
       try {
         privateKey.useVoid(ptr -> nativeDecapsulate(ptr, encapsulation, sharedSecret));
-        return new SecretKeySpec(sharedSecret, algorithm);
+        return new DestroyableSecretKey(sharedSecret, algorithm);
```
justsmth, line 211: The replaced `SecretKeySpec` has value-based `equals()`/`hashCode()`, but `DestroyableSecretKey` falls back to `Object` identity equality -- so two KEM secrets with identical bytes are no longer `.equals()`.

[`EvpHmac.HmacState.setKey()` uses `Objects.equals(this.key, key)`](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/src/com/amazon/corretto/crypto/provider/EvpHmac.java#L331) to skip re-keying, so that short-circuit silently stops firing (lost optimization, not a correctness bug).

Is this part of the change intentional? Should it be documented?

WillChilds-Klein: Yes. This was intentional. I've added a source comment explaining.

Only ML-KEM uses DestroyableSecretKey today, so EvpHmac's performance shouldn't be affected.

src/com/amazon/corretto/crypto/provider/DestroyableSecretKey.java
```diff
@@ -0,0 +1,112 @@
...
+  //      destroyed instances as equal to a fresh all-zero key, which is a misleading and
+  //      potentially security-relevant lie.
+  //   2. A content-based hashCode forces us to read the key bytes, which is both a
+  //      pointless perf hit on every hash and another place that would need destroy-state
+  //      synchronization.
+  // Identity-based equality (Object.equals) is the safer default for stateful keys.
+
+  // Block serialization. SecretKey extends Serializable, but emitting the raw key bytes
+  // in a serialized form would defeat the destroyable contract.
+  private void writeObject(final ObjectOutputStream out) throws IOException {
```
geedo0, line 104: Was this specifically asked for by the requestor? Seems a bit arbitrary to support getEncoded while arguing that writeObject defeats the purpose of this class.

src/com/amazon/corretto/crypto/provider/DestroyableSecretKey.java
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * <p>Instances are not serializable: {@link SecretKey} extends {@link java.io.Serializable}, but
+ * serializing a destroyable key would silently copy the raw bytes outside the JVM, defeating the
+ * purpose. {@code writeObject}/{@code readObject} throw {@link NotSerializableException}.
+ */
+final class DestroyableSecretKey implements SecretKey, Destroyable {
+  private static final long serialVersionUID = 1L;
+
+  private final String algorithm;
+  // not final: bytes cleared on destroy() via Arrays.fill
```
geedo0: this comment confuses me, it's declared final and it says "not final". Suggest dropping this comment.

justsmth (dismissed):

geedo0 (dismissed):

src/com/amazon/corretto/crypto/provider/DestroyableSecretKey.java
WillChilds-Klein: yeah, vestigial. will remove.

src/com/amazon/corretto/crypto/provider/DestroyableSecretKey.java
WillChilds-Klein: not asked for by the requester. writeObject are blocked to prevent _implicit_ serialization of secret key material by e.g. logging frameworks. `getEncoded()` at least has to be explicitly called to emit secret key material. the comment isn't clear, I'll update it.

I could go either way on this, but am biased towards this more conservative approach that we can relax later if we need to implement writeObject.

justsmth (dismissed):

src/com/amazon/corretto/crypto/provider/DestroyableSecretKey.java
```diff
@@ -0,0 +1,112 @@
...
+final class DestroyableSecretKey implements SecretKey, Destroyable {
+  private static final long serialVersionUID = 1L;
+
+  private final String algorithm;
+  private final byte[] key;
+  private volatile boolean destroyed;
+
+  DestroyableSecretKey(final byte[] key, final String algorithm) {
+    Objects.requireNonNull(key, "key must not be null");
+    Objects.requireNonNull(algorithm, "algorithm must not be null");
```
prasden: Can we use `Utils.requireNonNull()` instead?

https://github.com/corretto/amazon-corretto-crypto-provider/blob/e9d5f430cd0ac4a3d1fd8b36bb2fe847cce35fd5/src/com/amazon/corretto/crypto/provider/Utils.java#L586-L591

csrc/mlkem_spi.cpp
```diff
@@ -19,7 +19,7 @@ static EVP_PKEY_CTX_auto setupMlKemContext(JNIEnv* pEnv, jlong evpKeyPtr)
     return ctx;
 }
 
-JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_MlKemSpi_nativeEncapsulate(
+extern "C" JNIEXPORT void JNICALL Java_com_amazon_corretto_crypto_provider_MlKemSpi_nativeEncapsulate(
```
prasden, line 22: Replace in `mlkem_gen.cpp` as well? 
https://github.com/corretto/amazon-corretto-crypto-provider/blob/e9d5f430cd0ac4a3d1fd8b36bb2fe847cce35fd5/csrc/mlkem_gen.cpp#L12-L14

src/com/amazon/corretto/crypto/provider/DestroyableSecretKey.java
WillChilds-Klein: sure

csrc/mlkem_spi.cpp
WillChilds-Klein: yeah, probably. i'll evaluate whether this is needed anywhere else for a follow-up

prasden (approved):

justsmth (approved):

PR #551, opened 2026-06-29
JCA conformance harness: scaffolding

WillChilds-Klein:
## Summary

First of two PRs implementing #550. Lands the **scaffolding** for running the JDK's own jtreg crypto conformance tests against ACCP, to catch behavioral divergence from the JDK-default providers (SunJCE, SunEC, SunRsaSign).

This PR is deliberately inert with respect to the test suite itself — it wires the harness end-to-end in `--smoke` mode and leaves the actual OpenJDK test-tree checkout + run to the follow-up.

## What's here

- **`tests/jca-conformance/run-jtreg.sh`** — driver script.
  - `--smoke` (default): builds/loads ACCP, verifies it installs as the highest-priority provider under the JDK under test, serves an overridden SPI (SHA-256), and that the config files parse. No OpenJDK checkout, fast, no external test dependency.
  - `--full`: placeholder; implemented in PR 2/2.
- **`tests/jca-conformance/jtreg-test-roots.txt`** — JCA-relevant `test/jdk` subtrees (ciphers, Mac, Signature, KeyFactory, EC/RSA, provider tests), excluding areas ACCP doesn't implement (TLS/JSSE, PKCS#11, Kerberos).
- **`tests/jca-conformance/exclusions/{common,jdk17,jdk21}.txt`** — per-JDK exclusion lists, empty to start. A failing test gets an entry here with a required rationale comment (linking `DIFFERENCES.md` or a tracking issue), keeping CI green while making the divergence backlog explicit in source.
- **`tests/jca-conformance/README.md`** — design, exclusion model, and the JCK-access note (JCK can't be fetched anonymously; tracked separately).
- **`.github/workflows/jca-conformance.yml`** — GHA job, matrix JDK 17 / 21, builds ACCP and runs the harness in `--smoke` mode.

## Provider installation

ACCP is installed as provider #1 via a generated `java.security.properties` override. The native library self-extracts from the jar via ACCP's `Loader`, so only the jar needs to be on the classpath (no `java.library.path`).

## Test plan

- [x] `shellcheck tests/jca-conformance/run-jtreg.sh` clean
- [x] `--smoke` passes locally on JDK 17 (ACCP installs as provider #1, serves SHA-256)
- [ ] CI: smoke job green on JDK 17 and 21

## Follow-up (PR 2/2)

[...]

Issue #550, opened 2026-06-23
Add CI test suites for JCK and JTREG


WillChilds-Klein:
## Summary

To ensure maximal interoperability with JDK-default JCA/JCE providers, we should add CI jobs that run the following test suites against ACCP on each supported JDK 17+ LTS:

- [JCK](https://openjdk.org/groups/conformance/JckAccess/) (Java Compatibility Kit) — the conformance harness used to certify JDK distributions.
- [jtreg](https://openjdk.org/jtreg/) — OpenJDK's regression test harness; the JDK ships JCE/JCA tests under `test/jdk/com/sun/crypto/`, `test/jdk/javax/crypto/`, `test/jdk/sun/security/`, etc., that exercise concrete provider behavior.

## Why

ACCP overrides default JCE algorithm SPIs (Cipher, KeyAgreement, KeyFactory, KeyGenerator, KeyPairGenerator, KEM, Mac, MessageDigest, SecretKeyFactory, Signature, etc.) and is registered ahead of SunJCE/SunEC in the provider list. Subtle behavioral divergence from JDK-default providers — exception types, edge-case handling of inputs, parameter spec acceptance, key encoding round-trips — is exactly what JCK/jtreg are designed to catch. Today our CI runs ACCP-authored unit tests; we don't run the JDK's own provider-conformance tests against ACCP.

## Scope

- Add a new GHA workflow (or extend an existing CodeBuild matrix) running JCK + jtreg with `-Djava.security.providers...` configured to register ACCP, against the matrix `[JDK17, JDK21]` (current and prior LTS releases that support the modern KEM/EdDSA APIs ACCP exposes).
- Tag any pre-existing ACCP-vs-JDK behavioral differences as expected-deviations with documentation pointing to the rationale, so the harness can mark them excluded rather than failing CI.

## References

- JCK access: https://openjdk.org/groups/conformance/JckAccess/
- jtreg: https://openjdk.org/jtreg/
- Pattern reference: WillChilds-Klein/melitta#9 (analogous proposal for s2n-tls/JSSE)

## Acceptance criteria

- [ ] CI workflow runs JCK and jtreg against `JDK17` and `JDK21` with ACCP installed.
- [ ] Any expected behavioral deviations are documented and excluded from the harness in a list reviewable in source.
- [ ] Workflow is gating (red CI on regression).

PR #552, opened 2026-06-29
JCA conformance harness: jtreg run + exclusions

WillChilds-Klein:
## Summary

Second of two PRs for #550, **stacked on #551**. Turns the scaffolding into a working conformance run against ACCP and switches CI to it.

> [!NOTE]
> Stacked on #551 (branch-1 scaffolding). Until #551 merges, this PR's diff includes the scaffolding commit too; review #551 first. Once #551 merges to main, this collapses to just the `--full` delta.

## Harness (`run-jtreg.sh --full`)

- **Builds jtreg from source** at the tag pinned in `jtreg-version.txt` (`openjdk/jtreg`). No third-party prebuilt binaries.
- **Sparse-checks-out the OpenJDK test content** for the JDK under test from a pinned per-version GA tag (`jdk-tags.txt`), via partial (`--filter=blob:none`) + sparse + shallow clone — only the crypto subtrees + jtreg support dirs (`test/lib`, `test/jtreg-ext`), not the whole JDK source (~226 MB vs multi-GB). Per-version pins because the test set differs by JDK version.
- **Installs ACCP at provider priority 1** via a renumbered `java.security` override (ACCP at 1, JDK defaults shifted to 2..N+1).

### Why renumber rather than `security.provider.1=ACCP`

A bare `security.provider.1=ACCP` *overwrites* slot 1 (the `SUN` provider), evicting it — which breaks JDK-internal `SecureRandom` bootstrap with `InternalError: SHA-1 not available` (ACCP doesn't register SUN's legacy `SHA`/`SHA1` alias) and fails nearly every test. Renumbering keeps every default present so legacy aliases fall through to SUN. This was the main harness bug found and fixed while bringing the run up.

## Seeded exclusions

All triaged as **not** ACCP algorithm-conformance issues:

- `common.txt`: `DefaultProviderList`, `Versions`, `SecurityProviderModularTest` — meta-tests asserting the JDK's exact default provider set, which ACCP intentionally perturbs by installing at priority 1.
- `jdk21.txt`: `CryptoPermissions/InconsistentEntries` — writes a policy file into the read-only JDK install dir (`AccessDeniedException`); environmental, would fail the same against SunJCE.

Each entry carries a rationale comment per the exclusion model.

## Results against ACCP (local)

| JDK | passed | excluded | exit |
|-----|--------|----------|------|
| 17  | 533    | 3        | 0    |
| 21  | 567    | 4        | 0    |

## Workflow

`jca-conformance.yml` now runs `--full` on the JDK 17/21 matrix and uploads the jtreg report as a CI artifact.

## Test plan

- [x] `shellcheck` clean
- [x] JDK17 full run green (533 passed, 3 excluded)
- [x] JDK21 full run green (567 passed, 4 excluded)
- [ ] CI: full job green on JDK 17 and 21

Tracking: #550


WillChilds-Klein:
Superseded by #557 (top of stacked PR #558), stacked on #556. Recreated as a same-repo PR with base `jca-conformance-scaffolding` so GitHub can stack the two layers; the branch and commits are unchanged.

PR #500, opened 2025-10-08
Support and test X25519 key agreement

WillChilds-Klein:
# Notes

See t/V1960180637. X25519 wasn't previously tested or made available for key exchange. This PR ameliorates that.

We also fix up the ML-KEM benchmarks to [clarify](https://github.com/corretto/amazon-corretto-crypto-provider/pull/500/commits/d4e12429ab6643dee7a640c2ca0cf6ce17015b02) that ML-KEM is only available for BC on JDK21+.

# Testing
- unit tests in CI

benchmarks:

```
$ uname -a
Darwin 7cf34deb0968 24.6.0 Darwin Kernel Version 24.6.0: Wed Oct 15 21:12:05 PDT 2025; root:xnu-11417.140.69.703.14~1/RELEASE_ARM64_T6030 arm64

$ ./gradlew -PaccpLocalJar="../../build/cmake/AmazonCorrettoCryptoProvider.jar" -PincludeBenchmark="KeyGenEc" lib:jmh
...
Benchmark          (curve)                    (provider)   Mode  Cnt       Score       Error  Units
KeyGenEc.generate   X25519  AmazonCorrettoCryptoProvider  thrpt    5  111700.507 ± 19121.063  ops/s
KeyGenEc.generate   X25519                            BC  thrpt    5   23498.144 ±   825.249  ops/s
KeyGenEc.generate   X25519                         SunEC  thrpt    5   10652.555 ±  1568.122  ops/s

$ ./gradlew -PaccpLocalJar="../../build/cmake/AmazonCorrettoCryptoProvider.jar" -PincludeBenchmark="KeyAgreementEcdh" lib:jmh
...
Benchmark               (curve)                    (provider)   Mode  Cnt      Score     Error  Units
KeyAgreementEcdh.agree   X25519  AmazonCorrettoCryptoProvider  thrpt    5  47817.783 ± 676.210  ops/s
KeyAgreementEcdh.agree   X25519                            BC  thrpt    5  18598.858 ± 939.353  ops/s
KeyAgreementEcdh.agree   X25519                         SunEC  thrpt    5  10810.782 ± 241.436  ops/s
```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #553, opened 2026-07-10
Make ACCP X25519 usable by JSSE for TLS 1.3

WillChilds-Klein:
## Summary

Resolves P470070813.

JSSE's TLS 1.3 x25519 handshake uses the `XDH` algorithm name and initializes the `KeyPairGenerator` with `NamedParameterSpec`. Two JCA-integration defects prevented JSSE from ever using ACCP's X25519 implementation, causing a silent fallback to SunEC for **both** the ephemeral keygen and the ECDH key agreement.

### Defect 1 — `KeyPairGenerator.XDH` rejected `NamedParameterSpec`
`XDHGen` did not override `initialize(AlgorithmParameterSpec, SecureRandom)`, so it inherited `KeyPairGeneratorSpi`'s `UnsupportedOperationException`. When JSSE (`XDHKeyExchange$XDHEPossession`) called `initialize(NamedParameterSpec.X25519, ...)`, `KeyPairGenerator$Delegate` transparently failed over to SunEC.

**Fix:** `XDHGen` now accepts a `NamedParameterSpec` naming X25519 (a no-op, since ACCP always generates X25519 XDH keys) and rejects any other curve with `InvalidAlgorithmParameterException` so the JCA can still fail over (e.g. for X448) rather than silently emit an X25519 key. `NamedParameterSpec` is a JDK 11+ type and ACCP's main sources compile to an older bytecode target, so the curve name is read reflectively — the same pattern already used in `EcParameters`. This SPI is already gated to JDK 11+ via `shouldRegisterX25519`.

### Defect 2 — `KeyAgreement.XDH` not registered
ACCP registered `KeyAgreement` only under the `X25519` name (added in #500), but JSSE calls `KeyAgreement.getInstance("XDH")`, which fell through to SunEC.

**Fix:** Register `EvpKeyAgreement$XDH` under both `XDH` and `X25519`, mirroring the dual registration already used for `KeyPairGenerator`/`KeyFactory`.

## Testing

- `KeyPairGeneratorTest`: new `initialize(NamedParameterSpec.X25519)` test asserting keygen runs on ACCP, plus a negative test that an unsupported curve (X448) throws `InvalidAlgorithmParameterException`.
- `EvpKeyAgreementTest`: new `XDH`-name `KeyAgreement` variant that generates keys via ACCP's `XDH` `KeyPairGenerator` and performs pairwise agreement interop with **SunEC** (matching the real cross-provider TLS handshake).

I manually confirmed the pre-fix failover (both keygen and KA landing on `sun.security.ec.*`) and the post-fix behavior (ACCP used end-to-end, shared secret matches SunEC) before writing the regression tests.

---

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
+++ b/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
@@ -236,7 +236,8 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
               "SupportedKeyClasses",
               "java.security.interfaces.XECPublicKey|java.security.interfaces.XECPrivateKey");
       // Register under both the "XDH" and "X25519" names, mirroring the dual registration used for
-      // KeyPairGenerator/KeyFactory. JSSE's TLS 1.3 handshake requests KeyAgreement.getInstance("XDH"),
+      // KeyPairGenerator/KeyFactory. JSSE's TLS 1.3 handshake requests
+      // KeyAgreement.getInstance("XDH"),
       // so without the "XDH" alias it silently fails over to another provider (e.g. SunEC).
       addService("KeyAgreement", "XDH", "EvpKeyAgreement$XDH", xdhKeyAgreementAttributes);
       addService("KeyAgreement", "X25519", "EvpKeyAgreement$XDH", xdhKeyAgreementAttributes);
--- a/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
+++ b/src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
@@ -231,13 +231,15 @@ public final class AmazonCorrettoCryptoProvider extends java.security.Provider {
             "java.security.interfaces.ECPublicKey|java.security.interfaces.ECPrivateKey"));
 
     if (shouldRegisterX25519) {
-      addService(
-          "KeyAgreement",
-          "X25519",
-          "EvpKeyAgreement$XDH",
+      final Map<String, String> xdhKeyAgreementAttributes =
           singletonMap(
               "SupportedKeyClasses",
-              "java.security.interfaces.XECPublicKey|java.security.interfaces.XECPrivateKey"));
+              "java.security.interfaces.XECPublicKey|java.security.interfaces.XECPrivateKey");
+      // Register under both the "XDH" and "X25519" names, mirroring the dual registration used for
+      // KeyPairGenerator/KeyFactory. JSSE's TLS 1.3 handshake requests KeyAgreement.getInstance("XDH"),
+      // so without the "XDH" alias it silently fails over to another provider (e.g. SunEC).
+      addService("KeyAgreement", "XDH", "EvpKeyAgreement$XDH", xdhKeyAgreementAttributes);
+      addService("KeyAgreement", "X25519", "EvpKeyAgreement$XDH", xdhKeyAgreementAttributes);
     }
 
     if (shouldRegisterEcParams) {
--- a/src/com/amazon/corretto/crypto/provider/XDHGen.java
+++ b/src/com/amazon/corretto/crypto/provider/XDHGen.java
@@ -18,6 +18,21 @@ class XDHGen extends EvpKeyPairGenerator {
   // KeyPairGenerator for the TLS 1.3 handshake.
   private static final String X25519_CURVE_NAME = "X25519";
 
+  // java.security.spec.NamedParameterSpec is a JDK 11+ type not available at our bytecode target,
+  // so it is resolved reflectively. Null on JDKs that lack it; shouldRegisterX25519 already gates
+  // this SPI to JDK 12+, so the class is present whenever initialize(...) is reachable.
+  private static final Class<?> NAMED_PARAMETER_SPEC_CLASS;
+
+  static {
+    Class<?> clazz = null;
+    try {
+      clazz = Class.forName("java.security.spec.NamedParameterSpec");
+    } catch (final ClassNotFoundException e) {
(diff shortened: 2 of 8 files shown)


geedo0 (changes requested):
>  Two JCA-integration defects prevented JSSE from ever using ACCP's X25519 implementation, causing a silent fallback to SunEC for both the ephemeral keygen and the ECDH key agreement.

Yikes, can we add an assert to our unit test that would've detected this silent failover? Like just assert that the provider name is ACCP when instantiating without a named provider?

WillChilds-Klein:
> > Two JCA-integration defects prevented JSSE from ever using ACCP's X25519 implementation, causing a silent fallback to SunEC for both the ephemeral keygen and the ECDH key agreement.
> 
> Yikes, can we add an assert to our unit test that would've detected this silent failover? Like just assert that the provider name is ACCP when instantiating without a named provider?

@geedo0 -- Done in [9feecc4](https://github.com/corretto/amazon-corretto-crypto-provider/pull/553/commits/9feecc42b50b835e16a03c5e1a06dfbfeb5749f8). Put the check in MiscSingleThreadedTests so we can manipulate global state and install ACCP rather than specifying the provider in `getInstance`.

geedo0 (dismissed):

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -243,13 +243,16 @@ private void buildServiceMap() {
...
+      final Map<String, String> xdhKeyAgreementAttributes =
           singletonMap(
               "SupportedKeyClasses",
-              "java.security.interfaces.XECPublicKey|java.security.interfaces.XECPrivateKey"));
+              "java.security.interfaces.XECPublicKey|java.security.interfaces.XECPrivateKey");
+      // Register under both the "XDH" and "X25519" names, mirroring the dual registration used for
+      // KeyPairGenerator/KeyFactory. JSSE's TLS 1.3 handshake requests
+      // KeyAgreement.getInstance("XDH"),
+      // so without the "XDH" alias it silently fails over to another provider (e.g. SunEC).
+      addService("KeyAgreement", "XDH", "EvpKeyAgreement$XDH", xdhKeyAgreementAttributes);
```
justsmth, line 254: Registering under "XDH" means delayed provider selection will now hand ACCP X448 keys too -- `SupportedKeyClasses` matches both curves. I traced `engineInit` -> `translateKey` -> `pkcs82Evp` and a non-X25519 key correctly comes back as `InvalidKeyException`, so the JCA falls through to SunEC. 

Since this path is newly reachable, can we add a test for it? Something like init'ing `KeyAgreement.getInstance("XDH")` with a SunEC X448 private key and asserting we end up on SunEC (or that ACCP throws `InvalidKeyException`). It'd be the KA-side twin of `initializeWithUnsupportedNamedParameterSpecThrows`.

src/com/amazon/corretto/crypto/provider/XDHGen.java
```diff
@@ -2,17 +2,72 @@
...
+   * supports it rather than silently producing an X25519 key.
+   *
+   * <p>{@code NamedParameterSpec} was introduced in JDK 11, but ACCP's main sources are compiled
+   * for an older bytecode target, so the spec's curve name is read reflectively rather than by
+   * importing the type directly.
+   *
+   * <p>The {@code random} parameter is ignored; key generation always draws from AWS-LC's DRBG.
+   */
+  @Override
+  public void initialize(final AlgorithmParameterSpec params, final SecureRandom random)
```
justsmth, line 53: Now that this overrides `initialize(AlgorithmParameterSpec, ...)`, the note in `EvpKeyPairGenerator.initialize(int, ...)` ("initialize(AlgorithmParameterSpec, SecureRandom) is intentionally left to throw UnsupportedOperationException") is out of date.

src/com/amazon/corretto/crypto/provider/XDHGen.java
```diff
@@ -2,17 +2,72 @@
...
+    }
+    // Nothing else to configure: generateKeyPair() always produces an X25519 key pair.
+  }
+
+  // Returns the curve name if |params| is a java.security.spec.NamedParameterSpec, else null.
+  // Uses reflection because NamedParameterSpec is a JDK 11+ type not available at our bytecode
+  // target. shouldRegisterX25519 already gates this SPI to JDK 12+, so the type is present at
+  // runtime whenever this method is reachable.
+  private static String getNamedCurve(final AlgorithmParameterSpec params) {
+    if (!"java.security.spec.NamedParameterSpec".equals(params.getClass().getName())) {
```
justsmth: `NamedParameterSpec` isn't final, and SunEC matches it with `instanceof`, so a subclass would get rejected here and quietly fail over -- the same failure mode this PR is fixing, just for a rarer call. 

A cached `Class.forName(...).isInstance(params)` would match SunEC's behavior. If the exact-class match is deliberate, a short comment saying so would help.

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
WillChilds-Klein: Good call, [added](https://github.com/corretto/amazon-corretto-crypto-provider/pull/553/changes/d4603b13cad73c2f2a982ae852cd3286b39b2fc5#diff-d763874bbd7c7378067c3c87ed335e098a5ac1a0c21c5e50ed1e475b0e39c51eR118).

src/com/amazon/corretto/crypto/provider/XDHGen.java
WillChilds-Klein: Edge case, but that makes sense. updated.

justsmth (approved):

geedo0 (approved):

PR #555, opened 2026-07-28
Support SHA3 digests in RSASSA-PSS

WillChilds-Klein:
*Issue #, if available:* N/A

*Description of changes:*

RSASSA-PSS could not use the SHA3 family of digests for either the message digest or the MGF1 digest.

**Root cause.** The PSS signature path resolves JCE digest names to AWS-LC `EVP_MD*` pointers via `Utils.getMdPtr` → `Utils.jceDigestNameToAwsLcName`. That mapper stripped *all* hyphens (`"SHA-512/256"` → `"SHA512-256"`), so `"SHA3-256"` became `"SHA3256"` — a name AWS-LC's `EVP_get_digestbyname` does not recognize. AWS-LC expects the hyphen retained (short name is exactly `"SHA3-256"`), so `getMdPtr` threw `IllegalArgumentException` and the PSS parameter set was rejected with `InvalidAlgorithmParameterException`.

**Fix.** Preserve the hyphen for `SHA3-` prefixed names, mirroring the logic already present in `DigestUtils.normalizeDigestName` (added in #540). Non-SHA3 names are unaffected. No `MessageDigest` service registration is needed — the PSS SPI passes digests to native AWS-LC as `EVP_MD*` pointers, and AWS-LC supports SHA3 (including in its FIPS module).

**Tests.** SHA3 PSS coverage is added to `EvpSignatureSpecificTest`, alongside the existing RSASSA-PSS tests:
- Extended the RSASSA-PSS cross-verification loop in `simpleCorrectnessSHAAlgorithms` to cover SHA3-224/256/384/512 (message digest == MGF1 digest) against BouncyCastle, driven by a shared `PSS_DIGESTS` constant (SHA-1/224/256/384/512 + SHA3-224/256/384/512).
- **Drive-by fix:** that loop's version guard was inverted (`versionCompare("2.0.0", ...) <= 0` instead of `> 0`), so the entire PSS block had been silently skipped on every 2.x release since it was written. Corrected so it runs on 2.0.0+. (Verified by reverting the `Utils` fix and confirming the loop then fails on SHA3, i.e. it really executes now.)
- Added `testRsaPssMixedMgf1Digest`, parameterized over `PSS_DIGESTS`: for each message digest, every MGF1 digest pairing (independent digests, which BouncyCastle disallows) cross-verified against SunRsaSign. Gated to JDK 17+, where SunRsaSign gained SHA3 RSASSA-PSS support (older JDKs still exercise ACCP's SHA3 PSS via the BouncyCastle loop above).

The bare `"SHA-3"` name (no output length) remains rejected, as already asserted by the existing `testRsaPSSBadParams`.

**Verification.**
-  new tests + params, CI passes

---

### DIFF
--- a/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
@@ -5,6 +5,7 @@ package com.amazon.corretto.crypto.provider.test;
 import static com.amazon.corretto.crypto.provider.test.TestUtil.JAVA_VERSION;
 import static com.amazon.corretto.crypto.provider.test.TestUtil.NATIVE_PROVIDER;
 import static com.amazon.corretto.crypto.provider.test.TestUtil.assertThrows;
+import static com.amazon.corretto.crypto.provider.test.TestUtil.assumeMinimumJavaVersion;
 import static com.amazon.corretto.crypto.provider.test.TestUtil.assumeMinimumVersion;
 import static com.amazon.corretto.crypto.provider.test.TestUtil.versionCompare;
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -26,6 +27,7 @@ import java.security.KeyFactory;
 import java.security.KeyPair;
 import java.security.KeyPairGenerator;
 import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
 import java.security.Provider;
 import java.security.PublicKey;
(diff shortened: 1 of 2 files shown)


geedo0 (approved):

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -783,6 +809,40 @@ private static void simpleCorrectnessSignVerify(
...
   }
 
+  /**
+   * ACCP allows the RSASSA-PSS message digest and MGF1 digest to differ. For each supported message
+   * digest, exercise every MGF1 digest pairing (including the matching-digest cases) and
+   * cross-verify against SunRsaSign, which, like ACCP, permits them to differ. BouncyCastle
+   * requires them to match, so it is not used here; the matching-digest cases are also
+   * cross-verified against BouncyCastle in {@link #simpleCorrectnessSHAAlgorithms()}.
+   *
+   * <p>SunRsaSign only gained SHA3 support for RSASSA-PSS in JDK 17, so this test is gated on that.
```
jakemas, line 819: is this `<p>` intended? Is it to provide paragraphs for some auto docustring rendering for a manual or something, didn't see others in the file.

jakemas (approved):

prasden (approved):

tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
WillChilds-Klein: sorry @jakemas , missed this earlier. this is standard javadoc format. but we don't use any javadoc tools on source comments, let alone test comments. i guess this is just how claude documents its java changes. lack of <p> close tag seems valid

https://stackoverflow.com/questions/8460993/p-end-tag-p-is-not-needed-in-html

PR #561, opened 2026-08-05
Add BouncyCastle JSSE ML-KEM integration test; make ACCP ML-KEM usable by JSSE

WillChilds-Klein:
## What

Adds an end-to-end integration test proving that **BouncyCastle's JSSE provider (BCJSSE) can source ML-KEM from ACCP over the standard JCA** during a real TLS 1.3 handshake, when ACCP is installed as the top provider. Wiring up that test surfaced two spots where ACCP's ML-KEM SPIs were stricter than the standard JCA contract that JSSE callers rely on, which this PR also fixes.

## Why

BCJSSE, left in its default (unpinned) configuration, resolves ML-KEM via JEP-452's `javax.crypto.KEM` API with no explicit provider, so normal JCA precedence applies and a foreign top provider can serve the request. This test is a positive-signal, cross-JDK probe that ACCP actually gets used on that path -- a `RecordingProvider` wrapping ACCP observes the `KEM` / `KeyPairGenerator` / `KeyFactory` lookups a genuine handshake drives.

## Changes

**ACCP interop fixes (`src/`)**

- `MlKemGen.initialize(AlgorithmParameterSpec, SecureRandom)` previously threw `UnsupportedOperationException` unconditionally. JSSE stacks (and ordinary callers) invoke `KeyPairGenerator.initialize(new NamedParameterSpec("ML-KEM-768"), random)` before `generateKeyPair()`. It now accepts a `NamedParameterSpec` matching the SPI's bound parameter set -- mirroring the existing reflective, JDK8-target-safe pattern in `XDHGen` -- and rejects a mismatch with `InvalidAlgorithmParameterException` so the JCA can fail over.
- `MlKemSpi.engineNewEncapsulator/Decapsulator` rejected a `null` `AlgorithmParameterSpec` and a non-null `SecureRandom`. The `javax.crypto.KEM` convenience overloads (`KEM.newEncapsulator(pk, random)`, `KEM.newDecapsulator(pk)`) always pass exactly those. A null spec now defaults to the key's parameter set (each ACCP ML-KEM key already carries it); a supplied `SecureRandom` is accepted and ignored (AWS-LC draws from its own DRBG). A non-null spec is still validated against the key's parameter set.

**Test (`tst/.../jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java`)**

- Drives BCJSSE-to-BCJSSE TLS 1.3 handshakes over the pure `MLKEM512/768/1024` groups and the hybrid `X25519MLKEM768` group, asserting ACCP served the ML-KEM lookups. Includes a negative control (ACCP absent -> no lookups observed).
- Uses an RSA leaf certificate purely to authenticate the handshake so the ML-KEM key exchange runs (an EC leaf trips BCJSSE server-side credential selection in this setup; cert type is incidental to what's under test).
- Takes the standard provider-mutation test locks (`SAME_THREAD` + `RESOURCE_GLOBAL` / `RESOURCE_PROVIDER`) so it does not race other provider-touching tests.
- **Skips** (does not fail) unless the delegation path is actually reachable: `javax.crypto.KEMSpi` present (JDK 21+ or KEM-backported JDK 17), bctls carries `KEMSpiUtil` (BC >= 1.84), and ACCP was built with ML-KEM enabled (`-DTARGET_JDK_VERSION=17`+).

**Build (`build.gradle`)**

- Bumps the test-scoped BouncyCastle deps `1.81 -> 1.85` and adds `bctls-jdk18on` (which carries `KEMSpiUtil`, the delegation path first shipped in 1.84).

## Testing

- New test: 6/6 pass on JDK 21 with `-DTARGET_JDK_VERSION=21` (pure 512/768/1024, hybrid, parametric set, negative control).
- Regression: `MlKemTest` (29), `MlKemUtilsTest` (4), `EvpKeyFactoryTest` (143) all pass -- the SPI relaxations preserve existing behavior (mismatched-parameter-set specs are still rejected).
- Full `./gradlew build` (spotless + static analysis) is green.

## Notes for reviewers

- Draft: the ACCP `src/` behavior changes are the substantive part and deserve the closest look -- particularly whether tolerating (and ignoring) a caller-supplied `SecureRandom` is the desired contract, versus documenting ACCP's DRBG-only stance some other way.
- The test is gated to skip on JDKs / BC versions / ACCP builds where the path isn't reachable, so it is safe across the existing CI matrix (it will skip on the ML-KEM-less shipped artifacts and on non-KEM JDKs).


src/com/amazon/corretto/crypto/provider/MlKemGen.java
```diff
@@ -2,13 +2,30 @@
...
+  // so it is resolved reflectively. Null on JDKs that lack it, in which case initialize(spec, ...)
+  // rejects all specs, which is correct there.
+  private static final Class<?> NAMED_PARAMETER_SPEC_CLASS;
+
+  static {
+    Class<?> clazz = null;
+    try {
+      clazz = Class.forName("java.security.spec.NamedParameterSpec");
+    } catch (final ClassNotFoundException e) {
+      // JDK 10 or older; getNamedParameter will reject all specs, which is correct there.
```
WillChilds-Klein: do a check on JDK version and rethrow if runtime version is >10

src/com/amazon/corretto/crypto/provider/MlKemGen.java
```diff
@@ -2,13 +2,43 @@
...
+      // signals a broken runtime -- fail fast instead of silently disabling spec initialization.
+      if (Utils.getJavaVersion() > 10) {
+        throw new AssertionError(
+            "java.security.spec.NamedParameterSpec must be present on JDK 11+", e);
+      }
+    } catch (final NoSuchMethodException e) {
+      // The class exists but lacks its public getName() accessor -- impossible on a sane JDK 11+.
+      throw new AssertionError(
+          "java.security.spec.NamedParameterSpec.getName() must be present", e);
+    }
```
geedo0, line 37: Are you being forced to handle these? Not sure why we're trying to handle these pretty much impossible things.

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
```diff
@@ -0,0 +1,446 @@
...
+  }
+
+  /**
+   * Same claim exercised via the hybrid group -- BCJSSE decomposes it into an X25519 half and an
+   * ML-KEM-768 half, and the ML-KEM half still goes through the KEM API path.
+   */
+  @Test
+  public void bcjsseSourcesHybridMlKemFromAccp() throws Exception {
+    runHandshakeAndAssertAccpServedMlKem(HYBRID_GROUP);
+  }
```
geedo0: Can't you just pass these in as parameters to the test on line 114?

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
```diff
@@ -0,0 +1,446 @@
...
+  // ---------------------------------------------------------------------------
+  // Handshake harness
+  // ---------------------------------------------------------------------------
+
+  /**
+   * Drives one BCJSSE TLS 1.3 handshake pinned to {@code namedGroup} on both peers, on the loopback
+   * interface. Uses a JKS keystore holding a self-signed RSA leaf so cert operations don't
+   * intersect the KEM path being measured. Returns true when both peers completed handshake + echo.
+   */
+  private boolean drivehandshake(
```
geedo0: nit: `driveHandshake`

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
```diff
@@ -0,0 +1,446 @@
...
+  // ---------------------------------------------------------------------------
+
+  /**
+   * A JCA {@link Provider} that delegates every service lookup to a wrapped delegate (ACCP by
+   * default), and records the (type, algorithm) pairs it is asked for. Used to observe -- rather
+   * than merely infer -- that BCJSSE routes ML-KEM through the JCA to the top provider. Only
+   * ML-KEM-related services are tallied; other services (EC, RSA, SHA-256, ...) pass through
+   * unrecorded so the numbers stay meaningful.
+   */
+  private static final class RecordingProvider extends Provider {
```
geedo0: This is a cool trick 🤯. Should you extract this to a separate class for more re-use? The TODO below suggests that we're thinking about asserting other attributes in the future.

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
```diff
@@ -0,0 +1,446 @@
...
+
+      final boolean negotiated = drivehandshake(namedGroup, recorder, true);
+      assertTrue(negotiated, "TLS 1.3 handshake should complete over " + namedGroup);
+
+      // Observed ML-KEM service lookups against the recording provider.
+      final int kem = recorder.kemMlKemLookups();
+      final int kpg = recorder.keyPairGeneratorMlKemLookups();
+      final int kf = recorder.keyFactoryMlKemLookups();
+      assertTrue(
+          kem + kpg + kf > 0,
```
geedo0: Seems vulnerable to false positives, not sure if we should tighten this up by doing something like recording the BC provider and asserting 0 uses.

src/com/amazon/corretto/crypto/provider/MlKemGen.java
WillChilds-Klein: No. intention is to fail fast if some of these assumptions/invariants are violated.

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
WillChilds-Klein: Yeah, this was a really neat workaround that claude and i came up with discussing how we could do observability. I think it makes sense to factor this out into a reusable fixture.

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
WillChilds-Klein: ah, that's a neat idea so `ACCP recorder -> ACCP -> BC recorder -> BC -> JDK stuff`. let me try that out.

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
WillChilds-Klein: will fix.

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
WillChilds-Klein: yes, can consolidate.

tst/com/amazon/corretto/crypto/provider/test/RecordingProvider.java
```diff
@@ -0,0 +1,76 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.amazon.corretto.crypto.provider.test.jdk17plus;
```
WillChilds-Klein: should put this in the `.test` package

geedo0 (dismissed):

src/com/amazon/corretto/crypto/provider/MlKemGen.java
```diff
@@ -25,10 +55,56 @@ public void initialize(int keysize, SecureRandom random) {
...
+    if (!parameterSet.matchesAlgorithmName(name)) {
+      throw new InvalidAlgorithmParameterException(
+          "Unsupported ML-KEM parameter set: "
+              + name
+              + ". This KeyPairGenerator only supports "
+              + parameterSet.getAlgorithmName()
+              + ".");
+    }
+    // Nothing else to configure: generateKeyPair() always produces a |parameterSet| key pair.
+  }
```
justsmth, line 92: Could we add focused unit tests for the new SPI behavior? The BCJSSE test covers the happy path indirectly, but the current `MlKemTest` cases always pass an explicit spec and null random.

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
```diff
@@ -0,0 +1,381 @@
...
+      final Provider bc = newBcProvider();
+
+      // Build the exact ordering deterministically regardless of what was installed before.
+      Security.removeProvider(accp.getName());
+      Security.removeProvider(bc.getName());
+      Security.insertProviderAt(accpRecorder, 1);
+      Security.insertProviderAt(accp, 2);
+      Security.insertProviderAt(bcRecorder, 3);
+      Security.insertProviderAt(bc, 4);
+      Security.addProvider(new BouncyCastleJsseProvider());
```
justsmth: `addProvider` is a no-op if BCJSSE is already installed, so this does not always produce the provider ordering described above. Remove it before rebuilding the list:
```java
Security.removeProvider(BouncyCastleJsseProvider.PROVIDER_NAME);
// ...
Security.addProvider(new BouncyCastleJsseProvider());
```

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
```diff
@@ -0,0 +1,381 @@
...
+  // ---------------------------------------------------------------------------
+
+  /**
+   * Drives one BCJSSE TLS 1.3 handshake pinned to {@code namedGroup} on both peers, on the loopback
+   * interface. Uses a JKS keystore holding a self-signed RSA leaf so cert operations don't
+   * intersect the KEM path being measured. Returns true when both peers completed handshake + echo.
+   */
+  private boolean driveHandshake(
+      String namedGroup, RecordingProvider recorder, boolean expectAccpServesMlKem)
+      throws Exception {
```
justsmth: Nit: `expectAccpServesMlKem` is always true, and the caller already makes stronger recorder assertions.

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
```diff
@@ -0,0 +1,381 @@
...
+                  + "but no ML-KEM service was requested. AccpRecorder: "
+                  + accpRecorder.summary());
+      // A KEM lookup specifically is the strongest signal that the JEP-452 delegation path fired.
+      assertTrue(
+          kem > 0,
+          () ->
+              "Expected at least one KEM.getInstance(\""
+                  + KEM_ALG
+                  + "\") to hit the top provider during the handshake. AccpRecorder: "
+                  + accpRecorder.summary());
```
justsmth:
```java
"Expected at least one ML-KEM KEM lookup to hit the top provider during "
     + namedGroup
     + " handshake. AccpRecorder: "
```

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
```diff
@@ -0,0 +1,381 @@
...
+      //   5. JDK providers (and BCJSSE, appended last; it is looked up by name anyway)
+      // The recorders serve nothing, so each sits directly in front of the real provider it
+      // observes: a lookup ACCP satisfies at position 2 never reaches BcRecorder at position 3,
+      // which is what makes BcRecorder a meaningful negative control (see the assertion below).
+      final BiPredicate<String, String> mlKemServices =
+          BouncyCastleJsseMlKemIntegrationTest::isMlKemService;
+      final RecordingProvider accpRecorder = new RecordingProvider("AccpRecorder", mlKemServices);
+      final RecordingProvider bcRecorder = new RecordingProvider("BcRecorder", mlKemServices);
+      final Provider accp = AmazonCorrettoCryptoProvider.INSTANCE;
+      final Provider bc = newBcProvider();
```
justsmth: Nit: we already expose BC through `TestUtil.BC_PROVIDER`. Can this just use it?
```java
 final Provider bc = TestUtil.BC_PROVIDER;
```

README.md
```diff
@@ -143,6 +143,11 @@ JDK's [KEM interface](https://docs.oracle.com/en/java/javase/17/docs/api/java.ba
 ./gradlew -DTARGET_JDK_VERSION=17 build
 ``` 
 
+For ML-KEM key generation (`KeyPairGenerator`) and KEM encapsulation (`KEM.newEncapsulator`), any
+caller-supplied `SecureRandom` is accepted but ignored: AWS-LC always draws ML-KEM randomness from
+its own DRBG. Passing a custom `SecureRandom` (for example to force a specific entropy source) has
+no effect on these operations.
```
justsmth, line 149: Since this is an observable JCA behavior difference, could we also record it in DIFFERENCES.md?
```
## ML-KEM ignores caller-supplied SecureRandom
For ML-KEM key generation and encapsulation, ACCP accepts but ignores caller-supplied `SecureRandom`; AWS-LC uses its own DRBG.
```

src/com/amazon/corretto/crypto/provider/MlKemGen.java
justsmth, line 37: The JDK 10-and-older fallback seems unreachable because `MlKemGen` is registered only when `Utils.isMlKemSupported()` succeeds. 

Should this just fail unconditionally on any `ReflectiveOperationException`?
```java
 } catch (final ReflectiveOperationException e) {
   throw new AssertionError(
       "NamedParameterSpec must be available when ML-KEM is registered", e);
 } 
```

Or move this SPI into `jdk17plus` and use `NamedParameterSpec` directly?

src/com/amazon/corretto/crypto/provider/MlKemGen.java
WillChilds-Klein: yeah that's reasonable

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
WillChilds-Klein: agreed. we'll do this once for the test class with @BeforeAll to save provider state going in, @BeforeAll to clear provider state and do the insertions, then @AfterAll to restore original state.

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
WillChilds-Klein: will add

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
WillChilds-Klein: change to void return

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/BouncyCastleJsseMlKemIntegrationTest.java
WillChilds-Klein: yes

README.md
WillChilds-Klein: yes

src/com/amazon/corretto/crypto/provider/MlKemGen.java
WillChilds-Klein: hm, in PR #564 we're pulling up minimum build JDK to 17 and producing exclusively multi-release JARs. i'd like to address your feedback here in a follow up to that PR removes reflection entirely.

justsmth (approved):

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/MlKemTest.java
```diff
@@ -168,6 +170,59 @@ public void testKeyGeneration(String paramSet) throws Exception {
...
+        () -> keyGen.initialize(new NamedParameterSpec("ML-KEM-512"), random));
+
+    // A spec that is not a NamedParameterSpec is rejected.
+    final AlgorithmParameterSpec notNamed = new AlgorithmParameterSpec() {};
+    assertThrows(
+        InvalidAlgorithmParameterException.class, () -> keyGen.initialize(notNamed, random));
+
+    // Keysize-based initialization is unsupported for ML-KEM.
+    assertThrows(UnsupportedOperationException.class, () -> keyGen.initialize(768, random));
+  }
```
justsmth, line 224: NP: These cover the KeyPairGenerator half nicely -- the other half of the SPI relaxation (`MlKemSpi` accepting a null spec and a non-null `SecureRandom`) is still only exercised through the BCJSSE test, which skips on non-KEM JDKs. A direct round trip via the JEP-452 convenience overloads would pin that contract here too:
```java
 KeyPair pair = KeyPairGenerator.getInstance(paramSet, NATIVE_PROVIDER).generateKeyPair();
 KEM kem = KEM.getInstance(paramSet, NATIVE_PROVIDER);
 KEM.Encapsulated enc = kem.newEncapsulator(pair.getPublic(), new SecureRandom()).encapsulate();
 SecretKey recovered = kem.newDecapsulator(pair.getPrivate()).decapsulate(enc.encapsulation());
 assertArrayEquals(enc.key().getEncoded(), recovered.getEncoded());
 
```

justsmth (approved):
The PR description has also gone stale.

geedo0 (approved):

PR #562, opened 2026-08-12
Support PKCS#1 v1.5 RSA signatures with SHA3

WillChilds-Klein:
*Issue #, if available:* N/A

*Description of changes:*

Follow-up to #555 (SHA3 for RSASSA-PSS). That PR did not cover the PKCS#1 v1.5 RSA signature algorithms (`SHAxwithRSA`), which remained SHA2-only: they are registered as fixed, individually-named `EvpSignature` classes hardcoded to a digest, and no SHA3 variants existed.

**Change.** Register `SHA3-224withRSA`, `SHA3-256withRSA`, `SHA3-384withRSA`, `SHA3-512withRSA`, backed by new `EvpSignature$SHA3_<n>withRSA` classes. The JCA algorithm names keep the hyphen (`SHA3-256withRSA`), which isn't a legal Java identifier, so the backing class names use an underscore — the registration maps between the two, the same way `RSASSA-PSS` maps to `EvpSignature$RSASSA_PSS`.

No native changes are needed:
- ACCP's signing path passes the digest to AWS-LC as an `EVP_MD*` pointer via `EVP_DigestSignInit` and sets `RSA_PKCS1_PADDING`; it's digest-agnostic.
- The digest name (`"SHA3-256"`) resolves through `Utils.getMdPtr` → `jceDigestNameToAwsLcName`, which already preserves the SHA3 hyphen (from #555).
- AWS-LC's `kPKCS1SigPrefixes` DigestInfo table already includes `NID_sha3_224/256/384/512`.

**Tests** (mirroring #555's considerations):
- Added `testRsaPkcs1Sha3`, parameterized over the four SHA3 RSA algorithms: ACCP sign/verify round-trip, plus cross-verification against BouncyCastle (both directions, all JDKs) and SunRsaSign (both directions, gated to JDK 17+ where SunRsaSign gained SHA3 support).
- Excluded `SHA3-*` from `simpleCorrectnessSHAAlgorithms`, whose `SHAxwith...` name regex can't parse the SHA3 names — this mirrors the existing `RSASSA-PSS` exclusion there; coverage is provided by `testRsaPkcs1Sha3` instead.
- Verified the new tests actually exercise the feature by removing the registration and confirming exactly the four cases fail with "no such algorithm: SHA3-*withRSA".

**Verification.**
- `EvpSignatureSpecificTest`: all pass (0 failed).
- No regressions: `EvpSignatureTest` (67,398 cases) also passes.
- `spotlessJavaCheck` clean.

Note: validated on a non-FIPS build locally (JDK 17). SHA3 is present in AWS-LC's FIPS module, so this is expected to work in FIPS builds as well, but that path was not exercised locally.

### DIFF
--- a/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
@@ -687,6 +687,11 @@ public final class EvpSignatureSpecificTest {
           || algorithm.startsWith("ML-DSA")) {
         continue;
       }
+      // SHA3 PKCS#1 v1.5 RSA signatures don't fit this test's "SHAxwith..." name pattern; they are
+      // covered separately by testRsaPkcs1Sha3.
+      if (algorithm.startsWith("SHA3-")) {
+        continue;
+      }
       String bcAlgorithm = algorithm;
       AlgorithmParameterSpec keyGenSpec = null;
       String keyGenAlgorithm = null;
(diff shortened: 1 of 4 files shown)


tst/com/amazon/corretto/crypto/provider/test/EvpSignatureSpecificTest.java
```diff
@@ -687,6 +687,11 @@ public void simpleCorrectnessSHAAlgorithms() throws Throwable {
           || algorithm.startsWith("ML-DSA")) {
         continue;
       }
+      // SHA3 PKCS#1 v1.5 RSA signatures don't fit this test's "SHAxwith..." name pattern; they are
+      // covered separately by testRsaPkcs1Sha3.
+      if (algorithm.startsWith("SHA3-")) {
+        continue;
```
WillChilds-Klein, line 693: source changes look sound, let's get this PR merged.

this is suspect. i'm working on an independent test refactor to remove `testRsaPkcs1Sha3` and other one-off algo tests. we'll probably need to keep one-offs that have different input structures.

geedo0 (approved):

prasden (approved):

PR #564, opened 2026-08-14
Build Multi Release JAR unconditionally, drop build JDKs < 17, test ML-KEM in FIPS mode

WillChilds-Klein:
## Ship a Multi-Release JAR (and compile native ML-KEM, incl. FIPS) so ML-KEM is available on JDK 17+ from a JDK 8-target build

**TL;DR:** 1/ We were building the shipped artifact at the default target JDK version (8) internally, which stripped the ML-KEM SPI out of the JAR and made ML-KEM impossible to use -- even when running on a KEM-capable JDK. 2/ Now that the SPI is always included via the Multi-Release JAR, ML-KEM is always registered on JDKs that support it (Corretto 17 / 21+); previously it was not. Without MR, supporting JDK 8/11 is mutually exclusive with supporting ML-KEM in a single JAR.

### Problem
ML-KEM registration is gated by `Utils.isMlKemSupported()`, which requires `Class.forName("com.amazon.corretto.crypto.provider.MlKemSpi")` to succeed. `MlKemSpi` lives under `src/.../jdk17plus/` and is excluded by `CMakeLists.txt` whenever `TARGET_JDK_VERSION < 17` (the default is 8). Because cmake performs all Java compilation (`compileJava.enabled = false` in `build.gradle`), a default-target artifact ships **without** the KEM SPI, so ML-KEM never registers even when running on a KEM-capable JDK (the Corretto 17 backport, or 21+).

The **native** ML-KEM JNI sources (`csrc/mlkem_gen.cpp`, `csrc/mlkem_spi.cpp`) were gated the same way, so enabling only the Java side would `UnsatisfiedLinkError` at runtime (e.g. `MlKemGen.generateEvpMlKemKey`).

### Fix
ACCP now always emits a **Multi-Release JAR** (regardless of `TARGET_JDK_VERSION`), and the build JDK must expose the KEM API.

### Detecting KEM support (build-JDK probe, not a version check)
`javax.crypto.KEM` is GA in JDK 21+ and backported into some 17 builds (e.g. Corretto 17), but a **stock JDK 17 does not have it**. So rather than trusting `BUILD_JDK_VERSION`, cmake compiles a one-line `import javax.crypto.KEM;` probe with the build JDK. If the probe fails to compile, the build **fails fast** with a `FATAL_ERROR` instructing the user to point `JAVA_HOME` at a KEM-capable JDK (JDK 21+, or a KEM-backported JDK 17 such as Corretto 17) -- ACCP no longer produces a build on a KEM-less JDK. If the probe succeeds, the overlay is built. The CI `select_build_jdk.sh` helper runs the same probe to repoint `JAVA_HOME` at a capable JDK before building.

### CI
The CI build scripts (`tests/ci/run_accp_basic_tests.sh`, `run_accp_test_integration.sh`, `run_cryptofuzz.sh`) source a new `tests/ci/select_build_jdk.sh` helper that probes for a JDK exposing `javax.crypto.KEM` (by compiling a one-line probe) and points `JAVA_HOME` at it, so the ML-KEM overlay compiles regardless of which JDK the CI image defaults to; tests still run on `TEST_JAVA_HOME`. The GitHub `macos_ci_jobs.yml` JDK 8/11 jobs likewise build with Corretto 17.

### FIPS
ML-KEM is **now enabled for non-experimental FIPS builds** too: ML-KEM is within the FIPS 140-3 validated boundary of the AWS-LC-FIPS cryptographic module that ACCP consumes -- CMVP certificate [#5314](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/5314) ([security policy PDF](https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp5314.pdf)). Two independent changes make this work. First, the Multi-Release overlay always ships `MlKemSpi` in the JAR, so ML-KEM *registers* on any KEM-capable runtime (Corretto 17 / 21+) regardless of FIPS -- registration keys solely off `Utils.isMlKemSupported()` (a Java-version + `javax.crypto.KEM` + `MlKemSpi` check) and was never FIPS-gated. Second, `csrc/mlkem_gen.cpp` / `csrc/mlkem_spi.cpp` are moved out of the `(EXPERIMENTAL_FIPS OR NOT FIPS)` guard and compiled for FIPS, so those registered services are actually *functional*. Before this, a `TARGET_JDK_VERSION >= 17` FIPS build would register ML-KEM but throw `UnsatisfiedLinkError` on first use (native symbols absent), and the shipped default-target (JDK 8) FIPS artifact omitted the SPI entirely so ML-KEM never registered at all.

The ML-KEM **seed-format** private-key encoding (RFC 9935) is a separate feature that is not yet in the FIPS AWS-LC-FIPS 3.x, so its native helpers stay under the FIPS guard and its tests are skipped under non-experimental FIPS (see Tests). These are removable once AWS-LC-FIPS is bumped to v5.0.0 (full ML-KEM private-key parsing).

### JAR / library size
Negligible.

**Jar (Java overlay):** the overlay is only the `jdk17plus` classes (one source file, `MlKemSpi.java`), which compile to 6 class files totaling **13,896 bytes uncompressed** (~6-8 KB compressed):
```
MlKemSpi.class                      4034
MlKemSpi$MlKemEncapsulatorSpi.class 3217
MlKemSpi$MlKemDecapsulatorSpi.class 2967
MlKemSpi$MlKemSpi1024.class         1228
MlKemSpi$MlKemSpi512.class          1225
MlKemSpi$MlKemSpi768.class          1225
```
Base classes are **not** duplicated into `META-INF/versions/17`; the only other additions are the `Multi-Release: true` manifest line and the `META-INF/versions/17/` directory entries.

For scale, the platform artifact is dominated by that native library (multiple MB); the ML-KEM additions are well under ~0.2% of the jar. This is far cheaper than publishing a second JDK17-target artifact, which would roughly double the footprint by duplicating the native library.

### Tests

**Why this PR touches tests (prior coverage gaps).** ML-KEM was previously exercised only very narrowly, so widening where it runs is what drives the test diff -- this is not net-new coverage of previously untested code:
- The `jdk17plus` tests (`MlKemTest`, `BouncyCastleJsseMlKemIntegrationTest`) were compiled and run **only** when the build was given `TARGET_JDK_VERSION >= 17`. At the default target (8), both gradle (`shouldExcludeJdk17Plus()`) and cmake (`INCLUDE_JDK17PLUS_DIR`) dropped the entire `jdk17plus/` tree, *including its tests*. In CI that meant ML-KEM tests ran in only three dedicated **non-FIPS** jobs that exist specifically to "include the jdk17plus folder" (`codecov-ci.yml` target 17; `macos_ci_jobs.yml` target 17 and target 21).
- ML-KEM was therefore **never** tested in default-target builds -- the jdk17plus tests were excluded -- and [**never** in **any FIPS build**](https://github.com/search?q=repo%3Acorretto%2Famazon-corretto-crypto-provider+--target-jdk-version&type=code) (ML-KEM was disabled in FIPS entirely, so there was nothing to test).

The test changes here close those two gaps:
1. **Gate moved from build-target to the test JVM.** Because the base is now always target-8 and the MRJAR overlay carries the 17 classes, "should the jdk17plus tests compile/run" can no longer key off `TARGET_JDK_VERSION` (now always 8); cmake gates jdk17plus test compilation on `TEST_JAVA_MAJOR_VERSION >= 17` instead. This makes ML-KEM tests run on every KEM-capable test JVM, not just in the target-17/21 jobs.
2. **FIPS skip guards.** With ML-KEM now registered and functional in FIPS, the existing `MlKemTest` / `EvpKeyFactoryTest` / `MlKemUtilsTest` execute under a FIPS build for the first time and hit AWS-LC-FIPS 3.1.0 gaps (seed-format private keys, and DER marshal/parse of ML-KEM keys). Those specific cases are wrapped in `assumeTrue(!isFips() || isExperimentalFips())` (details below); they still run fully in non-FIPS and experimental-FIPS.

### Notes
- ACCP must be built on a KEM-capable JDK (Corretto 17, or 21+). The build probes for `javax.crypto.KEM` and, if it is absent, **fails fast** with a `FATAL_ERROR` rather than silently omitting ML-KEM; in CI the `select_build_jdk.sh` helper repoints `JAVA_HOME` at a capable JDK so this does not trip.

### Follow-ups
- #568 to re-enable BC JSSE and EvpKeyFactory checks (most important)
- #569 to re-enable seed key parsing


commitmsg.txt
geedo0, line 1: this should not be checked in.

CMakeLists.txt
```diff
@@ -168,6 +168,42 @@ else()
...
+# ML-KEM on 17+ by shipping a Multi-Release JAR: the base is compiled at
+# TARGET_JDK_VERSION with jdk17plus/ excluded, and the jdk17plus/ classes (the
+# KEM SPI, which references javax.crypto.KEM) are compiled for 17 and packed
+# into META-INF/versions/17. JDK 8/11 never see the versioned classes, so there
+# is no UnsupportedClassVersionError and Utils.isMlKemSupported() returns false
+# there; JDK 17 (KEM backport) and 21+ load the overlay and register ML-KEM.
+# Requires a build-time JDK that actually exposes the KEM API. It is GA in JDK
+# 21+ and backported into some 17 builds (e.g. Corretto 17), but a stock JDK 17
+# does not have it -- so probe by compiling a tiny source rather than trusting
+# the version number.
```
geedo0: "stock JDK 17 does not have it" suggests a stronger assertion that it really is. My understanding is that "stock JDK 17" does have KEM API support as of the currently downloadable/updated version out right now from Oracle or OpenJDK. And that this has been true for at least 2 years(?). The only way you'd get in this situation is if you had a system with a really old JDK17 install. Have you observed something different from this?

CMakeLists.txt
```diff
@@ -168,6 +168,42 @@ else()
...
+set(BUILD_JDK_HAS_KEM FALSE)
+set(_KEM_PROBE_DIR "${CMAKE_CURRENT_BINARY_DIR}/kem-probe")
+file(MAKE_DIRECTORY ${_KEM_PROBE_DIR})
+file(WRITE ${_KEM_PROBE_DIR}/KemProbe.java "import javax.crypto.KEM; public final class KemProbe {}\n")
+execute_process(
+    COMMAND ${Java_JAVAC_EXECUTABLE} -d ${_KEM_PROBE_DIR} ${_KEM_PROBE_DIR}/KemProbe.java
+    RESULT_VARIABLE _KEM_PROBE_RC
+    OUTPUT_QUIET ERROR_QUIET)
+if(_KEM_PROBE_RC EQUAL 0)
+    set(BUILD_JDK_HAS_KEM TRUE)
```
geedo0: Nice, I feel like this probe can simplify/improve a lot of the other complex logic we have in this build.

CMakeLists.txt
```diff
@@ -253,6 +298,27 @@ add_jar(
     INCLUDE_JARS ${BUILD_CLASSPATH_LIST}
 )
 
+# Compile the jdk17plus overlay (the KEM SPI) for JDK 17 and stage it under
+# META-INF/versions/17. Uses -source/-target 17 (NOT --release 17) so the KEM
+# API backported onto the build JDK stays visible, mirroring set_java_compile_flags().
+# The base code-only-jar plus the build classpath satisfy the overlay's compile deps.
+if(BUILD_MLKEM_MRJAR)
```
geedo0: High level comment, this implementation tightly couples multi-release jars to ML-KEM support when the system should support arbitrary features in the future. This is going to invite awkward bolting on in the future. Maybe we can prompt for a more decoupled implementation that collects all the various features we can build into a Multi-Release Jar and have a more global flag that activates this system?

CMakeLists.txt
```diff
@@ -353,13 +432,19 @@ else()
...
-        set(C_SRC ${C_SRC} csrc/mlkem_gen.cpp csrc/mlkem_spi.cpp)
-    endif()
+endif()
+
+# ML-KEM has an approved implementation in the FIPS branch of AWS-LC that ACCP
+# consumes, so its native sources are compiled for ALL builds (including FIPS)
+# whenever ML-KEM ships -- either compiled into the base (TARGET >= 17) or as a
+# Multi-Release overlay (BUILD_MLKEM_MRJAR). The JNI symbols must exist wherever
+# ML-KEM registers.
+if(INCLUDE_JDK17PLUS_DIR OR BUILD_MLKEM_MRJAR)
```
geedo0: Can we just shortcut this and always build MLKEM now? It's unclear to me when we wouldn't build it in with the current tip of ACCP. I guess you can checkout the latest source with an old JDK17 and have the expectation of not building MLKEM native layer code, but that's such an edge case that we shouldn't try to accommodate for it.

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/MlKemTest.java
```diff
@@ -295,6 +295,14 @@ public void testGenericAlgorithmHandling(String paramSet) throws Exception {
   @ParameterizedTest
   @MethodSource("mlKemParamSets")
   public void testPrivateKeyEncodingIsSeedFormat(String paramSet) throws Exception {
+    // ML-KEM seed-format keys require seed support in the underlying AWS-LC,
```
geedo0, line 361: Unrelated to this change?

commitmsg.txt
WillChilds-Klein: no it should not. i only reviewed the files changed diff before staging, not everything that ended up getting staged. removed.

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/MlKemTest.java
WillChilds-Klein: before this PR, ML-KEM was only built/enabled in "experimental FIPS" mode, not regular "FIPS" mode. now that ML-KEM is built/enabled/tested in regular FIPS mode, we need to account for currently validated AWS-LC-FIPS's lack of ML-KEM seed formatting.

this was not clear in the PR description. will clarify.

CMakeLists.txt
WillChilds-Klein: I wonder if we can get to a place of always shipping multi-release JARs if the build JDK supports it. looking into this...

CMakeLists.txt
WillChilds-Klein: yeah, as discussed offline we're OK with assuming ML-KEM on JDK17

CMakeLists.txt
WillChilds-Klein: yeah, as discussed we can simplify this -- no builds on JDK8, drop runtime support for older JDK17

WillChilds-Klein:
Moving to draft pending CodeBuild CI Dockerfile changes

CMakeLists.txt
```diff
@@ -89,6 +89,24 @@ if (NOT DEFINED TEST_JAVA_HOME)
...
+  if (NOT DEFINED TEST_JAVA_MAJOR_VERSION)
+    execute_process(
+      COMMAND ${TEST_JAVA_EXECUTABLE} -version
+      ERROR_VARIABLE _ACCP_TEST_JAVA_VER
+      OUTPUT_VARIABLE _ACCP_TEST_JAVA_VER)
+    # Matches legacy "1.8.0_x", modern "17.0.x", and dot-less "25" / "26-ea": group 1 is the leading
+    # number, group 3 the (optional) next.
+    string(REGEX MATCH "version \"([0-9]+)(\\.([0-9]+))?" _ACCP_TEST_JAVA_M "${_ACCP_TEST_JAVA_VER}")
+    if (CMAKE_MATCH_1 STREQUAL "1")
+      set(TEST_JAVA_MAJOR_VERSION ${CMAKE_MATCH_3})
```
justsmth, line 107: If `java -version` fails or the regex doesn't match, `TEST_JAVA_MAJOR_VERSION` is silently empty and all jdk17plus tests drop out. `FATAL_ERROR` on no match?

tst/com/amazon/corretto/crypto/provider/test/EvpKeyFactoryTest.java
```diff
@@ -82,7 +82,18 @@ public static void setupParameters() throws Exception {
...
+        final String algorithm = service.getAlgorithm();
+        // AWS-LC-FIPS 3.1.0 cannot marshal/parse ML-KEM keys (no i2d_PUBKEY / d2i_PUBKEY for
+        // ML-KEM), so ML-KEM KeyFactory X509/PKCS8 round-trips fail in pure-FIPS builds. Skip
+        // ML-KEM here in pure FIPS; it is still exercised in non-FIPS and experimental-FIPS.
+        // TODO(#568): remove once ACCP hand-rolls ML-KEM SubjectPublicKeyInfo encode/decode for
+        // FIPS.
+        if (algorithm.toUpperCase().startsWith("ML-KEM")
+            && NATIVE_PROVIDER.isFips()
+            && !NATIVE_PROVIDER.isExperimentalFips()) {
+          continue;
```
justsmth, line 94: This drops ML-KEM from all six `allPairs()` tests, but only `keysSerialize`, `testX509Encoding`, and `testPKCS8Encoding` need the ASN.1 that FIPS lacks. The three `null*` tests all throw in `keyNeedsConversion`/`requireNonNullEncoding` before touching the real key's encoding. Can we keep ML-KEM in `KEYPAIRS` and put the assumption in those three?

README.md
```diff
@@ -149,11 +149,7 @@ Mac algorithms with precomputed key and associated secret key factories (expert
 
 ### ML-KEM Considerations
 
-JDK's [KEM interface](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/javax/crypto/KEM.html) was backported to JDK17, but not to earlier JDK versions. ACCP JARs support all LTS JDK versions since JDK8, so **ACCP's default build configuration and maven artifacts** omit ML-KEM as it relies on the KEM interface. To enable ML-KEM in locally built JARs, you'll need to build with a JDK ≥17 and specify `TARGET_JDK_VERSION` ≥17 like so:
-
-```
-./gradlew -DTARGET_JDK_VERSION=17 build
-``` 
+JDK's [KEM interface](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/javax/crypto/KEM.html) was backported to JDK 17 and is GA in JDK 21+, but is absent from earlier JDK versions. ACCP always ships ML-KEM in its Multi-Release JAR (the `jdk17plus` overlay under `META-INF/versions/17`), so ML-KEM is available at runtime on any JDK 17+; on JDK 8-16 runtimes the base JAR loads without it. Because the overlay is always built, building ACCP requires a KEM-capable build JDK 17+ (see [Build it yourself](#build-it-yourself)). `TARGET_JDK_VERSION` only selects the base bytecode level (default 8) and no longer gates ML-KEM.
```
justsmth: `Utils.isMlKemSupported()` returns false on JDK 18-20, so "any JDK 17+" overstates it -- here and in the three algorithm-list entries. Can we say "JDK 17 or 21+"?

CMakeLists.txt
```diff
@@ -655,12 +732,22 @@ link_with_openssl(amazonCorrettoCryptoProvider)
...
+# The jdk17plus tests (e.g. MlKemTest) reference javax.crypto.KEM and only run on
+# a KEM-capable runtime. The test jar is NOT multi-release -- it is compiled in a
+# single pass and must load on the TEST JVM:
+#  - TEST JVM 17+: include the jdk17plus tests and compile the whole jar with
+#    -source/-target 17 (NOT --release, which would hide the backported KEM API).
+#  - TEST JVM < 17 (JDK 8/11): the versioned ML-KEM overlay is never loaded there,
+#    so exclude the jdk17plus tests and keep the target-level compile flags so
+#    every test class still loads on that JVM.
+if(TEST_JAVA_MAJOR_VERSION VERSION_GREATER_EQUAL 17)
+    set(CMAKE_JAVA_COMPILE_FLAGS ${BASE_JAVA_COMPILE_FLAGS} -Xlint:-options -source 17 -target 17)
```
justsmth: `>= 17` includes JDK 18-20, which lack `javax.crypto.KEM` -- as the condition this replaces spelled out, and as `Utils.isMlKemSupported()` still does. `MlKemTest` compiles fine against the KEM-capable build JDK, then fails to load on such a test JVM. Can we run the KEM probe against `TEST_JAVA_EXECUTABLE` here too?

CMakeLists.txt
```diff
@@ -253,6 +290,31 @@ add_jar(
...
+    # (for MlKemSpi) like every other Java compile, mirroring the test-jar flags below.
+    COMMAND ${Java_JAVAC_EXECUTABLE} ${BASE_JAVA_COMPILE_FLAGS} -Xlint:-options -source 17 -target 17
+            -classpath "$<TARGET_PROPERTY:code-only-jar,JAR_FILE>:${BUILD_CLASSPATH}"
+            -d ${MR_V17_DIR} ${ACCP_JDK17PLUS_SRC}
+    COMMAND ${CMAKE_COMMAND} -E touch ${MR_STAMP}
+    DEPENDS code-only-jar ${ACCP_JDK17PLUS_SRC}
+    COMMENT "Compiling jdk17plus overlay (ML-KEM SPI) -> META-INF/versions/17"
+)
+
 if (SIGNED_JAR)
```
justsmth, line 351: The MR append is only in the `else()` branch, so a `SIGNED_JAR` build ships no `META-INF/versions/17` and no `Multi-Release: true`. I assume that's deliberate (appending would break the signature), but could we say so here, since that's the release path?

justsmth (commented):
In [build.gradle](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/build.gradle#L936), it still keys off `targetJdkVersion`. cmake's equivalent gate went away in this PR -- should this one too?

tst/com/amazon/corretto/crypto/provider/test/EvpKeyFactoryTest.java
WillChilds-Klein: this skip is reverted in the next PR in the series, so i'd prefer to leave as-is to reduce churn.

CMakeLists.txt
WillChilds-Klein: yes, good call

CMakeLists.txt
WillChilds-Klein: we only officially support LTS versions -- 8, 11, 17, 21 for now. that said, KEM probe doesn't hurt here, will add.

CMakeLists.txt
WillChilds-Klein: yes

WillChilds-Klein:
> In [build.gradle](https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/build.gradle#L936), it still keys off `targetJdkVersion`. cmake's equivalent gate went away in this PR -- should this one too?

we only use gradle here for dependency management in development/public distribution, so yeah i think it's safe to remove.

justsmth (approved):

README.md
```diff
@@ -149,11 +149,12 @@ Mac algorithms with precomputed key and associated secret key factories (expert
 
 ### ML-KEM Considerations
 
-JDK's [KEM interface](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/javax/crypto/KEM.html) was backported to JDK17, but not to earlier JDK versions. ACCP JARs support all LTS JDK versions since JDK8, so **ACCP's default build configuration and maven artifacts** omit ML-KEM as it relies on the KEM interface. To enable ML-KEM in locally built JARs, you'll need to build with a JDK ≥17 and specify `TARGET_JDK_VERSION` ≥17 like so:
+JDK's [KEM interface](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/javax/crypto/KEM.html) was backported to JDK 17 and is GA in JDK 21+, but is absent from other versions (JDK 8-16 and the non-LTS 18-20). ACCP always ships ML-KEM in its Multi-Release JAR (the `jdk17plus` overlay under `META-INF/versions/17`), so ML-KEM is available at runtime on JDK 17 or 21+; on other runtimes the base JAR loads without it. Because the overlay is always built, building ACCP requires a KEM-capable build JDK -- JDK 17 or 21+ (see [Build it yourself](#build-it-yourself)). `TARGET_JDK_VERSION` only selects the base bytecode level (default 8) and no longer gates ML-KEM.
```
geedo0, line 152: I love this tradeoff of accepting a slightly restrictive build, but enabling highly compatible artifacts.

geedo0 (approved):

PR #566, opened 2026-08-17
Validate native AES buffer ranges

geedo0:
*Description of changes:*
Track array lengths and direct-buffer capacities in JIOBlobs and add range-checked accessors for native pointers.

Validate input and output spans in the AES-XTS, AES-CFB, and AES-CBC native entry points. Account for CBC buffering and padding when determining the maximum writable output span.


WillChilds-Klein (approved):
do the added checks incur a meaningful performance hit? my guess is "no".

PR #568, opened 2026-08-19
Native ML-KEM SPKI encode/decode for regular FIPS (follow-up to #564)

WillChilds-Klein:
Follow-up to #564, and a **precursor to #569**, which adds ML-KEM seed-format *parsing* in regular FIPS. Makes ML-KEM key serialization work in **regular FIPS** so ML-KEM-over-TLS (BCJSSE) and `EvpKeyFactoryTest` no longer have to be skipped.

## Problem
AWS-LC-FIPS-3.1.0's ML-KEM `EVP_PKEY_ASN1_METHOD` implements no `pub_encode`/`pub_decode` or `priv_encode`/`priv_decode`, so `i2d_PUBKEY`/`d2i_PUBKEY` and `EVP_PKEY2PKCS8`/`d2i_PrivateKey` return `UNSUPPORTED_ALGORITHM` for ML-KEM. ML-KEM *KEM ops* work in FIPS, but `getEncoded()`/`KeyFactory` on ML-KEM keys fail -- blocking ML-KEM-over-TLS in regular FIPS (BCJSSE cannot encode the ephemeral public key into the TLS `key_share`) and failing `EvpKeyFactoryTest`'s ML-KEM cases. #564 skipped those in regular FIPS as an interim.

## Change

Hand-roll ML-KEM key encode/decode in ACCP's native layer using the raw-key APIs AWS-LC-FIPS-3.1.0 *does* provide (all in the FIPS module), bypassing the unsupported ASN.1 marshalers. See source comments with links to RFC for format details. Non-FIPS behavior is unchanged: the hand-rolled encoders are only reached when the library marshaler (`i2d_PUBKEY` for public keys, `EVP_marshal_private_key` for private keys) fails.

## #564 skips lifted
- BCJSSE ML-KEM integration test (regular FIPS)
- `EvpKeyFactoryTest` ML-KEM cases (regular FIPS)

## Review fixes
- **Library encoders take precedence.** `encodePublicKey` now tries `i2d_PUBKEY` first and hand-rolls only on failure, matching the private-key path; previously the hand-rolled encoder ran first for ML-KEM in *every* build, so non-FIPS output came from ACCP rather than AWS-LC. The ordering is also what makes the length-based OID inference in the fallback sound, since `KEM_find_kem_by_nid` resolves the legacy Kyber-R3 NIDs under the same `EVP_PKEY_KEM` type at identical key lengths.
- **Full `OneAsymmetricKey` grammar accepted.** The regular-FIPS PKCS#8 fallover parser took only version 0 and required the SEQUENCE to end right after `privateKey`. It now accepts version 0 or 1, an optional `[0] attributes`, and (version 1 only) an optional `[1] publicKey`, in both the constructed and the primitive tag encodings that `EVP_parse_private_key` accepts. Keys in those shapes decoded in non-FIPS but were rejected in regular FIPS. Trailing content is still rejected.
- **Encoder cleanup and error scoping.** A failed `CBB_finish` does not write `*out` and leaves the CBB owning its buffer, so the old `OPENSSL_free(*out)` freed an uninitialized pointer and leaked; the failure paths now `CBB_cleanup` instead. Added the missing `OPENSSL_malloc` and `CBB_init` return checks, and replaced the bare `ERR_clear_error()` calls with a scoped `ErrorQueueMark` guard so an expected fallover failure no longer discards errors queued by the caller, and a genuinely failed fallover reports the underlying error instead of nothing.
- **Misleading decode errors fixed.** `d2i_PrivateKey` leaves its input pointer untouched on failure, so the trailing-data check running before the null check reported every decode failure as `Extra key information`.
- **Tests.** `OneAsymmetricKey` shapes accepted; malformed PKCS#8 and SPKI encodings rejected (unassigned algorithm OID, bad version, wrong-length `expandedKey`, non-zero BIT STRING padding, empty/truncated public key, trailing bytes); ACCP's SPKI output cross-checked against BouncyCastle in all build modes.

## Still to come (follow-up PR)
This PR does **not** parse seed-format ML-KEM private keys in regular FIPS; #569 adds that (via `EVP_PKEY_keygen_deterministic`) and lifts `testKeyFactorySelfConversion`. Seed-format *encoding* (emitting the compact 86-byte seed) remains impossible in regular FIPS -- AWS-LC-FIPS-3.1.0 stores no keygen seed -- and awaits the AWS-LC-FIPS v5.0.0 bump. The seed-format tests stay skipped in this PR.

### DIFF
--- a/csrc/java_evp_keys.cpp
+++ b/csrc/java_evp_keys.cpp
@@ -38,8 +38,11 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpKey_enc
         EVP_PKEY* key = reinterpret_cast<EVP_PKEY*>(keyHandle);
         OPENSSL_buffer_auto der;
 
-        // This next line allocates memory
-        int derLen = i2d_PUBKEY(key, &der);
+        // This next line allocates memory. AWS-LC-FIPS 3.1.0 has no i2d_PUBKEY for ML-KEM, so
+        // hand-roll its SubjectPublicKeyInfo (byte-identical to i2d_PUBKEY in non-FIPS builds).
+        // TODO [AWS-LC-FIPS 4.x]: drop the encodeMLKEMPublicKey fallback once the FIPS module has i2d_PUBKEY.
+        int derLen = (EVP_PKEY_id(key) == EVP_PKEY_KEM) ? static_cast<int>(encodeMLKEMPublicKey(key, &der))
+                                                        : i2d_PUBKEY(key, &der);
         CHECK_OPENSSL(derLen > 0);
         if (!(result = env->NewByteArray(derLen))) {
             throw_java_ex(EX_OOM, "Unable to allocate DER array");
@@ -733,6 +736,56 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpMlDsaPr
     return result;
 }
 #endif // !defined(FIPS_BUILD) || defined(EXPERIMENTAL_FIPS_BUILD)
+
+/*
+ * Class:     com_amazon_corretto_crypto_provider_EvpKemPrivateKey
+ * Method:    encodeMlKemPrivateKey
+ * Signature: (J)[B
+ */
+JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpKemPrivateKey_encodeMlKemPrivateKey(
+    JNIEnv* pEnv, jclass, jlong keyHandle)
+{
+    jbyteArray result = NULL;
+
+    try {
+        raii_env env(pEnv);
+
+        EVP_PKEY* key = reinterpret_cast<EVP_PKEY*>(keyHandle);
+        CHECK_OPENSSL(EVP_PKEY_id(key) == EVP_PKEY_KEM);
+
+        uint8_t* der;
+        size_t der_len;
+        CBB cbb;
+        CHECK_OPENSSL(CBB_init(&cbb, 0));
+        // Prefer the standard PKCS8 marshal, which yields the compact seed format in non-FIPS and
+        // experimental-FIPS builds. AWS-LC-FIPS 3.1.0 cannot marshal ML-KEM private keys (no
+        // priv_encode) and lacks seed support, so on failure fall back to hand-rolled expanded-format
+        // PKCS8 after clearing the error queue. Mirrors encodeMlDsaPrivateKey above.
+        // TODO [AWS-LC-FIPS 4.x]: drop the encodeExpandedMLKEMPrivateKey fallback once the FIPS module has priv_encode.
+        if (EVP_marshal_private_key(&cbb, key)) {
+            if (!CBB_finish(&cbb, &der, &der_len)) {
+                OPENSSL_free(der);
+                throw_java_ex(EX_RUNTIME_CRYPTO, "Error finalizing seed ML-KEM key");
+            }
+        } else {
+            ERR_clear_error();
+            der_len = encodeExpandedMLKEMPrivateKey(key, &der);
+        }
+        CBB_cleanup(&cbb);
+
+        if (!(result = env->NewByteArray(der_len))) {
+            OPENSSL_free(der);
+            throw_java_ex(EX_OOM, "Unable to allocate DER array");
+        }
+        // This may throw, if it does we'll just keep the exception state as we return.
+        env->SetByteArrayRegion(result, 0, der_len, (const jbyte*)der);
+        OPENSSL_free(der);
+    } catch (java_ex& ex) {
+        ex.throw_to_java(pEnv);
+    }
+
+    return result;
+}
 
 /*
  * Class:     com_amazon_corretto_crypto_provider_EvpKemKey
--- a/csrc/java_evp_keys.cpp
+++ b/csrc/java_evp_keys.cpp
@@ -38,11 +38,25 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpKey_enc
         EVP_PKEY* key = reinterpret_cast<EVP_PKEY*>(keyHandle);
         OPENSSL_buffer_auto der;
 
-        // This next line allocates memory. AWS-LC-FIPS 3.1.0 has no i2d_PUBKEY for ML-KEM, so
-        // hand-roll its SubjectPublicKeyInfo (byte-identical to i2d_PUBKEY in non-FIPS builds).
-        // TODO [AWS-LC-FIPS 4.x]: drop the encodeMLKEMPublicKey fallback once the FIPS module has i2d_PUBKEY.
-        int derLen = (EVP_PKEY_id(key) == EVP_PKEY_KEM) ? static_cast<int>(encodeMLKEMPublicKey(key, &der))
-                                                        : i2d_PUBKEY(key, &der);
+        // This next line allocates memory. On failure i2d_PUBKEY leaves |der| untouched.
+        int derLen = i2d_PUBKEY(key, &der);
+        if (derLen <= 0 && EVP_PKEY_id(key) == EVP_PKEY_KEM) {
+            // AWS-LC-FIPS 3.1.0 has no pub_encode for ML-KEM, so i2d_PUBKEY raises
+            // UNSUPPORTED_ALGORITHM there; hand-roll the SubjectPublicKeyInfo instead. The
+            // hand-rolled encoding is byte-identical to what i2d_PUBKEY produces in non-FIPS builds.
+            //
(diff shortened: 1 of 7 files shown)


csrc/keyutils.cpp
```diff
@@ -98,8 +119,141 @@ EVP_PKEY* der2EvpPrivateKey(const unsigned char* der,
...
+// d2i_PrivateKey.
+static EVP_PKEY* parseMLKEMPrivateKey(const unsigned char* der, const int derLen)
+{
+    CBS cbs, pkcs8, algorithm, oid, priv, expanded;
+    uint64_t version = 0;
+    CBS_init(&cbs, der, derLen);
+    if (!CBS_get_asn1(&cbs, &pkcs8, CBS_ASN1_SEQUENCE) || CBS_len(&cbs) != 0
+        || !CBS_get_asn1_uint64(&pkcs8, &version) || version != 0
+        || !CBS_get_asn1(&pkcs8, &algorithm, CBS_ASN1_SEQUENCE)
+        || !CBS_get_asn1(&algorithm, &oid, CBS_ASN1_OBJECT)) {
```
WillChilds-Klein: here and in other parse/encode CBS/CBB blocks, please format with one `|| ` clause per line

csrc/keyutils.cpp
```diff
@@ -98,8 +119,141 @@ EVP_PKEY* der2EvpPrivateKey(const unsigned char* der,
...
+    }
+    // EVP_PKEY_kem_new_raw_secret_key validates that the length matches the parameter set's
+    // secret_key_len and returns NULL otherwise (falling through to d2i_PrivateKey).
+    return EVP_PKEY_kem_new_raw_secret_key(nid, CBS_data(&expanded), CBS_len(&expanded));
+}
+
 EVP_PKEY* der2EvpPublicKey(const unsigned char* der, const int derLen, const char* javaExceptionClass)
 {
+    // AWS-LC-FIPS 3.1.0 cannot d2i_PUBKEY ML-KEM keys; parse ML-KEM SPKI via the raw public key.
+    // Non-ML-KEM inputs return nullptr here and fall through to d2i_PUBKEY below.
```
WillChilds-Klein: here and elsewhere mentioning AWS-LC-FIPS 3.1.0 drop a TODO to drop these when we bump to 4.x

csrc/keyutils.cpp
```diff
@@ -98,8 +119,141 @@ EVP_PKEY* der2EvpPrivateKey(const unsigned char* der,
...
+    CHECK_OPENSSL(out);
+    size_t raw_len = 0;
+    CHECK_OPENSSL(EVP_PKEY_get_raw_public_key(key, nullptr, &raw_len));
+    int nid = mlkemNidForPublicKeyLen(raw_len);
+    if (nid == NID_undef) {
+        throw_java_ex(EX_ILLEGAL_ARGUMENT, "Invalid ML-KEM public key size");
+    }
+    OPENSSL_buffer_auto raw_pub(raw_len);
+    CHECK_OPENSSL(EVP_PKEY_get_raw_public_key(key, raw_pub, &raw_len));
+    // SubjectPublicKeyInfo ::= SEQUENCE { algorithm SEQUENCE { OID }, subjectPublicKey BIT STRING }.
```
WillChilds-Klein, line 188: link to RFC

csrc/keyutils.cpp
WillChilds-Klein: or rather, one get/put per line. inline len checks ok

WillChilds-Klein: Done in 7158509. Wrapped the CBS parse blocks in `// spotless:off` (matching the CBB encode blocks) and reformatted to one get/put per line with length checks inline. Clause order is preserved -- the `CBS_get_*` calls advance the cursor and `||` short-circuits, so the order is significant and left unchanged; only line breaks moved. Applies to both conditionals in `parseMLKEMPublicKey` and `parseMLKEMPrivateKey`.

csrc/keyutils.cpp
WillChilds-Klein: Done in 7158509. Added `// TODO [AWS-LC-FIPS 4.x]: ...` markers at every AWS-LC-FIPS 3.1.0 workaround site -- 8 total, greppable via `TODO [AWS-LC-FIPS 4.x]`: `keyutils.cpp` (der2EvpPrivateKey, the SPKI section header, der2EvpPublicKey, encodeExpandedMLKEMPrivateKey), `java_evp_keys.cpp` (both encode fallbacks), and `keyutils.h` (both decls).

csrc/keyutils.cpp
WillChilds-Klein: Done in 7158509. Added an RFC 9935 Section 4 (Subject Public Key Fields) link at the SubjectPublicKeyInfo structure comment. The section header also cites Section 4 + Appendix A (ASN.1 module); the private-key paths cite Section 6 + Appendix A.

csrc/keyutils.cpp
```diff
@@ -98,8 +120,154 @@ EVP_PKEY* der2EvpPrivateKey(const unsigned char* der,
...
+    // secret_key_len and returns NULL otherwise (falling through to d2i_PrivateKey).
+    return EVP_PKEY_kem_new_raw_secret_key(nid, CBS_data(&expanded), CBS_len(&expanded));
+}
+
 EVP_PKEY* der2EvpPublicKey(const unsigned char* der, const int derLen, const char* javaExceptionClass)
 {
+    // AWS-LC-FIPS 3.1.0 cannot d2i_PUBKEY ML-KEM keys; parse ML-KEM SPKI via the raw public key.
+    // Non-ML-KEM inputs return nullptr here and fall through to d2i_PUBKEY below.
+    // TODO [AWS-LC-FIPS 4.x]: drop this ML-KEM SPKI parse once the FIPS module supports d2i_PUBKEY for ML-KEM.
+    EVP_PKEY* mlkem = parseMLKEMPublicKey(der, derLen);
```
WillChilds-Klein: in this implementation, we try to parse ML-KEM expanded keys for _all_ public keys, incurring a performance hit for non-ML-KEM key types. on the one hand, overhead is bounded to 3 `CBS_get_asn1` calls + 1 `OBJ_cbs2nid` call. on the other hand, it's unnecessary. can we refactor this to try normal `d2i_PUBKEY` parsing first, then fail over to `parseMLKEMPublicKey`?

same for `d2i_PrivateKey` above.

WillChilds-Klein: Done in bd6d7ff. Flipped both `der2EvpPublicKey` and `der2EvpPrivateKey` to try the standard `d2i_PUBKEY`/`d2i_PrivateKey` first and only fall over to the hand-rolled ML-KEM parser when d2i cannot decode the key. Now non-ML-KEM keys (and, in non-FIPS builds, ML-KEM keys, which d2i handles natively) pay zero ML-KEM parsing cost; the hand-rolled path runs only in regular FIPS where d2i_* raises UNSUPPORTED_ALGORITHM on ML-KEM.

Notes:
- I clear the OpenSSL error queue (`ERR_clear_error()`) after the failed d2i before the fallover, since d2i_* pushes UNSUPPORTED_ALGORITHM for ML-KEM in FIPS.
- The private path keeps its `evpType == EVP_PKEY_KEM` guard on the fallover, so non-KEM key decoding (RSA/EC) is byte-for-byte unchanged.
- "Extra key information" strictness on trailing bytes is preserved for the d2i-success case.
- Same change flows into #569 (rebased on top).

csrc/java_evp_keys.cpp
```diff
@@ -731,6 +756,62 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_provider_EvpMlDsaPr
...
+        size_t der_len = 0;
+        CBB cbb;
+        CHECK_OPENSSL(CBB_init(&cbb, 0));
+        // Prefer the standard PKCS8 marshal, which yields the compact seed format in non-FIPS and
+        // experimental-FIPS builds. AWS-LC-FIPS 3.1.0 cannot marshal ML-KEM private keys (no
+        // priv_encode) and lacks seed support, so on failure fall back to hand-rolled expanded-format
+        // PKCS8. Preferring the library encoder also keeps every non-ML-KEM EVP_PKEY_KEM key off
+        // encodeExpandedMLKEMPrivateKey's length-based OID inference; see the equivalent note in
+        // encodePublicKey above. Mirrors encodeMlDsaPrivateKey above.
+        // TODO [AWS-LC-FIPS 4.x]: drop the encodeExpandedMLKEMPrivateKey fallback once the FIPS module has priv_encode.
```
justsmth, line 785: NP: Mention seed encode needs FIPS 5.0?

justsmth (approved):

jakemas (approved):

csrc/java_evp_keys.cpp
WillChilds-Klein: will address in follow-up

PR #569, opened 2026-08-19
Parse seed-format ML-KEM private keys in regular FIPS (follow-up to #568)

WillChilds-Klein:
Follow-up to #568. Adds ML-KEM **seed-format private-key parsing** in regular FIPS.

## Context

AWS-LC-FIPS 3.1.0's ML-KEM ASN.1 method implements no `priv_decode`/`priv_encode`. #568 hand-rolled the X.509 SubjectPublicKeyInfo and the expanded-format PKCS#8 for regular FIPS; it did not handle the seed CHOICE of [RFC 9935 Section 6](https://datatracker.ietf.org/doc/html/rfc9935#section-6).

## Change

`parseMLKEMPrivateKey` now peeks the privateKey CHOICE and handles both arms:

- **expandedKey** `OCTET STRING` -> `EVP_PKEY_kem_new_raw_secret_key` (public key left unpopulated, as in AWS-LC)
- **seed** `[0] IMPLICIT OCTET STRING` -> `EVP_PKEY_keygen_deterministic`, expanding the raw `d || z` into a full key pair

The parser itself is compiled in every build so its grammar cannot drift per flavor, but it only runs as a fallover after `d2i_PrivateKey` fails, so in non-FIPS and experimental FIPS it is effectively dead; in regular FIPS it is the entire ML-KEM decoder. The seed arm's expansion helper, and with it the `<openssl/experimental/kem_deterministic_api.h>` include that has no API-stability guarantee, are compiled only for regular FIPS -- the one flavor that can reach them -- and stubbed out elsewhere.

`MlKemUtils.expandPrivateKey` now routes through `der2EvpPrivateKey` and works in regular FIPS too.

`build.gradle` pins `awsLcMainTag` to `v1.72.0`. AWS-LC `be77a8ad3` (#3159) raised the CMake Go floor to 1.20 with a `FATAL_ERROR`, first released in v1.72.1; our Ubuntu CI image ships Go 1.18 and is difficult to change, so v1.72.0 is the highest mainline tag that still builds there. This costs no ML-KEM capability: v1.72.0's `kem_key_st` already carries a seed field, `pkey_kem_keygen` retains it, and `kem_priv_encode` re-emits the seed CHOICE.

## Behavior changes to `MlKemUtils.expandPrivateKey`

It previously short-circuited on the input length, handing anything already expanded straight back. It now parses and re-encodes unconditionally, which is what lets it work under FIPS, and that has three consequences worth calling out:

- **Inputs outside the grammar are rejected rather than echoed.** The `both` CHOICE (`SEQUENCE { seed, expandedKey }`) is the case that matters in practice, because it is what BouncyCastle's ML-KEM `getEncoded()` emits by default, and BC's `getAlgorithm()` is the parameter-set name so the method's ML-KEM check passes. Pinned by `testBothChoiceRejected`.
- **The rejection is unchecked.** `der2EvpPrivateKey` is now given `EX_RUNTIME_CRYPTO` rather than `EX_INVALID_KEY_SPEC`: `InvalidKeySpecException` is checked, and JNI would otherwise have thrown one out of a method whose signature does not declare it, where no caller can legally catch it. `@throws` tags added to the javadoc.
- **The output is always the canonical minimal form** -- version 0 `PrivateKeyInfo`, no `attributes`, no `publicKey`. Those optional `OneAsymmetricKey` fields are still accepted on input, but they are not carried over, so the method is idempotent on its own output rather than byte-preserving on arbitrary input. Pinned by `testExpandPrivateKeyNormalizesEncoding`.

## Parser strictness

Two places where the hand-rolled code is deliberately not just a transcription of AWS-LC:

- **Trailing bytes after the CHOICE, inside the privateKey `OCTET STRING`, are rejected.** `kem_priv_decode` extracts the CHOICE and never looks at what follows it, in every AWS-LC version. Ignoring them would let unbounded caller-controlled data ride along inside an encoding ACCP declares well-formed, and would make two distinct DER encodings decode to the same key so that `getEncoded()` no longer round-trips. Nothing any AWS-LC encoder emits has such trailing bytes, so this cannot refuse a key the library produced. Pinned build-conditionally in `testMalformedPkcs8PrivateKeysRejected`.
- **`encodeExpandedMLKEMPrivateKey` cross-checks the secret-key length against the public-key length** instead of inferring the ML-KEM OID from the secret-key length alone. Legacy Kyber-R3 sits under the same `EVP_PKEY_KEM` type with byte-for-byte identical key lengths, and unlike `encodeMLKEMPublicKey` this function has no caller-side guarantee that the library refused to encode the key -- `MlKemUtils.expandPrivateKey` reaches it in every build. Requiring both lengths to name the same parameter set refuses a KEM key of another family rather than relabeling it with an ML-KEM OID.

## Parser fidelity

The parser otherwise mirrors AWS-LC **v5.0.0** -- the decoder it is destined to be deleted in favor of -- not the v1.72.0 now pinned, so removing it after the AWS-LC-FIPS bump is as close to a no-op as it can be. It therefore rejects the `both` CHOICE (AWS-LC: "Case 3 ... not implemented yet") and spells attributes as constructed `[0]` and publicKey as primitive `[1]`.

v1.72.0 differs on that last point, and **neither grammar contains the other**, so the flavors are not strictly ordered: v1.72.0 accepts a constructed `[1]` where a primitive `[1]` publicKey belongs (it matches neither of its tags and never checks for trailing data inside the SEQUENCE), while this parser accepts a version 1 key carrying both a constructed `[0]` attributes and a primitive `[1]` publicKey, which v1.72.0 rejects. What makes that safe is not a containment relation but the fallover structure: outside regular FIPS a key is accepted if *either* decoder accepts it, so this parser can only ever widen what those builds accept, and in regular FIPS it is the only decoder. The divergence disappears once the pinned mainline tag reaches v5.x. The two input classes where the answer is version-dependent are asserted build-conditionally, requiring only that the bogus field is never absorbed into the key.

## Known difference

Regular FIPS cannot *serialize* the seed CHOICE: AWS-LC-FIPS 3.1.0's `KEM_KEY` has no seed field, so `getEncoded()` emits `expandedKey` where the other builds emit `seed`. Both CHOICEs are *parsed* everywhere, so a seed-format key imported via `KeyFactory` works in every build; it just re-encodes to the expanded form under FIPS. This is the only intended FIPS-vs-non-FIPS difference in ML-KEM key handling for well-formed keys, it is documented in `README.md`, and it goes away with the AWS-LC-FIPS v5.0.0 bump. Experimental FIPS tracks mainline AWS-LC and is unaffected.

<details>
<summary><b>ACCP / BouncyCastle ML-KEM compatibility matrix</b> (parse and serialize, per build flavor)</summary>

Only one of the three ACCP builds differs from the others for well-formed keys, and only in what it *emits* -- parsing of well-formed keys is identical across all three.

### A. Private key PKCS#8 -- parse (RFC 9935 Section 6 CHOICEs)

| Encoding | ACCP non-FIPS | ACCP exp-FIPS | ACCP regular FIPS | BC 1.85 |
|---|---|---|---|---|
| `seed [0]` (0x80, 86 bytes) | accept | accept | accept | accept |
| `expandedKey` (0x04) | accept | accept | accept | accept |
| `both SEQUENCE` (0x30) | **reject** | **reject** | **reject** | accept |
| CHOICE + trailing bytes inside the privateKey `OCTET STRING` | ignored by AWS-LC | ignored by AWS-LC | **reject** | -- |

ACCP's acceptance set for well-formed keys is uniform by design. In non-FIPS/exp-FIPS it comes from AWS-LC's `kem_priv_decode`; in regular FIPS from `parseMLKEMPrivateKey`. The `both` rejection is not a FIPS artifact -- no AWS-LC flavor implements Case 3 ("not implemented yet"), so ACCP rejects it everywhere. BC accepts it because BC is the thing that produces it. The last row is the one place ACCP is stricter than the library it wraps, and it only bites in regular FIPS, where ACCP's parser is the whole decoder.

Shell tolerance is otherwise uniform: RFC 5208 version 0 and RFC 5958 version 1, with optional `[0]` attributes and `[1]` publicKey, all accepted in every build.

### B. Private key PKCS#8 -- serialize

| Producer | `getEncoded()` of a generated key | re-encode of a seed-imported key | re-encode of an expanded-imported key |
|---|---|---|---|
| ACCP non-FIPS | `seed` (86 B) | `seed`, byte-identical | `expandedKey`, byte-identical |
| ACCP exp-FIPS | `seed` (86 B) | `seed`, byte-identical | `expandedKey`, byte-identical |
| ACCP regular FIPS | **`expandedKey`** | **widened to `expandedKey`** | `expandedKey`, byte-identical |
| BC 1.85 | **`both`** | `seed`, byte-identical | `expandedKey`, byte-identical |

`MlKemUtils.expandPrivateKey` yields the canonical minimal `expandedKey` PKCS#8 in all three builds and is idempotent on that form. It is not byte-preserving on arbitrary input: optional `OneAsymmetricKey` fields are accepted and dropped, and encodings it cannot parse raise `RuntimeCryptoException` rather than being handed back.

### C. End-to-end interop

| Direction | non-FIPS | exp-FIPS | regular FIPS |
|---|---|---|---|
| ACCP `getEncoded()` -> BC | works, BC re-emits byte-identically | same | works, BC re-emits byte-identically |
| **BC `getEncoded()` -> ACCP** | **fails** (`InvalidKeySpecException`) | **fails** | **fails** |
| **BC `getEncoded()` -> `MlKemUtils.expandPrivateKey`** | **fails** (`RuntimeCryptoException`) | **fails** | **fails** |
| BC `getPrivateKey(true)` (seed) -> ACCP | works | works | works |
| BC `getPrivateKey(false)` (expanded) -> ACCP | works | works | works |
| Public key SPKI, either direction | byte-identical | byte-identical | byte-identical |

Public keys are a non-issue: one canonical SPKI form, byte-identical round-trip both ways in every build, no FIPS guard on the test.

### D. Two gotchas that are not encoding-format issues

**1. BC's default private-key encoding is unreadable by ACCP, in every build.** `bcPriv.getEncoded()` is the `both` SEQUENCE. There is no BC system property in `bcprov-jdk18on-1.85` to change the default -- the only lever is `MLKEMPrivateKey.getPrivateKey(true|false).getEncoded()`. Anyone feeding BC-generated ML-KEM private keys to ACCP has to call that explicitly, including when calling `MlKemUtils.expandPrivateKey`, which rejects the `both` form rather than echoing it. This is the sharpest edge in the matrix, and FIPS mode has nothing to do with it.

**2. Public-key recovery from an `expandedKey`-only import differs between the two providers.**

| | ACCP (all builds) | BC 1.85 |
|---|---|---|
| public key after `seed` import | recovered, matches | recovered, matches |
| public key after `expandedKey` import | **throws `RuntimeCryptoException`** | recovered, matches |
| `getSeed()` after `expandedKey` import | n/a | absent |

BC recovers it because FIPS 203 embeds the encapsulation key inside `dk` (`dk_PKE || ek || H(ek) || z`); AWS-LC's `EVP_PKEY_kem_new_raw_secret_key` chooses not to populate `pub`, and ACCP inherits that. So a key imported into ACCP from `expandedKey` PKCS#8 is decapsulate-only -- carry the SPKI separately if the public half is needed. Round-tripping the same key through BC would appear to "fix" it, which makes this an easy trap. The BC column here was verified directly against BC 1.85; only the ACCP column is pinned by a test.

### E. What changes at the AWS-LC-FIPS milestones

Only two cells above are mutable, and both are TODO-tagged in the source:

| Cell | Milestone | Tagged at |
|---|---|---|
| Regular FIPS widens `seed` -> `expandedKey` on re-encode (B row 3), and cannot emit `seed` at all | AWS-LC-FIPS 5.0 (module must retain the keygen seed) | `encodeMlKemPrivateKey` in `java_evp_keys.cpp`; the two guards in `testPrivateKeyEncodingIsSeedFormat` / `testPrivateKeyChoicesParse` |
| Regular FIPS answers with `parseMLKEMPrivateKey` rather than AWS-LC's decoder (A column 3, including the trailing-bytes row) | AWS-LC-FIPS 4.x (module implements `priv_decode` for ML-KEM) | `der2EvpPrivateKey`, `mlkemKeyFromSeed`, the experimental include, and the seed branch in `keyutils.cpp` |

After 4.x the parse column is AWS-LC's answer in all three builds and the hand-rolled parser -- along with the experimental-header include and the two build-conditional assertions -- is deleted; after 5.0 the regular-FIPS serialize row becomes identical to the other two, and every build/provider difference above except BC's `both` default disappears.

Every cell in A, B, and C is pinned by a currently-passing test: `testPrivateKeyChoicesParse`, `testBothChoiceRejected`, `testMalformedPkcs8PrivateKeysRejected`, `testExpandPrivateKeyNormalizesEncoding`, `testImportedPrivateKeyPublicKeyAvailability`, `testPublicKeyEncodingMatchesBouncyCastle`, `testPrivateKeyEncodingIsSeedFormat`, `testBouncyCastleInteroperability`. `MlKemTest` is 60 tests: 60 pass in non-FIPS and in experimental FIPS, and 57 pass / 3 aborted in regular FIPS. Those 3 aborts are all of `testPrivateKeyEncodingIsSeedFormat`, which is why "regular FIPS cannot emit seed" is the sole documented divergence for well-formed keys.

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ ACCP did not track a FIPS branch/release version of AWS-LC until ACCP v2.3.0. Be
 | 2.4.0               | 1.30.1         | 2.0.13              |
 | 2.4.1               | 1.30.1         | 2.0.13              |
 | 2.5.0               | 1.47.0         | 3.0.0               |
-| 2.6.0               | 1.72.0         | 3.1.0               |
+| 2.6.0               | 5.0.0          | 3.1.0               |
 
 Notable differences between ACCP and ACCP-FIPS:
 * ACCP uses [the latest release of AWS-LC](https://github.com/aws/aws-lc/releases), whereas, ACCP-FIPS uses [a timestamped FIPS branch of AWS-LC](https://github.com/aws/aws-lc/branches).
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ ACCP did not track a FIPS branch/release version of AWS-LC until ACCP v2.3.0. Be
 | 2.4.0               | 1.30.1         | 2.0.13              |
 | 2.4.1               | 1.30.1         | 2.0.13              |
 | 2.5.0               | 1.47.0         | 3.0.0               |
-| 2.6.0               | 5.0.0          | 3.1.0               |
+| 2.6.0               | 1.72.0         | 3.1.0               |
 
 Notable differences between ACCP and ACCP-FIPS:
 * ACCP uses [the latest release of AWS-LC](https://github.com/aws/aws-lc/releases), whereas, ACCP-FIPS uses [a timestamped FIPS branch of AWS-LC](https://github.com/aws/aws-lc/branches).
@@ -187,6 +187,7 @@ Notable differences between ACCP and ACCP-FIPS:
 * In FIPS-mode, RSA keys are limited to 2048, 3072, or 4096 bits in size with public exponent F4.
 * Due to the fact that an older branch of AWS-LC is used in FIPS-mode, there will be performance differences between ACCP and ACCP-FIPS. We highly recommend performing detailed performance testing of your application if you choose to experiment with ACCP-FIPS.
 * Between versions 2.1.0 and 2.3.3 (inclusive), ACCP-FIPS does not register SecureRandom by default due to the performance of AWS-LC’s entropy source in FIPS-mode, with older versions of AWS-LC. Since version 2.4.0, ACCP-FIPS behaves as ACCP: it registers SecureRandom from AWS-LC by default. [A system property](https://github.com/corretto/amazon-corretto-crypto-provider#other-system-properties) is available to change the default behavior.
+* ACCP-FIPS cannot *serialize* an ML-KEM private key in the seed format of [RFC 9935](https://datatracker.ietf.org/doc/html/rfc9935#section-6): the `KEM_KEY` of AWS-LC-FIPS 3.1.0 has no field in which to retain the keygen seed, so `PrivateKey.getEncoded()` emits the `expandedKey` CHOICE where ACCP emits `seed`. Both CHOICEs are still *parsed* in every build, so a seed-format key imported through `KeyFactory` works everywhere; it simply re-encodes to the expanded form under FIPS. For well-formed keys this is the only intended behavioral difference in ML-KEM key handling between the FIPS and non-FIPS builds, and it goes away when ACCP-FIPS moves to an AWS-LC-FIPS branch whose `KEM_KEY` retains the seed. Builds configured with `-DEXPERIMENTAL_FIPS=1` track the non-FIPS AWS-LC branch and so are unaffected.
 
 ACCP-FIPS is only supported on the following platforms:
 
--- a/csrc/keyutils.cpp
+++ b/csrc/keyutils.cpp
@@ -7,15 +7,15 @@
 #include <openssl/ec.h>
 #include <openssl/err.h>
 #include <openssl/evp.h>
+// EVP_PKEY_keygen_deterministic (seed -> ML-KEM key) lives in this experimental header. Every AWS-LC
+// flavor ACCP builds against declares it with the same signature, so the seed decode below is compiled
+// unconditionally rather than only for regular FIPS: one parser for all flavors cannot drift per build,
+// and the grammar it accepts stays a single thing to reason about.
+// TODO [AWS-LC-FIPS 4.x]: drop this experimental include once the FIPS module can d2i seed-format ML-KEM keys.
+#include <openssl/experimental/kem_deterministic_api.h>
 #include <openssl/obj.h>
 #include <openssl/pkcs8.h>
 #include <openssl/rsa.h>
-#if defined(FIPS_BUILD) && !defined(EXPERIMENTAL_FIPS_BUILD)
-// EVP_PKEY_keygen_deterministic (seed -> ML-KEM key) lives in this experimental header. Only regular
-// FIPS needs it: AWS-LC-FIPS 3.1.0's ASN.1 method cannot decode seed-format ML-KEM private keys.
-// TODO [AWS-LC-FIPS 4.x]: drop this experimental include once the FIPS module can d2i seed-format ML-KEM keys.
-#include <openssl/experimental/kem_deterministic_api.h>
-#endif
 
 namespace AmazonCorrettoCryptoProvider {
 
@@ -43,12 +43,15 @@ EVP_PKEY* der2EvpPrivateKey(const unsigned char* der,
     // no ML-KEM parsing overhead.
     EVP_PKEY* result = d2i_PrivateKey(evpType, NULL, &der_mutable_ptr, derLen);
 
-    // Regular-FIPS fallover: AWS-LC-FIPS 3.1.0 has no priv_decode for ML-KEM, so d2i_PrivateKey fails
-    // on ML-KEM PKCS8. Only when the caller asked for a KEM key and d2i_PrivateKey produced nothing do
-    // we hand-roll the parse via the raw secret key. This matches the non-FIPS d2i expanded-decode
-    // path exactly: both reconstruct the key by setting only the raw secret key
-    // (KEM_KEY_set_raw_secret_key), leaving the public key unpopulated. parseMLKEMPrivateKey returns
-    // nullptr for inputs it cannot handle.
+    // Fallover for ML-KEM PKCS8: AWS-LC-FIPS 3.1.0 has no priv_decode for ML-KEM, so d2i_PrivateKey
+    // fails on every ML-KEM private key there. Only when the caller asked for a KEM key and
+    // d2i_PrivateKey produced nothing do we hand-roll the parse via the raw APIs. Each CHOICE
+    // reconstructs the key exactly the way mainline AWS-LC's decoder does, so the two agree per CHOICE:
+    //   - expandedKey sets only the raw secret key (KEM_KEY_set_raw_secret_key), leaving the public key
+    //     unpopulated, so getPublicKey().getEncoded() fails for such a key in every build;
+    //   - seed derives the whole key pair from the seed, public key included.
+    // parseMLKEMPrivateKey returns nullptr for inputs it cannot handle, which in a non-FIPS build is
+    // everything: it accepts nothing d2i_PrivateKey has not already accepted there.
     // TODO [AWS-LC-FIPS 4.x]: drop this raw-key decode path once the FIPS module supports d2i_PrivateKey for ML-KEM.
     if (result == nullptr && evpType == EVP_PKEY_KEM) {
         EVP_PKEY* mlkem = parseMLKEMPrivateKey(der, derLen);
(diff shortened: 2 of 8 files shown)


tst/com/amazon/corretto/crypto/provider/test/MlKemUtilsTest.java
```diff
@@ -80,15 +80,13 @@ private static boolean mlKemDisabled() {
...
-    // TODO: remove this guard once AWS-LC-FIPS is bumped to v5.0.0, which provides
-    // FIPS-validated ML-KEM private-key (seed-format) parsing.
+    // MlKemUtils.expandPrivateKey's native method (expandPrivateKeyInternal) is compiled only in
+    // non-FIPS / experimental-FIPS builds. Seed PARSING via KeyFactory is supported in regular FIPS
+    // (see MlKemTest#testSeedFormatPrivateKeyParses).
     assumeTrue(
         !AmazonCorrettoCryptoProvider.INSTANCE.isFips()
             || AmazonCorrettoCryptoProvider.INSTANCE.isExperimentalFips(),
-        "ML-KEM seed-format keys are unavailable in FIPS builds");
+        "MlKemUtils.expandPrivateKey is unavailable in regular FIPS");
```
WillChilds-Klein: this PR should unblock `MlKemUtils.expandPrivateKey` in regular FIPS

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/MlKemTest.java
```diff
@@ -256,14 +263,9 @@ public void testParameterSpecMismatch() throws Exception {
...
-    // underlying AWS-LC -- present only in non-FIPS / experimental-FIPS builds (matches the
-    // native #if !defined(FIPS_BUILD) || defined(EXPERIMENTAL_FIPS_BUILD) guard).
-    // TODO: remove this guard once AWS-LC-FIPS is bumped to v5.0.0, which provides
-    // FIPS-validated ML-KEM private-key (seed-format) parsing.
-    assumeTrue(
-        !NATIVE_PROVIDER.isFips() || NATIVE_PROVIDER.isExperimentalFips(),
-        "ML-KEM seed-format keys are unavailable in FIPS builds");
+    // Round-trips ACCP's own ML-KEM key pair through its KeyFactory. In regular FIPS the private
+    // key serializes in expanded format (seed format is unavailable there), so this exercises the
+    // expanded encode/decode; in other builds it exercises the seed format. Public keys use X.509.
```
WillChilds-Klein: with our changes to csrc/ we should always serialize seed format, in both regular FIPS and experimental FIPS. there should be no behavioral differences between ML-KEM key parse/serialization between regular/experimental FIPS.

also, in all the places you remove `assumeTrue(!NATIVE_PROVIDER.isFips() || NATIVE_PROVIDER.isExperimentalFips()`, please remove the added comment entirely.

csrc/keyutils.cpp
```diff
@@ -495,23 +587,17 @@ size_t encodeExpandedMLKEMPrivateKey(const EVP_PKEY* key, uint8_t** out)
...
-        nid = NID_MLKEM1024;
-        break;
-    default:
-        throw_java_ex(EX_ILLEGAL_ARGUMENT, "Invalid ML-KEM secret key size");
+    // Neither length identifies an ML-KEM parameter set on its own (see mlkemNidForPublicKeyLen), and
+    // unlike encodeMLKEMPublicKey this has no caller-side guarantee that the key is ML-KEM at all:
+    // MlKemUtils.expandPrivateKey reaches it in every build. Requiring both lengths to name the same
+    // parameter set refuses a KEM key of another family rather than relabeling it with an ML-KEM OID.
+    // Querying the public-key length is safe even for keys holding no public key, since
+    // kem_get_pub_raw reports the parameter set's length before it looks at the key's own public key.
```
justsmth: Nit: The `mlkemNidForPublicKeyLen` note says Kyber-R3 shares both public and secret key lengths with ML-KEM. For that known collision, these lookups still produce the same ML-KEM NID, so this does not provide the general "refuses a KEM key of another family" guarantee stated here or the Kyber-specific guarantee in the PR description. It only rejects a family whose secret length collides but whose public length does not. 

Suggestion: Narrow the wording accordingly?

src/com/amazon/corretto/crypto/utils/MlKemUtils.java
justsmth: `PrivateKey.getEncoded()` is permitted to return null. Passing that directly to the native method reaches `GetArrayLength(nullptr)`. This should store and validate the encoding before passing it further.

README.md
```diff
@@ -187,6 +187,7 @@ Notable differences between ACCP and ACCP-FIPS:
...
+* ACCP-FIPS cannot *serialize* an ML-KEM private key in the seed format of [RFC 9935](https://datatracker.ietf.org/doc/html/rfc9935#section-6): the `KEM_KEY` of AWS-LC-FIPS 3.1.0 has no field in which to retain the keygen seed, so `PrivateKey.getEncoded()` emits the `expandedKey` CHOICE where ACCP emits `seed`. Both CHOICEs are still *parsed* in every build, so a seed-format key imported through `KeyFactory` works everywhere; it simply re-encodes to the expanded form under FIPS. For well-formed keys this is the only intended behavioral difference in ML-KEM key handling between the FIPS and non-FIPS builds, and it goes away when ACCP-FIPS moves to an AWS-LC-FIPS branch whose `KEM_KEY` retains the seed. Builds configured with `-DEXPERIMENTAL_FIPS=1` track the non-FIPS AWS-LC branch and so are unaffected.
```
justsmth, line 190: Nit: The PR description explains that BouncyCastle's default `getEncoded()` uses the unsupported `both` CHOICE, but that information will be difficult for users to find after merge. 

Suggestion: Could we also add it to the ML-KEM Considerations section, including the required `getPrivateKey(true|false)` conversion?

csrc/keyutils.cpp
WillChilds-Klein: Yeah, good point. Kyber is fully deprecated so key length collisions can safely be ignored. will update.

README.md
WillChilds-Klein: We don't really talk about Bouncy Castle in the README (one instance is where we use a non-JCA standard algorithm name cribbed from BC). I'd prefer to keep that documentation closer to the BC interop tests where it's demonstrated.

WillChilds-Klein: Eventually, we'll support "both" encoding after @jakemas adds it to AWS-LC and we bump our tracking version accordingly. So, this discrepancy should go away ~soon.

justsmth (approved):

jakemas (approved):

Issue #560, opened 2026-07-31
Please add support for AES-CTR mode


sfc-gh-grubin:
I'd love to see AES-CTR exposed by ACCP so that we have a good FIPS validated implementation of it available in Java.

Thanks.

SalusaSecondus:
@WillChilds-Klein , I see that both CBC and CFB don't support the "Hybrid" modes of context reuse and I'm wondering why. (I remember that AES key scheduling can be expensive.)

For CTR I see a few options and would like your opinion:

1. Don't both saving the context at all. (Simple but takes a key schedule hit.)
2. Full GCM style with Lazy, Eager, and Hybrid.
3. CBC style (save only in Lazy but always init the key)
4. Save context *except* in Eager. But notice if the key is unchanged and don't re-init with the same key if the key is unchanged. (This is my preference.)

Thoughts?

WillChilds-Klein:
Hi @SalusaSecondus,

4. sounds like a reasonable balance between simplicity and performance.

SalusaSecondus:
Looks like we can close this now that #571 has merged. (Though, I guess we could wait for a new release.)

PR #571, opened 2026-08-20
Add support for AES/CTR/NoPadding

SalusaSecondus:
*Issue #:* #560 

*Description of changes:*

This change adds support for `AES/CTR/NoPadding` to ACCP. The code is based heavily on the existing implementation of `AES/CFB` (almost copy/paste with find/replace) with aspects also pulled from `CBC`.

I apologize for the large PR, but over half of it is just test code, so that's not too bad 😁 

## Changes

- New implementation of AES-CTR (Java and C++)
- Unit tests
  - There are no official KATs from NIST, so I couldn't do that
  - Strong dependency on compatibility tests against both SunJCE and BouncyCastle
  - Compatibility tests against new hand-built reference implementation using ECB
  - Specific tests to cover edge-cases in resource handling
- New benchmark code
- Update to README

## How I did native resource management

I manage resources pretty much the same that CBC does it (always preserve for LAZY, release for everything else) with the sole change that if the key is unchanged, I don't re-init the key. This avoids the cost of key-scheduling.

Pros:
- Simpler logic
- Minimizes native calls to release contexts (no more than once per Cipher object)
- Minimizes AES key schedule costss

Cons:
- Increases pressure on the Janitor because every use of CTR results a mess that needs to be cleaned up.
- Incurs the above cost even if a Cipher object is only used *once* (so, no savings)

This is slightly simpler than the existing GCM logic which only preserves a context if Cipher object is used twice (with the same key). That is under the assumption that many objects are used once, many are used many times, but very few are only used twice.

I am not wed to my current design. If you think I should change this (simpler or more complicated) please let me know.


SalusaSecondus:
@WillChilds-Klein , I can't trigger the the workflows anymore. Mind kicking them off for me?

SalusaSecondus:
A friendly ping?

WillChilds-Klein:
Hi @SalusaSecondus , sorry for the delay. Would you mind merging in main (particularly contents of #572) to restart CI? looks like it's failed in an odd manner...

tst/com/amazon/corretto/crypto/provider/test/AesCtrTest.java
```diff
@@ -0,0 +1,1155 @@
...
+    }
+    return output;
+  }
+
+  /**
+   * Increments the counter block as a single big-endian 128-bit integer, wrapping to zero on
+   * overflow. This is SP 800-38A's standard incrementing function applied to the full block, which
+   * is what both AWS-LC's {@code EVP_aes_*_ctr} and SunJCE's {@code CounterMode} implement — as
+   * opposed to incrementing only a narrow trailing counter field, which is a property of
+   * constructions layered above the primitive (e.g. RFC 3686) rather than of the primitive itself.
```
WillChilds-Klein, line 577: looks like python-cryptography uses vectors from RFC 3686. can we also test against that RFC's KATs?

https://cryptography.io/en/45.0.3/development/test-vectors/#symmetric-ciphers

https://datatracker.ietf.org/doc/html/rfc3686.html#section-6

README.md
```diff
@@ -53,6 +53,8 @@ Cipher algorithms:
     * AES_\<n\>/CBC/ISO10126Padding, where n can be 128, 192, or 256
 * AES/CFB/NoPadding
     * AES_\<n\>/CFB/NoPadding, where n can be 128 or 256
+* AES/CTR/NoPadding
+    * AES_\<n\>/CTR/NoPadding, where n can be 128 or 256
```
WillChilds-Klein: also adding support for 192, no?

csrc/aes_ctr.cpp
```diff
@@ -0,0 +1,280 @@
...
+        }
+
+        // We need to return the context.
+        if (ctx_container == nullptr) {
+            // This should not happen, as init should always be called with a ctx container
+            EVP_CIPHER_CTX_free(ctx_);
+            throw java_ex(EX_ERROR, "THIS SHOULD NOT BE REACHABLE. No container is provided to return the context.");
+        }
+
+        jlong tmpPtr = reinterpret_cast<jlong>(ctx_);
```
WillChilds-Klein, line 70: why not just use the original `ctx_ptr`?

sfc-gh-grubin: Came as a copy paste from https://github.com/corretto/amazon-corretto-crypto-provider/blob/c287593f8ff31a8dcbf958474baf4c3e00cb1dea/csrc/aes_cfb.cpp#L61

I'll look and figure out why that's there and whether it's needed.

sfc-gh-grubin: We only hit this line if `ctx_ptr` is `nullptr` (`ctx_ptr` is loaded into `ctx_` in the initializer).

```c++
        if (ctx_ != nullptr) {
            // if there is a context, we don't need to do anything.
            return;
        }

        // There is no context, so we need to create one.
        ctx_ = EVP_CIPHER_CTX_new();
        if (ctx_ == nullptr) {
            throw_openssl(EX_RUNTIME_CRYPTO, "EVP_CIPHER_CTX_new failed.");
        }
```

So, at this point we have a *new* `ctx_` value that needs to be returned to the caller.

sfc-gh-grubin:
Feedback addressed. You'll need to kick off the CI again as it doesn't love me anymore 😁

tst/com/amazon/corretto/crypto/provider/test/AesCtrTest.java
sfc-gh-grubin: Done

WillChilds-Klein:
something is up with the CI here. new CodeBuild jobs aren't executed and old ones haven't cleared. i'm going to try to close/re-open this PR to (hopefully) reset CI.

tst/com/amazon/corretto/crypto/provider/test/AesCtrTest.java
```diff
@@ -0,0 +1,1303 @@
...
+  /**
+   * Two distinct {@code SecretKeySpec} instances wrapping identical key bytes are treated as a key
+   * change by the SPI's identity comparison — safe, if conservative, since it merely forgoes the
+   * key-schedule reuse optimization. This confirms output correctness without asserting on which
+   * internal path was taken.
+   *
+   * <p>Documented limitation: this pins intent rather than guarding a failure mode. Because both
+   * candidate implementations (identity comparison, or a value comparison over the encoded bytes)
+   * yield correct output for equal key bytes, the test is insensitive to changing one into the
+   * other — confirmed by mutation, where it passed under both mutations that broke the two tests
```
WillChilds-Klein: our CI is barfing on the emdash

```
/codebuild/output/src3679658278/src/github.com/corretto/amazon-corretto-crypto-provider/tst/com/amazon/corretto/crypto/provider/test/AesCtrTest.java:1128: error: unmappable character (0xE2) for encoding US-ASCII
   * other ??? confirmed by mutation, where it passed under both mutations that broke the two tests
           ^
/codebuild/output/src3679658278/src/github.com/corretto/amazon-corretto-crypto-provider/tst/com/amazon/corretto/crypto/provider/test/AesCtrTest.java:1128: error: unmappable character (0x80) for encoding US-ASCII
   * other ??? confirmed by mutation, where it passed under both mutations that broke the two tests
            ^
/codebuild/output/src3679658278/src/github.com/corretto/amazon-corretto-crypto-provider/tst/com/amazon/corretto/crypto/provider/test/AesCtrTest.java:1128: error: unmappable character (0x94) for encoding US-ASCII
   * other ??? confirmed by mutation, where it passed under both mutations that broke the two tests
```

i have the following in my global CLAUDE.md, which has been effective in preventing this type of error for me

```
 ## ASCII-only in code
 
 NEVER use em dashes (_), en dashes (_), curly/smart quotes (" " ' '), Unicode arrows (_), non-breaking spaces, or any non-ASCII character (codepoint above U+007F) in source code, comments, javadoc, or strings. Use ASCII substitutes: `--` for dashes, straight `"` `'` for quotes, `->` for arrows, `...` for ellipsis. Before committing, run `grep -rnP '[^\x00-\x    7F]'` on changed files to catch strays.
```

SalusaSecondus: Fixed. I hate AI. ☹️

sfc-gh-grubin (approved):

SalusaSecondus:
@WillChilds-Klein , friendly ping?

WillChilds-Klein:
_Written by Claude Opus 5 (Claude Code) on my behalf._

The em dashes are still there at `f81432d` -- that commit changed one line, the occurrence at old line 1128, which happens to be the last one the CodeBuild log printed. The log is truncated at the top, so it undercounts.

Six U+2014 (`e2 80 94`) characters remain in `tst/.../AesCtrTest.java`, all inside javadoc:

| Line | Text |
| --- | --- |
| 173 | `where the first call left off ... a keystream-reuse hazard` |
| 575 | `SunJCE's {@code CounterMode} implement ... as` |
| 847 | `encrypting ... a random IV could not reproduce` |
| 1030 | `Verified by mutation ... commenting out the {@code key == nullptr} guard` |
| 1057 | `Verified by mutation ... forcing {@code` |
| 1121 | `change by the SPI's identity comparison ... safe, if conservative` |

I scanned every file the PR touches; this is the only one affected. (README.md has non-ASCII too, but all of it predates this branch.)

Why it only breaks in CodeBuild: the build passes no `-encoding` to javac, so javac uses the platform charset. CodeBuild runs with no locale set, which gives US-ASCII, while a dev box with a UTF-8 locale compiles the same source without complaint. `LANG=C ./gradlew build` reproduces it locally.

The fix:

```
sed -i 's/\xe2\x80\x94/--/g' tst/com/amazon/corretto/crypto/provider/test/AesCtrTest.java
grep -nP '[^\x00-\x7F]' tst/com/amazon/corretto/crypto/provider/test/AesCtrTest.java   # expect no output
```

Separately, and not for this PR: this class of failure is invisible until CodeBuild runs. Either a non-ASCII grep over `src/` and `tst/` in the build, or an explicit `-encoding US-ASCII` on javac, would surface it on the first local compile.

SalusaSecondus:
Grr. My find-replace missed them. I'll do your strategy and get it fixed momentarily.

SalusaSecondus:
Just rebased on main to get the gradle version bump so this will build on java-25

WillChilds-Klein:
/codebuild_run([d275f7c](https://github.com/corretto/amazon-corretto-crypto-provider/pull/571/commits/d275f7c6958ab7efd384edc9bfc31304ce2ab7cf))

SalusaSecondus:
@WillChilds-Klein, if CI keeps fighting us we could close this PR and then you'd create the actual PR merge the changes in. So long as I don't push any more changes to this branch that might have the checks happy.

Another possibility would be that I merge this into `ctr-work` on this repo (as opposed to my fork) and then you do the PR from there.

csrc/aes_ctr.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+    } catch (java_ex& ex) {
+        ex.throw_to_java(env);
+        return -1;
+    }
+}
+
+extern "C" JNIEXPORT jint JNICALL Java_com_amazon_corretto_crypto_provider_AesCtrSpi_nUpdate(JNIEnv* env,
+    jclass,
+    jint opMode,
```
WillChilds-Klein: this is only used for init, and not in this method. can we drop it?

csrc/aes_ctr.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+    } catch (java_ex& ex) {
+        ex.throw_to_java(env);
+        return -1;
+    }
+}
+
+extern "C" JNIEXPORT jint JNICALL Java_com_amazon_corretto_crypto_provider_AesCtrSpi_nUpdateFinal(JNIEnv* env,
+    jclass,
+    jint opMode,
```
WillChilds-Klein: this is only used for init, and not in this method. can we drop it?

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
```diff
@@ -0,0 +1,612 @@
...
+        this.opMode = ENC_MODE;
+        break;
+      case Cipher.DECRYPT_MODE:
+      case Cipher.UNWRAP_MODE:
+        this.opMode = DEC_MODE;
+        break;
+      default:
+        throw new InvalidAlgorithmParameterException("Invalid opmode: " + opmode);
+    }
+    this.needsKeyInit = this.key != key; // Identity equality check
```
WillChilds-Klein: if init is called twice with the same key, then will flip needsKeyInit to false on the second call, when we've never actually done the "key init" in update() on L285

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
```diff
@@ -0,0 +1,612 @@
...
+        } catch (final InvalidKeyException ex) {
+          // This should be impossible. Our key has changed out from under us
+          cleanUpNativeContextIfNeeded(maybeCtxContainer);
+          throw new RuntimeCryptoException("Unexpected error", ex);
+        } catch (final Exception ex) {
+          cleanUpNativeContextIfNeeded(maybeCtxContainer);
+          throw ex;
+        }
+      }
+    } else if (saveContext) {
```
WillChilds-Klein: from claude: the saved-context branches of doFinal and update dropped the try/catch that AesCbcSpi uses, so a native failure skips the trailing needsNativeInit = true and a retry resumes on a context at an indeterminate counter position.

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
```diff
@@ -0,0 +1,612 @@
...
+      int inputOffset,
+      int inputLen,
+      ByteBuffer outputDirect,
+      byte[] outputArray,
+      int outputOffset);
+
+  @Override
+  protected byte[] engineWrap(final Key key) throws IllegalBlockSizeException, InvalidKeyException {
+    try {
+      final byte[] encoded = Utils.encodeForWrapping(key);
```
WillChilds-Klein: here and on L597, can we use the other overload of this function and pass `this.provider`?

https://github.com/corretto/amazon-corretto-crypto-provider/blob/3c7a841b8c5b66001e3fdd29ea288edda25d32f8/src/com/amazon/corretto/crypto/provider/Utils.java#L186

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
```diff
@@ -212,6 +212,16 @@ private void buildServiceMap() {
...
 
+    addService(
+        "Cipher",
+        "AES/CTR",
+        "AesCtrSpi",
+        true,
+        singletonMap("SupportedModes", "CTR"),
+        "AES_128/CTR",
+        "AES_192/CTR",
+        "AES_256/CTR");
```
WillChilds-Klein, line 223: from claude: AES_128/CTR, AES_192/CTR and AES_256/CTR are registered as plain aliases, so the key-size assertion those Standard Names exist to make is not enforced and a 256-bit key passes as AES_128.

looks like other ciphers do this too, so feel free to defer this to me in a follow-up

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
```diff
@@ -0,0 +1,612 @@
...
+
+  @Override
+  protected void engineInit(final int opmode, final Key key, final SecureRandom random)
+      throws InvalidKeyException {
+    if (opmode != Cipher.ENCRYPT_MODE && opmode != Cipher.WRAP_MODE) {
+      throw new InvalidKeyException("IV required for decrypt");
+    }
+    try {
+      byte[] iv = new byte[IV_SIZE_IN_BYTES];
+      random.nextBytes(iv);
```
WillChilds-Klein, line 109: NPE if `random == null`

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
```diff
@@ -0,0 +1,612 @@
...
+      final byte[] outputArray,
+      final int outputOffset) {
+    final int result;
+    if (needsNativeInit) {
+      // We need to re-initialize the EVP_CipherCtx object.
+      if (context == null) {
+        // No context, so create it
+        final long[] ctxContainer = new long[] {0};
+        try {
+          final byte[] keyBytes = Utils.checkAesKey(key);
```
WillChilds-Klein: from claude: `Utils.checkAesKey(key)` runs on every operation, leaving a fresh unzeroed copy of key material on the heap per call and negating the needsKeyInit optimization, where CBC and CFB cache the key bytes once at init.

WillChilds-Klein (commented):
did a more thorough review, should be good to go once these are addressed.

don't worry about cutting a new PR, i think i figured out the CI stuff (persistently "failing" ones appear to be redundant and non-required, both for unknown reasons) and it shouldn't block us.

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
SalusaSecondus: This doesn't make sense.

1. From the immediately prior lines, this is only reached when `context` is `null`, so we do need the raw key bytes. If the context (based on the resource release mode) can be saved we avoid calling this method again next time.  
2. We don't generally zeroize the java byte arrays containing key bytes. I can add this though.

It is right that when we don't store the context that we pull the key bytes more commonly rather than just hold them in memory on the object. Do you have a preference here?

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
SalusaSecondus: Happy to fix however we should figure out a consistent behavior:
- https://github.com/corretto/amazon-corretto-crypto-provider/blob/3c7a841b8c5b66001e3fdd29ea288edda25d32f8/src/com/amazon/corretto/crypto/provider/AesCfbSpi.java#L103
- https://github.com/corretto/amazon-corretto-crypto-provider/blob/3c7a841b8c5b66001e3fdd29ea288edda25d32f8/src/com/amazon/corretto/crypto/provider/AesGcmSpi.java#L304

src/com/amazon/corretto/crypto/provider/AmazonCorrettoCryptoProvider.java
SalusaSecondus: Deferring to you then. I think that this will require looking at other providers to see what they do.

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
SalusaSecondus: Leaving unresolved so you can see the fix and resolve yourself.

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
SalusaSecondus: Leaving unresolved so you can see the fix and resolve yourself.

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
WillChilds-Klein: ah, true. defer to me in a follow-up.

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
WillChilds-Klein: LGTM

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
WillChilds-Klein: my read is that because `checkAesKey` [returns a copy](https://github.com/corretto/amazon-corretto-crypto-provider/blob/3c7a841b8c5b66001e3fdd29ea288edda25d32f8/src/com/amazon/corretto/crypto/provider/Utils.java#L673) (cf. "Returns a new array each time this method is called" in [javadoc](https://docs.oracle.com/javase/8/docs/api/javax/crypto/spec/SecretKeySpec.html#getEncoded--)), the secret contents of `keyBytes` are left unzeroed on the heap when this function exits. maybe we can zero `keyBytes` in a `finally` block on L277?

WillChilds-Klein (commented):
Written by Claude Opus 5 (Claude Code) on my behalf.

  CI on 80eff2a is red because :spotlessJavaCheck rejects the formatting of two files. Four of the six batches fail on it. AccpCiIntegrationTests and AccpCiFuzzingTests pass because they do not run the check.

  The violations all come from the "Address more feedback" commit:

  - AesCtrSpi.java -- needsNativeInit = true; plus its trailing comment runs past 100 columns, so google-java-format wants the assignment split across two lines.
  - AesCtrSpi.java -- the new Utils.buildUnwrappedKey(this.provider, ...) call overflows and needs wrapping.
  - AesCtrTest.java -- the // Shift to key2, but do a double init ... comment needs rewrapping, and a blank line before the RFC 3686 KATs carries trailing whitespace.

  Gradle prints the exact patch it wants:

  > The following files had format violations:
        src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
            @@ -400,7 +400,8 @@
            -    needsNativeInit = true; // Reset this value immediately so no matter what happens we reset on the next call
            +    needsNativeInit =
            +        true; // Reset this value immediately so no matter what happens we reset on the next call

  To fix:

  ./gradlew spotlessJavaApply

  Use spotlessJavaApply, not spotlessApply. The latter also runs clang-format over csrc/, which has drift unrelated to this PR and would drag unrelated files into the diff.

  Push that and I will re-run the builds. **If you would rather not, tick "Allow edits by maintainers" on this PR and I will push the formatting commit myself.**

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
WillChilds-Klein: >> negating the needsKeyInit optimization, where CBC and CFB cache the key bytes once at init.

yeah i don't really understand this either.

>> We don't generally zeroize the java byte arrays containing key bytes. I can add this though.

we've started to, where possible, across ACCP, aws-lc-rs, and aws-lc.

SalusaSecondus:
I'll fix and my apologies. I'm still getting my ACCP dev environment fully polished and so I forgot to add `./gradlew spotlessJavaApply` to my standard workflow. (That might be worth added as a precommit script.)

For the `Utils.checkAesKey()` call, that's right. I don't think we generally zeroize these buffers in Java. But we also have a tendency to have them live as long as the class, rather than being truly ephemeral.

So, I think we have the following choices (and I'm fine with any of them):

1. Zeroize this buffer when we're done with it
2. Cache the raw key bytes in a field so we only end up with one copy of them, but don't zeroize them. (This is what everything else does that I can find.)
3. Same as 2, but zeroize when the key is replaced by a new key
4. Same as 3, but also register a cleaner for when this Cipher object is garbage collected (I'd prefer not this one.)
5. Leave it as it is
6. Some variant of 2-4, but only done when `saveContext` is `false` (because when it is `true`, we don't need to keep the key bytes saved at all) 

Thoughts?

WillChilds-Klein:
no worries at all.

> Zeroize this buffer when we're done with it

I'd rather not save these off as state anywhere, I think it's easiest to just zeroize `keyBytes` in a finally block on L277. zeroization of intermediate buffers is defense-in-depth, so not _super crucial_, but nice to have.

SalusaSecondus:
@WillChilds-Klein , as I look at this and properly zeroising this, do you want me to just create a new `AutoCloseable` class to make this easier here and elsewhere?

WillChilds-Klein:
@SalusaSecondus -- sure, sounds good. thank you.

SalusaSecondus:
There we go. I checked spotless before committing. The new class is `AutoDestroyer` which is used with try-with-resources. You can register various things with it and it will wipe them when it closes. If there is no value to wiping it (null, empty, or read-only) it just ignores it. It always passes through the object to make it easy to use and insert in-line.

Hope you like it!

csrc/aes_ctr.cpp
```diff
@@ -0,0 +1,278 @@
...
+    jint inputLen,
+    jobject outputDirect,
+    jbyteArray outputArray,
+    jint outputOffset)
+{
+    try {
+        AesCtrCipher aes_ctr_cipher(env, nullptr, ctxPtr, true);
+
+        // update
+        JIOBlobs io_blobs(env, inputDirect, inputArray, outputDirect, outputArray);
```
WillChilds-Klein, line 241: note for other reviewers: one of the 2 input/output arrays here should always be non-null, meaining that buffer.cpp's input_ptr_ / output_ptr_ will always be initialized and pointer arithmetic below is sound

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
```diff
@@ -0,0 +1,615 @@
...
+        this.opMode = ENC_MODE;
+        break;
+      case Cipher.DECRYPT_MODE:
+      case Cipher.UNWRAP_MODE:
+        this.opMode = DEC_MODE;
+        break;
+      default:
+        throw new InvalidAlgorithmParameterException("Invalid opmode: " + opmode);
+    }
+    this.needsKeyInit |= this.key != key; // Identity equality check
```
WillChilds-Klein, line 175: concern from claude:

> The SPI stores the caller's SecretKey and decides staleness by object identity (needsKeyInit |= this.key != key), so re-initializing the same key object whose getEncoded() has changed keeps the old key schedule in a saved context: silent encryption under the wrong key. AesCbcSpi and AesCfbSpi snapshot the encoded bytes at init instead.

i don't share this concern as JDK documents that SecretKey must compose its hashCode from contents not reference

> Provider implementations of this interface must overwrite the equals and hashCode methods inherited from [Object](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Object.html), so that secret keys are compared based on their underlying key material and not based on reference.

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/javax/crypto/SecretKey.html

WillChilds-Klein (approved):

src/com/amazon/corretto/crypto/provider/AesCtrSpi.java
SalusaSecondus: More than that, `Key` objects are assumed to be immutable in Java. So if they were to ever change value it would horribly break much more than just this.

sgmenda:
```suggestion
    // This identity comparison is never unsafe: a `java.security.Key` is immutable.
    // A different `Key` with the same bytes is treated as a key change and re-inits,
    // as shown by `testDifferentKeyObjectSameBytes`.
    this.needsKeyInit |= this.key != key;
```

csrc/aes_ctr.cpp
sgmenda: I checked this.

sgmenda (approved):
lgtm

PR #574, opened 2026-08-26
Support "Generic" output algorithm in EvpKeyAgreement

torben-hansen:
JDK 27's TLS 1.3 hybrid key exchange (JEP 527) derives the classical DH
shared secret via `KeyAgreement.generateSecret("Generic")`. ACCP currently
rejects this with `InvalidKeyException("Unsupported algorithm: Generic")`,
which breaks TLS handshakes that route key agreement through ACCP.

The default JDK provider returns the raw agreed secret when asked for the
`"Generic"` output algorithm. This change makes `EvpKeyAgreement` do the
same: when `generateSecret("Generic")` is called, it wraps the raw agreed
secret in a `SecretKeySpec` with algorithm `"Generic"` (RAW format),
matching the JDK behavior so TLS handshakes using ACCP succeed.

## Changes

- `EvpKeyAgreement.java`: handle `"Generic"` in `engineGenerateSecret(String)`,
  returning the raw agreed secret as a `Generic`/RAW `SecretKeySpec`.
- `EvpKeyAgreementTest.java`: add a parameterized `genericSecret` test
  asserting the returned key has algorithm `Generic`, format `RAW`, and
  encoding equal to the raw shared secret.

## Testing

- Added unit test `genericSecret` covering the new algorithm across all
  existing key-agreement test parameters.

### DIFF
--- a/src/com/amazon/corretto/crypto/provider/EvpKeyAgreement.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpKeyAgreement.java
@@ -88,10 +88,16 @@ class EvpKeyAgreement extends KeyAgreementSpi {
     if (algorithm.equalsIgnoreCase("TlsPremasterSecret")) {
       return new SecretKeySpec(secret, "TlsPremasterSecret");
     }
+    // Both "com/sun/crypto/provider/DHKEM.java" (backing HPKE and the "DHKEM"
+    // KEM) and "sun/security/ssl/DHasKEM.java" (the classical half of TLS 1.3
+    // hybrid key exchange, e.g. X25519MLKEM768) obtain the raw agreed secret via
+    // generateSecret("Generic"), "Generic" being the default output algorithm of
+    // the javax.crypto.KEM API. Neither pins a provider when requesting
+    // ECDH/XDH, so ACCP wins the service and must accept "Generic" or it breaks
+    // HPKE and TLS handshakes for anyone who installs it.
     if (algorithm.equalsIgnoreCase("Generic")) {
       return new SecretKeySpec(secret, "Generic");
     }
-    ;
     final Matcher matcher = ALGORITHM_WITH_EXPLICIT_KEYSIZE.matcher(algorithm);
     if (matcher.matches()) {
       switch (matcher.group(1)) {
--- a/src/com/amazon/corretto/crypto/provider/EvpKeyAgreement.java
+++ b/src/com/amazon/corretto/crypto/provider/EvpKeyAgreement.java
@@ -88,6 +88,9 @@ class EvpKeyAgreement extends KeyAgreementSpi {
     if (algorithm.equalsIgnoreCase("TlsPremasterSecret")) {
       return new SecretKeySpec(secret, "TlsPremasterSecret");
     }
+    if (algorithm.equalsIgnoreCase("Generic")) {
+      return new SecretKeySpec(secret, "Generic");
+    }
     ;
     final Matcher matcher = ALGORITHM_WITH_EXPLICIT_KEYSIZE.matcher(algorithm);
     if (matcher.matches()) {
--- a/tst/com/amazon/corretto/crypto/provider/test/EvpKeyAgreementTest.java
+++ b/tst/com/amazon/corretto/crypto/provider/test/EvpKeyAgreementTest.java
@@ -279,6 +279,20 @@ public class EvpKeyAgreementTest {
         });
   }
 
+  @ParameterizedTest
+  @MethodSource("params")
+  public void genericSecret(TestParams params) throws GeneralSecurityException {
+    // For TLS 1.3 hybrid key exchange we must support the "Generic" algorithm
+    final byte[] rawSecret = params.rawSecrets[0][1];
+    params.nativeAgreement.init(params.pairs[0].getPrivate());
+    assertNull(params.nativeAgreement.doPhase(params.pairs[1].getPublic(), true));
+
+    final SecretKey genericKey = params.nativeAgreement.generateSecret("Generic");
+    assertEquals("Generic", genericKey.getAlgorithm());
+    assertEquals("RAW", genericKey.getFormat());
+    assertArrayEquals(rawSecret, genericKey.getEncoded());
+  }
+
   private static Stream<TestParams> aesKeysParams() {
     return params().stream()
         .filter(


tst/com/amazon/corretto/crypto/provider/test/EvpKeyAgreementTest.java
```diff
@@ -279,6 +279,20 @@ public void tlsMasterSecret(TestParams params) {
         });
   }
 
+  @ParameterizedTest
+  @MethodSource("params")
+  public void genericSecret(TestParams params) throws GeneralSecurityException {
```
WillChilds-Klein, line 284: can we add a test in our [non-generic key agreement test suite](https://github.com/corretto/amazon-corretto-crypto-provider/blob/fb52a05fc0f061542af82c270f6e1a0040f8ad1d/tst/com/amazon/corretto/crypto/provider/test/EvpKeyAgreementSpecificTest.java#L34) that addresses the specific compatibility concern? is it `X25519MLKEM768` or something else?

src/com/amazon/corretto/crypto/provider/EvpKeyAgreement.java
```diff
@@ -88,6 +88,9 @@ protected SecretKey engineGenerateSecret(final String algorithm)
     if (algorithm.equalsIgnoreCase("TlsPremasterSecret")) {
       return new SecretKeySpec(secret, "TlsPremasterSecret");
     }
+    if (algorithm.equalsIgnoreCase("Generic")) {
+      return new SecretKeySpec(secret, "Generic");
+    }
```
WillChilds-Klein, line 100: Since JDK 27's jtreg run is what surfaced this, could we cite the failing test here? This file already does exactly that ~60 lines up for another jtreg-driven behavior:

```java
    // to conform with JCE as described in JTREG test
    // "sun/security/ec/ECDHKeyAgreementParamValidation.java".
```

tst/com/amazon/corretto/crypto/provider/test/EvpKeyAgreementTest.java
WillChilds-Klein: Answering my own question after digging in: it isn't `X25519MLKEM768`, and I don't think a test in the specific suite is the right ask.

The hybrid group name never reaches a provider -- JSSE composes the hybrid itself and calls `KeyAgreement.getInstance("XDH")` + `generateSecret("Generic")` on the classical side and `KEM.getInstance("ML-KEM")` on the KEM side. So the compatibility surface is exactly what this test already exercises.

And it exercises it thoroughly: `params()` covers ECDH on P-256/P-384/P-521 plus an explicit curve, and X25519 under **both** the `X25519` and `XDH` service names. Those are the classical halves of both standardized hybrids (X25519MLKEM768 -> XDH/X25519, SecP256r1MLKEM768 -> ECDH P-256), and `XDH` is the name JSSE actually requests, per the comment already at L74-77. A near-identical assertion in `EvpKeyAgreementSpecificTest` would be duplicate coverage, so let's not add one.

What's actually missing is provenance, not an assertion -- the authoritative test already exists in the JDK and already did its job. Left a suggestion on the source change to cite the jtreg test the way this file already cites `sun/security/ec/ECDHKeyAgreementParamValidation.java`.

Worth noting why no ACCP test can really cover the concern: the end-to-end path needs a JDK that implements the hybrid groups, and CI is JDK 8/11/17/21. A JDK 27 job running the relevant jtreg subset would be the real protection, but that's a separate issue, not a blocker here.

src/com/amazon/corretto/crypto/provider/EvpKeyAgreement.java
```diff
@@ -88,6 +88,9 @@ protected SecretKey engineGenerateSecret(final String algorithm)
...
     ;
```
WillChilds-Klein: Unrelated drive-by, but this PR touches the line right above it: that stray empty statement (`;`) is dead and predates this change. Easy cleanup while you're in here, or leave it -- entirely up to you.

WillChilds-Klein: Removed in 6743e39 since I was in the file anyway.

WillChilds-Klein (approved):

WillChilds-Klein:
Two observations on `engineGenerateSecret(String)` that go beyond the added `Generic` branch. Both are pre-existing, but this change lands directly on top of them and the PR's motivation runs through both.

### 1. The agreed secret is consumed before the algorithm name is validated

The method opens with:

```java
byte[] secret = engineGenerateSecret();
```

and the no-arg `engineGenerateSecret()` calls `reset()`, nulling the `secret` field. Every validation failure after that point -- `Invalid key length`, `Unsupported algorithm`, `Unrecognized algorithm` -- therefore leaves the object unusable rather than just rejecting the request:

```java
ka.init(priv);
ka.doPhase(pub, true);
try {
  ka.generateSecret("Generic[32]");
} catch (InvalidKeyException expected) {
  // fall back to the plain name
}
ka.generateSecret("Generic"); // IllegalStateException: KeyAgreement has not been completed
```

SunEC validates the requested name first and only derives afterwards, so a caller there can probe. That distinction matters for this PR specifically: code that has to work against both pre- and post-JEP-527 providers is likely to try one output-algorithm name and fall back to another, and against ACCP the fallback fails for a reason unrelated to the name it asked for. It also sits awkwardly with the "KeyAgreement supports reuse without reinitialization" property documented in `DIFFERENCES.md`.

Suggestion: resolve the target algorithm (including the AES keysize) first, and call `engineGenerateSecret()` only once a valid target is known. As a smaller related point, `generateSecret(null)` currently NPEs out of `algorithm.equalsIgnoreCase(...)` after the secret has already been destroyed; SunEC guards this with an explicit `NoSuchAlgorithmException("Algorithm must not be null")`.

### 2. Unsupported names throw `InvalidKeyException` where JCA specifies `NoSuchAlgorithmException`

Both terminal throws use `InvalidKeyException`:

```java
throw new InvalidKeyException("Unsupported algorithm: " + matcher.group(1));
...
throw new InvalidKeyException("Unrecognized algorithm: " + algorithm);
```

[`KeyAgreement.generateSecret(String)`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/javax/crypto/KeyAgreement.html#generateSecret(java.lang.String)) documents `NoSuchAlgorithmException` "if the specified secret key algorithm is not available", and this method already declares `NoSuchAlgorithmException` but never throws it.

This is also relevant to the failure mode the description covers. A caller written as

```java
try {
  return ka.generateSecret("Generic");
} catch (NoSuchAlgorithmException e) {
  return new SecretKeySpec(ka.generateSecret(), "Generic");
}
```

does not get its fallback against ACCP -- the `InvalidKeyException` propagates and the handshake fails anyway. Adding `Generic` by name fixes the one name JDK 27 happens to use; switching these two throws to `NoSuchAlgorithmException` fixes the class, so the next output-algorithm name a JDK introduces degrades gracefully instead of breaking TLS handshakes that route through ACCP. `InvalidKeyException` would remain correct for the genuine `Invalid key length` case, where the name is recognized but the agreed secret is too short.

---

Both are independent of the `Generic` support itself and are reasonable follow-ups if you would rather keep this PR minimal, though (2) is a two-line change.

<sub>Raised from an automated review pass; flagging for a human to confirm, particularly the assumption about which output-algorithm name JSSE requests on JDK 27 and whether it has a fallback path.</sub>

src/com/amazon/corretto/crypto/provider/EvpKeyAgreement.java
```diff
@@ -88,7 +88,16 @@ protected SecretKey engineGenerateSecret(final String algorithm)
     if (algorithm.equalsIgnoreCase("TlsPremasterSecret")) {
       return new SecretKeySpec(secret, "TlsPremasterSecret");
     }
```
justsmth, line 90: Nit -- pre-existing: `engineGenerateSecret(String)` consumes the secret `engineGenerateSecret()` *before* validating the name. It should validate first.

src/com/amazon/corretto/crypto/provider/EvpKeyAgreement.java
justsmth, line 130: Nit -- pre-existing: Unknown names throw `InvalidKeyException` even though the method declares `NoSuchAlgorithmException` and that is what the JCA spec / SunEC use. A caller that catches `NoSuchAlgorithmException` to fall back to `generateSecret()` + `SecretKeySpec` does not get that fallback against ACCP.

justsmth (approved):

src/com/amazon/corretto/crypto/provider/EvpKeyAgreement.java
WillChilds-Klein: agreed, 1 [here](https://github.com/corretto/amazon-corretto-crypto-provider/pull/574#issuecomment-5441393772)

PR #575, opened 2026-08-27
Make the generic ML-KEM KeyPairGenerator parameter-set-selectable; harden MlDsaUtils against unencodable keys

WillChilds-Klein:
## Context

Two follow-ups deferred out of #569.

## 1. The generic `ML-KEM` KeyPairGenerator becomes parameter-set-selectable

ACCP registers a parameter-set-agnostic `ML-KEM` service alongside the sized
`ML-KEM-512`/`ML-KEM-768`/`ML-KEM-1024` ones for `KeyPairGenerator`, `KeyFactory` and `KEM`. Two of
those three generic services are already parameter-agnostic: `KEM` takes its parameter set from the
key it is handed, and `KeyFactory` reads it out of the encoding. Only `KeyPairGenerator "ML-KEM"` was
not: it was backed by `MlKemGen768`, so it accepted a `NamedParameterSpec` naming ML-KEM-768 and
rejected ML-KEM-512 and ML-KEM-1024 with `InvalidAlgorithmParameterException`. A caller working
through the generic service names -- a JSSE stack, or provider-agnostic application code -- had to
switch to a sized name for anything other than 768.

It is now backed by a new `MlKemGenGeneric` that overrides `initialize` to select the named parameter
set. This mirrors two existing precedents:

* SunEC's generic `XDH` service accepts either `X25519` or `X448`, while its `X25519`/`X448` services
  stay bound to their own curve.
* BouncyCastle's own generic ML-KEM `KeyPairGenerator` already accepts a `NamedParameterSpec` for all
  three parameter sets, and defaults to ML-KEM-768 when left uninitialized. Verified against
  bcprov-jdk18on 1.85.

The sized generators are unchanged: each still accepts only a spec naming its own parameter set and
rejects any other, so ACCP never silently returns a key of the wrong parameter set. That rejection is
what lets the JCA fail over to another provider when the caller obtained the generator *without*
naming one; a caller that named ACCP explicitly sees the `InvalidAlgorithmParameterException`, since
the JCA has no other provider to try. The uninitialized generic generator still produces ML-KEM-768,
so no existing caller changes behavior.

The parameter set may be named by any spec exposing a public `String getName()`, not only a
`NamedParameterSpec`. BouncyCastle's TLS stack initializes its ML-KEM `KeyPairGenerator` with
`MLKEMParameterSpec`, and BC's own generic generator accepts either shape there, so restricting this
to `NamedParameterSpec` would have left the JSSE interop this change is meant to unblock still broken
for that caller. The name is matched against the supported parameter sets either way, so a foreign
spec cannot select an unsupported set or bypass the sized generators' check.

The case-insensitive name lookup lands on `MlKemParameter` next to `matchesAlgorithmName`, which it
reuses, so every ML-KEM SPI agrees on which parameter set names are accepted. `fromKemName` delegates
to it rather than duplicating the mapping, and the supported-name list in error messages is built
from `values()` instead of hardcoded.

`MlKemGen.parameterSet` is `volatile`: `MlKemGenGeneric` writes it from `initialize()` while
`generateKeyPair()` reads it, so a generator shared between threads could otherwise read a stale
parameter set and produce a key of the wrong one.

### Tests

`MlKemTest` goes from 60 to 72 tests:

* selection through the generic service for each of the three parameter sets;
* the uninitialized ML-KEM-768 default;
* rejection of a name that is not an ML-KEM parameter set (`ML-KEM-4096`, `X25519`);
* the sized generators remaining strict;
* `testKeyPairGeneratorAcceptsForeignNamedSpec`, which drives both the generic and the sized
  generators with BouncyCastle's `MLKEMParameterSpec` and confirms the sized ones still reject a
  foreign spec naming a different parameter set;
* `testGenericServiceNameInteropWithBouncyCastle`, which drives key generation, cross-import through
  both providers' generic `KeyFactory`, and encapsulate/decapsulate in both directions using *only*
  the `"ML-KEM"` service names on both sides, for all three parameter sets. For 512 and 1024 that
  test could not have been written before this change.

That test and `testBouncyCastleInteroperability` differ only in which ACCP service name they use, so
they now share one helper parameterized on it rather than duplicating ~50 lines. The BouncyCastle
KEM-API availability check that `testBouncyCastleInteroperability` carried inline is also a shared
helper now.

## 2. `MlDsaUtils` rejects an unusable key instead of crashing the JVM

`MlDsaUtils.computeMu` and `MlDsaUtils.expandPrivateKey` checked their key argument for null and for
an ML-DSA algorithm name, then passed `key.getEncoded()` straight to JNI. `Key.getEncoded()` is
specified to return null for a key that does not support encoding -- a key held in hardware, say --
and neither check catches that, so the null reached `GetArrayLength(nullptr)` and took the process
down with a `SIGSEGV` rather than throwing. Confirmed for both methods by removing the fix and
re-running the new test: exit 1 with an `hs_err` file.

This is the same defect [justsmth found](https://github.com/corretto/amazon-corretto-crypto-provider/pull/569#discussion_r3863499611)
in `MlKemUtils.expandPrivateKey`, fixed there in #569; this applies the identical guard to the two
ML-DSA entry points and documents the throw.

`Key.getAlgorithm()` is likewise not specified to return a non-null value, so both classes null-check
it instead of dereferencing it -- previously a key returning null there produced a `NullPointerException`
from the `startsWith` call rather than the documented `IllegalArgumentException`. Every
`IllegalArgumentException` these two classes throw now carries a message.

## 3. Hardening in the ML-DSA JNI helpers

Found while reviewing the above.

`MlDsaUtils.expandPrivateKey` short-circuited on the input length:

```c
if (key_der_len > 54) { // If they key is already expanded, return it
    return keyBytes;
}
```

Anything longer than a 54-byte seed key was handed straight back to the caller, so a foreign or
malformed encoding round-tripped as though it were an expanded ML-DSA key -- including a PKCS8 blob
for an entirely different algorithm. The helper now parses and re-encodes every input through
`der2EvpPrivateKey`, matching `MlKemUtils.expandPrivateKey`, which already worked this way.
`der2EvpPrivateKey` rejects what it cannot parse and rejects a key that parses as another algorithm,
so those inputs raise instead of being echoed. The output is therefore always the canonical minimal
`expandedKey` PKCS8, which is documented on the Java method.

That rewrite also removes two problems in the code it replaced:

* `EVP_PKCS82PKEY`'s return value was unchecked and passed to `encodeExpandedMLDSAPrivateKey` by
  `OPENSSL_auto`'s implicit conversion, which calls `abort()` on a null pointer. A PKCS8 blob that
  parsed as a `PKCS8_PRIV_KEY_INFO` but not as an `EVP_PKEY` -- an unsupported algorithm OID -- would
  have taken the process down.
* the `BIO` from `BIO_new_mem_buf` was never freed.

In `computeMuInternal`, `EVP_parse_public_key`'s return value was unchecked; it only failed to crash
because `EVP_PKEY_get_raw_public_key` happens to null-check its argument. It is now checked, and a
well-formed SPKI for another algorithm is rejected rather than having mu computed over a non-ML-DSA
public key. The unused `EVP_PKEY_CTX` in that function is gone.

Finally, `MlDsaGen.initialize(AlgorithmParameterSpec, SecureRandom)` threw
`UnsupportedOperationException`. `KeyPairGenerator.Delegate` expects the checked
`InvalidAlgorithmParameterException` there, so the unchecked throw propagated to the caller instead of
letting the JCA fail over. These generators are bound to their parameter set at construction, so every
spec is still rejected -- just with the exception the JCA can act on.

### DIFF
--- a/csrc/util_class.cpp
+++ b/csrc/util_class.cpp
@@ -92,6 +92,13 @@ JNIEXPORT jint JNICALL Java_com_amazon_corretto_crypto_provider_Utils_getDigestL
  * Class:     com_amazon_corretto_crypto_utils_MlDsaUtils
  * Method:    expandPrivateKeyInternal
  * Signature: ([B)[B
+ *
+ * Every input is parsed and re-encoded, as in the ML-KEM helper below, so the output is always the
+ * canonical minimal expandedKey PKCS8. Deliberately not short-circuiting on the input length: that
+ * echoed any encoding longer than a 54-byte seed key back to the caller unvalidated.
+ *
+ * EX_RUNTIME_CRYPTO rather than EX_INVALID_KEY_SPEC because InvalidKeySpecException is checked and
+ * MlDsaUtils.expandPrivateKey does not declare it.
  */
 JNIEXPORT jbyteArray JNICALL
 Java_com_amazon_corretto_crypto_utils_MlDsaUtils_expandPrivateKeyInternal(JNIEnv* pEnv, jclass, jbyteArray keyBytes)
@@ -100,36 +107,29 @@ Java_com_amazon_corretto_crypto_utils_MlDsaUtils_expandPrivateKeyInternal(JNIEnv
     try {
         raii_env env(pEnv);
         jsize key_der_len = env->GetArrayLength(keyBytes);
-
-        if (key_der_len > 54) { // If they key is already expanded, return it
-            return keyBytes;
-        }
-        CHECK_OPENSSL(key_der_len == 54); // seed-only keys are always 54 bytes when PKCS8-encoded
         uint8_t* key_der = (uint8_t*)env->GetByteArrayElements(keyBytes, nullptr);
         CHECK_OPENSSL(key_der);
 
+        EVP_PKEY_auto key;
         try {
-            // Parse the seed key
-            BIO* key_bio = BIO_new_mem_buf(key_der, key_der_len);
-            CHECK_OPENSSL(key_bio);
-            PKCS8_PRIV_KEY_INFO_auto pkcs8
-                = PKCS8_PRIV_KEY_INFO_auto::from(d2i_PKCS8_PRIV_KEY_INFO_bio(key_bio, nullptr));
-            CHECK_OPENSSL(pkcs8.isInitialized());
-            EVP_PKEY_auto key = EVP_PKEY_auto::from(EVP_PKCS82PKEY(pkcs8));
-
-            // Expand the seed key and encode it before returning
-            OPENSSL_buffer_auto new_der;
-            int new_der_len = encodeExpandedMLDSAPrivateKey(key, &new_der);
-            CHECK_OPENSSL(new_der_len > 0);
-            if (!(result = env->NewByteArray(new_der_len))) {
-                throw_java_ex(EX_OOM, "Unable to allocate DER array");
-            }
-            env->SetByteArrayRegion(result, 0, new_der_len, (const jbyte*)new_der);
+            key.set(der2EvpPrivateKey(
+                key_der, key_der_len, EVP_PKEY_PQDSA, /*shouldCheckPrivate*/ false, EX_RUNTIME_CRYPTO));
         } catch (...) {
             env->ReleaseByteArrayElements(keyBytes, (jbyte*)key_der, JNI_ABORT);
             throw;
         }
         env->ReleaseByteArrayElements(keyBytes, (jbyte*)key_der, JNI_ABORT);
+
+        // Expand the key and encode it before returning. key.get() rather than an implicit
+        // conversion, which aborts the process on a null pointer.
+        OPENSSL_buffer_auto new_der;
+        int new_der_len = encodeExpandedMLDSAPrivateKey(key.get(), &new_der);
+        CHECK_OPENSSL(new_der_len > 0);
+
+        if (!(result = env->NewByteArray(new_der_len))) {
+            throw_java_ex(EX_OOM, "Unable to allocate DER array");
+        }
+        env->SetByteArrayRegion(result, 0, new_der_len, (const jbyte*)new_der);
     } catch (java_ex& ex) {
         ex.throw_to_java(pEnv);
         return 0;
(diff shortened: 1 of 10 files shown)


geedo0 (approved):

src/com/amazon/corretto/crypto/provider/MlKemGen.java
```diff
@@ -50,6 +52,12 @@ private MlKemGen(MlKemParameter parameterSet) {
     this.parameterSet = parameterSet;
   }
 
+  // Sole writer of |parameterSet| after construction. Package-private rather than private because a
+  // private member of this class is not inherited by the nested MlKemGenGeneric subclass.
```
justsmth, line 56: Nit: A nested class can still write `this.parameterSet` directly. (A "setter" is the better option, but this comment should be revised.

src/com/amazon/corretto/crypto/provider/MlKemGen.java
```diff
@@ -58,10 +66,13 @@ public void initialize(int keysize, SecureRandom random) {
   /**
    * Accepts the standard {@code NamedParameterSpec} initialization that JSSE providers (e.g.
    * BouncyCastle's TLS 1.3 stack) and application code perform before {@code generateKeyPair()}.
```
justsmth, line 68: Nit: The implementation also accepts BC's spec? Should that be mentioned?

csrc/util_class.cpp
```diff
@@ -215,8 +214,13 @@ JNIEXPORT jbyteArray JNICALL Java_com_amazon_corretto_crypto_utils_MlDsaUtils_co
...
             CBS_init(&cbs, pub_key_der, pub_key_der_len);
-            EVP_PKEY_auto pkey = EVP_PKEY_auto::from((EVP_parse_public_key(&cbs)));
-            EVP_PKEY_CTX_auto ctx = EVP_PKEY_CTX_auto::from(EVP_PKEY_CTX_new(pkey.get(), nullptr));
+            EVP_PKEY_auto pkey = EVP_PKEY_auto::from(EVP_parse_public_key(&cbs));
+            CHECK_OPENSSL(pkey.isInitialized());
+            // A well-formed SPKI for another algorithm parses fine, so without this check mu could
+            // be computed over a non-ML-DSA public key.
+            if (EVP_PKEY_id(pkey.get()) != EVP_PKEY_PQDSA) {
+                throw_java_ex(EX_ILLEGAL_ARGUMENT, "Not an ML-DSA public key");
+            }
```
justsmth, line 223: Nit: This check is the right defense-in-depth -- a well-formed RSA/EC SPKI would parse and otherwise have mu computed over a non-ML-DSA public key.

`testUnusableKeysAreRejected` never reaches it: every case is rejected in Java before JNI. Worth a `mlDsaDisabled()`-gated case that hands `computeMu` a `TestKey("ML-DSA-44", rsaSpki)` (or any well-formed non-PQDSA SPKI) and asserts `IllegalArgumentException`.

justsmth (approved):

PR #577, opened 2026-09-02
Make ML-KEM seed handling follow the linked AWS-LC, not the build flavor

WillChilds-Klein:
### Issue

- ACCP decided in four places whether ML-KEM seed-format PKCS#8 keys are supported by asking `isFips() && !isExperimentalFips()` -- a build-mode question standing in for a linked-libcrypto-capability one.
- It answered correctly only because `build.gradle` happens to pin mainline AWS-LC for non-FIPS and experimental FIPS and `AWS-LC-FIPS-3.1.0` for regular FIPS. It stops answering the moment a pin moves, a consumer links a different AWS-LC-FIPS, or ACCP builds against the tip of AWS-LC main -- which [AWS-LC's own CI does on every pull request](https://github.com/aws/aws-lc/pull/3471), so the breakage lands as a red AWS-LC build that no ACCP author sees.
- Two independent capabilities were conflated: whether the linked AWS-LC **retains the ML-KEM keygen seed** (so `getEncoded()` re-emits the RFC 9935 Section 6 seed CHOICE), and whether its `d2i_PrivateKey` **decodes ML-KEM PKCS#8 at all** (so ACCP's `parseMLKEMPrivateKey` is a fallover rather than the whole decoder).

### Description of changes

- **Seed retention is probed, not inferred.** `TestUtil.mlKemEmitsSeedEncoding()` generates a key and reads the CHOICE tag its PKCS#8 carries. Three sites use it: `MlKemTest.testPrivateKeyChoicesParse`, `testPrivateKeyEncodingIsSeedFormat`, and `EvpKeyFactoryTest.isMlKemWithExpandedOnlyEncoding` -- the last of which had been silently withdrawing its cross-provider PKCS#8 comparison from any FIPS build whose AWS-LC does retain the seed.
- **The probe is checked against a version oracle,** because a probe of ACCP's own output cannot distinguish "this AWS-LC keeps no seed" from "seed emission regressed", and unguarded it turns the second into a skipped test rather than a failure. `TestUtil.pinnedAwsLcRetainsMlKemSeed()` answers from the linked AWS-LC version -- true for the `awsLcMainTag` pin, false for the `AWS-LC-FIPS-3.x` family, empty for anything else -- and the new `MlKemTest.testPrivateKeyEncodingMatchesLinkedAwsLc` asserts the probe against it. Builds linking an AWS-LC neither pin names skip that one assertion, which is the only place the expected answer is genuinely unknown.
- **The malformed-PKCS#8 assertions stay strict wherever ACCP owns the decoder.** `TestUtil.linkedAwsLcLacksMlKemPrivDecode()` is true exactly for the AWS-LC-FIPS 3.x family, whose `d2i_PrivateKey` has no ML-KEM `priv_decode`. There `parseMLKEMPrivateKey` is the entire decoder, and its checks are the only thing rejecting a constructed `[1] publicKey` or a trailing byte inside the `privateKey` OCTET STRING, so those builds must reject both outright. Elsewhere AWS-LC's own decoder is entitled to ignore a field it does not read, so the assertion weakens to "must never be absorbed into the key". Both branches sit behind one helper, `assertMalformedFieldNotAbsorbed`.
- **Native:** `csrc/keyutils.cpp` compiles `mlKemKeyFromSeed` for every FIPS build rather than only non-experimental ones, so a build that defines `EXPERIMENTAL_FIPS_BUILD` while linking an AWS-LC without ML-KEM `priv_decode` gets the real helper instead of the `nullptr` stub.
- **Pin drift is called out where the pins live,** in a note beside `awsLcMainTag` / `awsLcFipsTag`. A stale test constant fails nothing; it stops matching and downgrades the assertions above to their tolerant branch. Note that `CMakeLists.txt` writes `awsLcVersionStr=AWS-LC <configured id>` in FIPS mode too, so both constants are matched as substrings -- an anchored match would never fire.

### Testing

- The first commit added two advisory GitHub Actions jobs building this PR against tip-of-main AWS-LC on AL2023 / Corretto 21 and running the full JUnit suite in non-FIPS and regular FIPS; the third reverts them, since equivalent coverage is moving to our internal CI definitions. Those runs were against the **second** commit, before the version oracle existed:
  - **Before the fix** ([run 33645086499](https://github.com/corretto/amazon-corretto-crypto-provider/actions/runs/33645086499)): non-FIPS green; FIPS red with 6 ML-KEM failures -- `testPrivateKeyChoicesParse` and `testMalformedPkcs8PrivateKeysRejected` across all three parameter sets -- plus `testPrivateKeyEncodingIsSeedFormat` skipped.
  - **After the fix** ([run 33648631705](https://github.com/corretto/amazon-corretto-crypto-provider/actions/runs/33648631705)): both modes `BUILD SUCCESSFUL`, 14/14 test containers `[ 0 tests failed ]`, no ML-KEM `[FALSE_ASSUMPTION]`. Both configurations were also reproduced locally with identical results.
- The fourth commit's version oracle is validated by this PR's pinned-AWS-LC CI, which covers the two versions the oracle names: regular FIPS on `AWS-LC-FIPS-3.1.0` takes the strict branch, and non-FIPS and experimental FIPS on `v1.72.0` take the tolerant one with the oracle expecting a seed encoding.
- **Not yet re-run against tip-of-main AWS-LC.** The oracle abstains there, which is what keeps AWS-LC's own CI green, but the jobs that would demonstrate it were reverted in the third commit, so that path is reasoned about rather than measured on this branch.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


jakemas (dismissed):

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/MlKemTest.java
```diff
@@ -756,22 +763,15 @@ public void testMalformedPkcs8PrivateKeysRejected(String paramSet) throws Except
...
+    try {
+      PrivateKey ignored =
+          keyFactory.generatePrivate(new PKCS8EncodedKeySpec(trailingInsideChoice));
+      assertArrayEquals(
+          pkcs8.getEncoded("DER"),
+          ignored.getEncoded(),
+          paramSet + " a trailing byte inside the privateKey OCTET STRING must be ignored");
+    } catch (InvalidKeySpecException expected) {
+      // Rejecting it outright is the stricter of the two acceptable answers.
     }
```
justsmth, line 781: This drops the only assertion anywhere that pins `parseMLKEMPrivateKey`'s deliberate trailing-byte rejection (the `CBS_len(&priv) != 0` checks in `keyutils.cpp`), which the parser comment justifies on security grounds. After this change, removing those checks goes green in every build.

The PR says there's no non-circular probe, but `getAwsLcVersionStr()` is a linked-libcrypto property, not a build flavor. Keeping the strict branch for the release-branch AWS-LC-FIPS preserves the coverage while tip-of-main builds still take the tolerant path:

```java
    if (TestUtil.linkedAwsLcLacksMlKemPrivDecode()) {
      // ACCP's fallover is the whole decoder here, so hold it to its documented strictness.
      assertPrivateKeyRejected(
          keyFactory,
          paramSet + " trailing byte inside the privateKey OCTET STRING",
          trailingInsideChoice);
    } else {
      try {
        PrivateKey ignored =
            keyFactory.generatePrivate(new PKCS8EncodedKeySpec(trailingInsideChoice));
        assertArrayEquals(
            pkcs8.getEncoded("DER"),
            ignored.getEncoded(),
            paramSet + " a trailing byte inside the privateKey OCTET STRING must be ignored");
      } catch (InvalidKeySpecException expected) {
        // Rejecting it outright is the stricter of the two acceptable answers.
      }
    }
```

with, in `TestUtil`:

```java
  /**
   * True when the linked AWS-LC is an AWS-LC-FIPS release branch that has no ML-KEM priv_decode, so
   * ACCP's fallover parser in keyutils.cpp is the whole ML-KEM PKCS#8 decoder. Keyed on the linked
   * library version, not the build flavor.
   * TODO [AWS-LC-FIPS 4.x]: delete once every supported AWS-LC-FIPS implements priv_decode for ML-KEM.
   */
  public static boolean linkedAwsLcLacksMlKemPrivDecode() {
    return NATIVE_PROVIDER.getAwsLcVersionStr().startsWith("AWS-LC-FIPS-3.");
  }
```

Same treatment could apply to the constructed `[1]` publicKey case above, though that one matters less since AWS-LC 5.x agrees with us.

csrc/keyutils.cpp
```diff
@@ -261,10 +266,17 @@ static EVP_PKEY* parseMLKEMPublicKey(const unsigned char* der, const int derLen)
...
+// Every FIPS build needs this, experimental FIPS included. Whether d2i_PrivateKey decodes the seed
+// CHOICE is a property of the linked AWS-LC, not of the build flavor: the AWS-LC-FIPS release
+// branches have no ML-KEM priv_decode at all, and a consumer can define EXPERIMENTAL_FIPS_BUILD
+// while linking one of them. Gating on !EXPERIMENTAL_FIPS_BUILD therefore left the fallover holding
+// the stub below in exactly the builds that need the real helper, turning every seed-format ML-KEM
+// private key into a decode failure. Compiling it for all FIPS builds costs nothing where AWS-LC
+// decodes the seed itself, because d2i_PrivateKey runs first in der2EvpPrivateKey and the fallover
+// only ever sees inputs it rejected.
+// TODO [AWS-LC-FIPS 4.x]: drop this seed-expansion helper once every AWS-LC-FIPS a FIPS build can
+//      link is able to d2i seed-format ML-KEM keys.
```
justsmth: Non-blocking. The motivating scenario ("a consumer can define `EXPERIMENTAL_FIPS_BUILD` while linking an AWS-LC-FIPS release branch") doesn't compile today: experimental FIPS unconditionally builds the ML-DSA paths (`encodeExpandedMLDSAPrivateKey`, the `MlDsaUtils` JNI, etc.) and `AWS-LC-FIPS-3.1.0` has no `EVP_PKEY_PQDSA`. The change itself is fine and strictly more robust; suggest the comment say only what's true:

```cpp
// Compiled for every FIPS build rather than only regular FIPS. Whether d2i_PrivateKey decodes the
// seed CHOICE is a property of the linked AWS-LC, not of the build flavor, and compiling this where
// AWS-LC already decodes the seed costs nothing: d2i_PrivateKey runs first in der2EvpPrivateKey, so
// the fallover only ever sees inputs it rejected. Non-FIPS builds always link mainline AWS-LC and
// get the stub below, keeping the experimental header out of them.
// TODO [AWS-LC-FIPS 4.x]: drop once the FIPS module can d2i seed-format ML-KEM keys.
```

Same trim for the include comment at L11-19.

csrc/keyutils.cpp
```diff
@@ -286,8 +298,8 @@ static EVP_PKEY* mlkemKeyFromSeed(int nid, const uint8_t* seed, size_t seed_len)
     return pkey;
 }
 #else
-// Unreachable here: d2i_PrivateKey decodes the seed CHOICE itself in this build, so the fallover only
-// ever sees inputs AWS-LC already rejected.
+// Unreachable here: non-FIPS builds link a mainline AWS-LC, whose d2i_PrivateKey decodes the seed
+// CHOICE itself, so the fallover only ever sees inputs AWS-LC already rejected.
```
justsmth: Nit: this is the same flavor->library inference the PR removes elsewhere, just a safer one (there is no non-FIPS AWS-LC without seed decode). Worth one clause acknowledging that so a future reader doesn't "fix" it in the other direction:

```cpp
// Unreachable here: non-FIPS builds link a mainline AWS-LC, whose d2i_PrivateKey decodes the seed
// CHOICE itself, so the fallover only ever sees inputs AWS-LC already rejected. This is a flavor
// inference, kept deliberately so the experimental header stays out of non-FIPS builds.
```

tst/com/amazon/corretto/crypto/provider/test/TestUtil.java
```diff
@@ -105,6 +108,53 @@ public static boolean isExperimentalFips() {
...
+   * happens to pin mainline for non-FIPS and experimental FIPS and a release branch for regular
+   * FIPS. It stops answering the moment either pin moves, or a consumer links a different
+   * AWS-LC-FIPS, or ACCP is built against the tip of AWS-LC main -- which AWS-LC's own CI does on
+   * every pull request. So probe for the capability rather than inferring it.
+   *
+   * <p>Computed on first use rather than in a static initializer: ACCP registers ML-KEM only on JDK
+   * 17 and above, and the question is meaningless where it does not.
+   *
+   * <p>TODO [AWS-LC-FIPS 5.0]: drop this probe once every AWS-LC-FIPS ACCP can link retains the
+   * seed.
```
justsmth: Nit: the probe measures keygen->encode, but `testPrivateKeyChoicesParse` uses it to predict import-seed->encode. Those coincide in every AWS-LC today (`KEM_KEY_set_raw_keypair_from_seed` stores the seed the same way keygen does) but are separate properties. One sentence here would cover it:

```java
   * <p>The probe observes freshly generated keys. Callers also use it to predict re-encoding of a
   * seed-imported key; those coincide because KEM_KEY_set_raw_keypair_from_seed retains the seed
   * exactly as keygen does, in every AWS-LC that retains it at all.
```

Also on the TODO (here and at `keyutils.cpp` L18, L277): "once every AWS-LC-FIPS ACCP can link ..." isn't a condition anyone can evaluate. The prior version-anchored phrasing ("once the FIPS module ...") was actionable; suggest restoring it.

tst/com/amazon/corretto/crypto/provider/test/jdk17plus/MlKemTest.java
WillChilds-Klein: yeah, good catch. my claude found the same thing and is currently implementing this and same for constructed public key.

WillChilds-Klein: also had to account for the non-FIPS AWS-LC version str

csrc/keyutils.cpp
prasden: nit: "Build _flavor_" seems a bit odd wording IMO, configuration might be better.

jakemas (approved):

justsmth (approved):

PR #578, opened 2026-09-02
Add security review workflow

nhatnghiho:
### Description of changes

Add security review workflow for ACCP. The reviewer will become available after this PR is merged.

### DIFF
--- a/.github/actions/check-authorization/action.yml
+++ b/.github/actions/check-authorization/action.yml
@@ -1,0 +1,43 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+name: 'check-authorization'
+description: 'A helper action to determine the authorization level of a pull request for running CI'
+outputs:
+  approval-env:
+    description: 'The target environment to use for the workflow'
+    value: ${{ steps.collab-check.outputs.result }}
+runs:
+  using: 'composite'
+  steps:
+    - name: Collaborator Check
+      uses: actions/github-script@v8
+      id: collab-check
+      with:
+        result-encoding: string
+        script: |
+          // Skip authorization check if event is not pull_request_target
+          if (context.eventName !== 'pull_request_target') {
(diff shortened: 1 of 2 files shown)


.github/actions/check-authorization/action.yml
```diff
@@ -0,0 +1,43 @@
...
+        result-encoding: string
+        script: |
+          // Skip authorization check if event is not pull_request_target
+          if (context.eventName !== 'pull_request_target') {
+            console.log(`Event type is ${context.eventName}, skipping authorization check`);
+            return "auto-approve";
+          }
+          
+          try {
+            const permissionResponse = await github.rest.repos.getCollaboratorPermissionLevel({
```
WillChilds-Klein, line 26: assuming this workflow defaults to read-only permissions. claude says

> getCollaboratorPermissionLevel needs push-level access, but the workflow token is scoped to `contents: read`, so the call always 403s and the catch block downgrades every PR -- including maintainers' -- to manual-approval.

but this workflow looks identical to aws-lc's...

https://github.com/aws/aws-lc/blob/main/.github/actions/check-authorization/action.yml

so, i guess 2 questions:

1. does this work today in aws-lc?
2. if not, can we drop it from this PR?

WillChilds-Klein (dismissed):

.github/actions/check-authorization/action.yml
nhatnghiho: According to this [Github API doc](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2026-03-10#get-repository-permissions-for-a-user), `getCollaboratorPermissionLevel` only needs `metadata: read`, which is already included in `contents: read`

1. Yes it's already working in aws-lc
2. We don't want to drop it because we don't want the reviewer to automatically run on untrusted code. This gate requires a team member's approval for the reviewer to start running.

WillChilds-Klein (approved):

prasden (approved):

crlorentzen (approved):

PR #580, opened 2026-09-04
Make build.gradle compatible with Gradle 9

WillChilds-Klein:
**Stack 2/4** -- based on #583. Review that one first; this PR's own diff is the single commit `Make build.gradle compatible with Gradle 9`.

| # | Branch | Contents |
|---|---|---|
| #583 | `accp-mac-jdk25-interim` | JDK 25 runtime job, builds/targets 21 |
| **#580 (this PR)** | `accp-gradle9-compat` | `build.gradle` Gradle 9 compatibility |
| #581 | `accp-gradle9-wrapper` | Wrapper 8.11.1 -> 9.7.1 |
| #582 | `accp-jdk25-ci` | JDK 25 build/target job |

Pure no-op refactor on the current 8.11.1 wrapper. It only removes the blockers so the wrapper can be bumped in #581.

## What changed

- Gradle 9 removed `Project.exec()` from task actions. The 14 execution-time `exec {}` blocks now go through an injected `ExecOperations` instance, which is the documented replacement and behaves identically on Gradle 8.
- The two multi-string dependency declarations become single-string notation. Multi-string notation is deprecated in Gradle 9 and fails in Gradle 10.

Single file, +25/-16.

## Verification

Full `./gradlew release` on the unmodified 8.11.1 wrapper, build/target/test all Corretto 21: `BUILD SUCCESSFUL`, exit 0.

The same change was also verified green on a 9.7.1 wrapper as part of #581.

## Ordering note

The internal CI Docker images need their `JAVA_HOME` moved off Corretto 11 before #581 lands. Tracked separately in the internal image package; nothing in this PR depends on it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import groovy.xml.*
+import javax.inject.Inject
+import org.gradle.process.ExecOperations
 
 plugins {
     id 'java'
@@ -11,6 +13,13 @@ plugins {
     id "com.google.osdetector" version "1.7.3"
     id "io.github.gradle-nexus.publish-plugin" version "2.0.0"
 }
+
+// Gradle 9 removed Project.exec() from task actions; ExecOperations is the
+// supported replacement. Inject it once and reuse it from every task action.
+interface InjectedExecOps {
+    @Inject ExecOperations getExecOps()
+}
+def execOps = objects.newInstance(InjectedExecOps).execOps
 
 group = 'software.amazon.cryptools'
 version = '2.6.0'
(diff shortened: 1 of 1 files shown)


prasden (dismissed):

build.gradle
```diff
@@ -296,7 +305,7 @@ def getStagedArtifact(platformClassifier, destDir) {
     println "Loaded staging id is " + stagingProperties['staging.id']
     def repoBase = "https://aws.oss.sonatype.org/content/repositories/" + stagingProperties['staging.id']
     mkdir "${destDir}"
-    exec {
+    execOps.exec {
```
justsmth, line 309: The script-local `def execOps` is captured by task closures, but is not visible inside this method. `downloadStagedJar` fails here with `Could not get unknown property 'execOps'` on both Gradle 8.11.1 and 9.7.1; the `-DstagingProperties` path also calls this method. Could we inject it locally as below, or pass it into `getStagedArtifact()` from both callers?

```suggestion
    def execOps = objects.newInstance(InjectedExecOps).execOps
    execOps.exec {
```

prasden (approved):

justsmth (approved):

PR #582, opened 2026-09-04
Add a macOS JDK 25 CI job

WillChilds-Klein:
## Description

- Adds a macOS Arm64 CI job that builds, targets and tests on Corretto 25, so the `jdk17plus` ML-KEM overlay runs on the newest LTS runtime.
- Gives every macOS Arm64 target-variant job a one-line header comment, so the four jobs read as a coverage matrix (17->17, 21->21, 25->21, 25->25) instead of near-duplicates.
- Bumps JaCoCo to 0.8.15: `release` depends on `coverage`, so the agent instruments the class-file 69 output of a JDK 25 target build, which 0.8.13 supports only experimentally.
- Pins google-java-format to 1.28.0. Spotless' default 1.24.0 throws `NoSuchMethodError` on a JDK 25 build JDK, and 1.29.0 will not run on JDK 17.
- Moves the benchmark script off Corretto 11, which can no longer run the Gradle 9 wrapper.

## Testing / verification

- Full `./gradlew release` with build JDK Corretto 25.0.4.1, `--target-jdk-version 25` and test JDK Corretto 25: 18636 tests, no failures, emitted classes at class-file major 69.
- `spotlessJavaCheck` passes on an unmodified tree with a Corretto 25 build JDK, so the pinned formatter both runs on 25 and agrees with the formatting already in the repo.
- JaCoCo 0.8.15's agent, core and report jars are still class-file major 49, which keeps the JDK 8 test jobs working.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -94,7 +94,13 @@ spotless {
   java {
     target 'src/**/*.java', 'tst/**/*.java'
     licenseHeaderFile 'build-tools/license-headers/LicenseHeader.java'
-    googleJavaFormat().reflowLongStrings()
+    // Pin google-java-format instead of taking spotless' default (1.24.0), which
+    // calls Log$DeferredDiagnosticHandler.getDiagnostics(); that signature changed
+    // in JDK 25, so the default throws NoSuchMethodError once the build JDK is 25.
+    // 1.28.0 is the newest release that still runs on a JDK 17 build JDK -- 1.29.0
+    // needs JCTree$JCAnyPattern, which does not exist before JDK 21 -- and it
+    // formats this codebase identically to 1.24.0.
+    googleJavaFormat('1.28.0').reflowLongStrings()
   }
 
   // clang-format is difficult to configure across all of our platforms so we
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -191,7 +197,10 @@ repositories {
 
 dependencies {
     // Separate so we can extract the jar for the agent specifically
-    jacocoAgent 'org.jacoco:org.jacoco.agent:0.8.13:runtime'
+    // 0.8.13 only supports Java 25 class files experimentally; 0.8.14 made 25
+    // official and 0.8.15 does the same for 26. The agent, core and report jars
+    // are still Java 5 bytecode, so this stays usable on the JDK 8 test runtime.
+    jacocoAgent 'org.jacoco:org.jacoco.agent:0.8.15:runtime'
```
prasden, line 196: NP: not sure if we need to document dependency upgrades?

prasden (dismissed):

justsmth (approved):

justsmth (dismissed):

prasden (approved):

justsmth (approved):

